PR #15, opened 2019-10-25
dependencies: update invenio-records-permissions

fenekku:
Depend on invenio-records-permissions with updated config. Travis will fail until that version is released. Blocked by: https://github.com/inveniosoftware/invenio-records-permissions/pull/19


fenekku:
Ah, I see and just have invenio-rdm-records depend on invenio-records-permissions. Ok I will move this there (if it's not already)

Issue #27, opened 2019-11-26
Organize invenio-app-rdm configuration


fenekku:
Like what was done in cookiecutter-invenio-rdm, organize the configuration to make it usable as a reference and make sure things are well set and organized.

PR #28, opened 2019-11-26
config: organize config.py [more]

fenekku:
- Grouped configuration
- Added documentation links
- Kept ad-hoc order of groups but sorted alphabetically
  within a group
- Removed some configurations that repeated defaults
  already provided by modules
- Closes: #27


fenekku:
Agreed. Once we work on customization, we will have a good idea of the THEME to include.

PR #34, opened 2019-12-04
ui: update default frontpage

### LINKED ISSUE #30: ui: front page
Ideas:

- RDM logo
- Product page gradient
- Search bar
- Features
- Partners

Basically copy the RDM product page. 

This will be changed not only by us, but by partners, so not much engineering is needed here.

fenekku:
invenio_app_rdm_default_frontpage
Close #30


invenio_app_rdm/config.py
```diff
@@ -112,6 +109,11 @@ def _(x):
 THEME_FRONTPAGE_TITLE = _('Invenio App RDM')
```
ppanero: Moderate: Shall we put here "The turn-key research data management repository"

invenio_app_rdm/theme/templates/invenio_app_rdm/frontpage.html
```diff
@@ -1,17 +1,178 @@
...
+
 {%- block page_body %}
-<div class="container marketing">
-  <div class="row">
-    <div class="col-lg-12">
-    <h1 class="text-center">Welcome to Invenio App RDM.</h1>
+
+<section class="invenio-rdm-project-section section-content-white-bg">
+  <div class="container">
+    <span class="section-title">You've installed InvenioRDM successfully!</span>
```
ppanero: Comment:

"You've successfully installed InvenioRDM!"

fenekku: Agreed! You (and I) prefer it because, readers naturally pay more attention to the first and last word of a sentence.

fenekku:
Here is what the frontpage without the "InvenioRDM Project" section:

invenio_app_rdm_default_frontpage2

I was also thinking of rewording the "InvenioRDM Project" section to state what it is, rather than what it wants to be (goals). Also adding the links to the forum and the gitter chat. But I think for now, just using the original one is good enough.

invenio_app_rdm/config.py
fenekku: Actually, we should probably just have the cookiecutter title override this anyway. Made a PR for that: https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/24

ppanero: Nice idea!

invenio_app_rdm/theme/templates/invenio_app_rdm/frontpage.html
ppanero: :+1:

fenekku:
Changed the `THEME_SITENAME`, but didn't touch the header part since I think you will be touching it.

PR #37, opened 2019-12-09
config: set external configuration via config.py

### LINKED ISSUE #36: Forcefully set configuration for InvenioRDM
The `invenio-app-rdm` module should ensure that InvenioRDM specific configurations are set. 

This is a non-trivial issue because:

- there is no guaranteed configuration loading order among invenio modules.
- `invenio-rdm-records` defines an InvenioRDM record in part via configuration, but it may be loaded after or before the configuration it is meant to change thus not working.  
- `invenio-rdm-records` used to forcefully set its configuration which would suffer from same problem but with invenio-app-rdm (this is fixed in https://github.com/inveniosoftware/invenio-rdm-records/pull/39) 

The simple solution is: https://github.com/inveniosoftware/invenio-app-rdm/pull/37
We could write an `ext.py` and re-implement `flask/config.py::from_object`, but I don't see any real benefit.

fenekku:
- This approach has the advantange of being very simple
  and non-coupling. But it is new.
- Using ext.py could achieve the same thing, but it's more code
- closes #36

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -8,15 +8,29 @@
 
 """Default configuration for Invenio App RDM.
 
-Each setting has a sensible default value. You overwrite and set
-instance-specific configuration by either:
+As a flavour extension, Invenio-App-RDM doesn't define configuration
+variables of its own, but rather forcefully sets other modules'
+configuration variables.
+
+Import below the configuration defined in other modules that should be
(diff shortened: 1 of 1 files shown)


fenekku:
Good!

> it is not fully explicit what are we importing

Yes. I initially was wondering about that too, but looking at the config_loader code, I found that it only loads all caps variables anyway. So we can be pretty safe in knowing that only CONFIG_VARIABLES are loaded.

PR #41, opened 2019-12-12
ui: several css fixes on color and default text changes

ppanero:
Changes:

- Navbar gradient inverted again in order to have the blue background behind `InvenioRDM` logo.
- Footer color was not matching. Now it follows the same blue.
- In the main page the app text has been changed to reflect the product page ("The turn-key...").
- "Powered by InvenioRDM" text in the footer.
- Removed update date from record view.
- Changed "Sign up" button color. Lars pointed out that red might reflect "negative" actions. Changed to green. The original "orange/yellow" did not match at all. We should work on a palette of colors. Feel free to change this color if needed.

Ready to be merged :)

### DIFF
--- a/invenio_app_rdm/theme/assets/scss/invenio_app_rdm/theme.scss
+++ b/invenio_app_rdm/theme/assets/scss/invenio_app_rdm/theme.scss
@@ -31,14 +31,14 @@ h6 {
 }
 
 .navbar {
-  background-image: linear-gradient(348deg, rgba(3, 119, 205, 1), rgba(3, 119, 205, 1) 52.52%, rgba(251, 130, 115, 1));
+  background-image: linear-gradient(20deg, rgba(3, 119, 205, 1), rgba(3, 119, 205, 1) 52.52%, rgba(251, 130, 115, 1));
   background-color: unset;
   border-color: transparent;
-}
 
-.navbar .btn-warning {
-    background-color: #fb8172;
-    border-color: #fb8172;
+  .btn[href^="/signup"] {
+    background-color: #44a244;
+    border-color: #44a244;
+  }
 }
 
 .centered {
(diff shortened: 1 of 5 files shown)


invenio_app_rdm/theme/assets/scss/invenio_app_rdm/theme.scss
```diff
@@ -31,14 +31,14 @@ h6 {
...
+  background-image: linear-gradient(20deg, rgba(3, 119, 205, 1), rgba(3, 119, 205, 1) 52.52%, rgba(251, 130, 115, 1));
   background-color: unset;
   border-color: transparent;
 }
 
 .navbar .btn-warning {
-    background-color: #fb8172;
-    border-color: #fb8172;
+    background-color: #44a244;
+    border-color: #44a244;
```
fenekku: I would change the `btn-warning` class to something else: there is a clash of meaning here: warning in green (#44a244).

invenio_app_rdm/theme/templates/invenio_app_rdm/footer.html
```diff
@@ -16,7 +16,7 @@
         {%- endif %}
       </div>
       <div class="col-xs-6 col-xs-pull-6">
-        {% trans invenio_rdm="http://inveniosoftware.org/products/rdm", invenio="http://inveniosoftware.org/products/framework" %}<a href="{{invenio_rdm}}">InvenioRDM</a> instance powered by <a href="{{invenio}}">Invenio</a>{% endtrans %}
+        {% trans invenio_rdm="http://inveniosoftware.org/products/rdm", invenio="http://inveniosoftware.org/products/framework" %}Powered by<a href="{{invenio_rdm}}">InvenioRDM</a>{% endtrans %}
```
fenekku: Let's remove the `invenio` attribute from `trans` since it isn't used anymore.

fenekku:
Agreed with the palette of color idea. We could then place the colors in the `variables.scss` and reuse them.

Issue #51, opened 2020-01-09
PDF previewer


mb-wali:
I have tried to upload a pdf file to a specific record. uploading worked fine 👍 , i can also download the file from the record.

BUT the issue is viewing the pdf file.

### console error
```
[2020-01-09 15:28:44,376] WARNING in views: Preview failed for test.pdf, in recid:d46f9-vz793
raise BuildError(endpoint, values, method, self)
werkzeug.routing.BuildError: Could not build url for endpoint 'invenio_records_ui.recid_files' with values ['filename', 'pid_value']. Did you mean 'invenio_records_ui.recid_file' instead?
```

### screenshot from postman:
![postman pdf](https://user-images.githubusercontent.com/44528277/72077046-e4c9de80-32f6-11ea-83aa-83cd48bcec63.PNG)

### screenshot from record page.
![screenshotpdfIssue](https://user-images.githubusercontent.com/44528277/72076547-f78fe380-32f5-11ea-84bb-2e9f7d04572b.PNG)

PR #52, opened 2020-01-10
ui: fix files preview in records view

ppanero:
Changes:
* `RECORDS_UI_ENDPOINTS` naming. Since some code is reused from `invenio-previewers` it was looking for `recid_files` and `recid_previewer` endpoing and not finding them. Now it is consistent.

Example:

``` console
curl -k -XPUT https://localhost:5000/api/records/54t33-38526/files/test.pdf -H "Content-Type: application/octet-stream" --data-binary @test.pdf
```

image

Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/51

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -276,13 +276,13 @@ RECORDS_UI_ENDPOINTS = {
         'route': '/records/<pid_value>',
         'template': 'invenio_app_rdm/record_view_page.html'
     },
-    'recid_file': {
+    'recid_files': {
         'pid_type': 'recid',
         'record_class': 'invenio_records_files.api:Record',
         'route': '/records/<pid_value>/files/<path:filename>',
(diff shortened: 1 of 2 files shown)


fenekku:
> Since some code [...] looking for `recid_files` and `recid_previewer` endpoints

I couldn't find that code... I don't fully understand what is going there :sweat_smile:

ppanero:
@fenekku I can confirm it works for PDF and images:

 (JPG)

image

(PDF)

image

However, for PDFs in local I was getting an JS error due to the secure content policy:

image

As you can see the previewer works (which was not happening before), it just that it cannot load the content:

image

Fixed by adding this:
```
from invenio_app.config import APP_DEFAULT_SECURE_HEADERS

APP_DEFAULT_SECURE_HEADERS['content_security_policy'] = {}
```
(Only needed for `--local` according to my testing). 

IMO, it's good to go. We can fix the `secure-content-policy` in a separate issue. WDYT?

PR #59, opened 2020-01-27
theme: add config for instance theme file

zzacharo:


invenio_app_rdm/theme/webpack.py
```diff
@@ -10,13 +10,17 @@
...
+from flask import current_app
 from flask_webpackext import WebpackBundle
 
 theme = WebpackBundle(
     __name__,
     'assets',
     entry={
-        'invenio-app-rdm-theme': './scss/invenio_app_rdm/theme.scss',
+        'invenio-app-rdm-theme': os.path.join('./scss/instance_theme', current_app.config.get(
+            'INSTANCE_THEME_FILE', './scss/invenio_app_rdm/theme.scss')),
```
fenekku: Obviously depends on the application context issue resolution. But also:

Normal: This results in `''./scss/instance_theme/./scss/invenio_app_rdm/theme.scss'`. What about this:

```python
'invenio-app-rdm-theme': current_app.config.get('INSTANCE_THEME_SCSS', './scss/invenio_app_rdm/theme.scss')
```

where `INSTANCE_THEME_SCSS` is a full path to the scss file? We should also add `INSTANCE_THEME_SCSS` to the `config.py` so that this configuration variable is documented.

zzacharo: @fenekku you are right, my intention was to create the fixed `instance_theme` folder only if the config variable `INSTANCE_THEME_FILE ` existed otherwise to fallback to `'./scss/invenio_app_rdm/theme.scss'`. Also I wanted to hide as much of this from the end user and not have to oblige a specific structure as in `assets` approach. But as you noticed what I am doing now is wrong but I would apply probably do is calculating if the config variable is there and then apply the `.scss/instance_theme` prefix otherwise fallback to the 
`'./scss/invenio_app_rdm/theme.scss'`. What do you think?

fenekku: hmm... checking if the config variable is there and then adding the prefix if it is prevents us from using the config variable ourselves as the default... 

What about hiding/auto-prefixing the `scss` part and having users and libraries need to namespace (pretty much like the templates case): `instance_theme/my_theme.scss` or `invenio_app_rdm/theme.scss` should be passed?

PR #62, opened 2020-02-10
search: add search page

zzacharo:
* Integrates `react-searchkit` and removes `invenio-search-ui`.
* Initial implementation of the search page under a `search` module.
* Adds template customization.
* Adds css customization. This will change when semantic-ui theming
  will be integrated.


invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/templates.js
```diff
@@ -0,0 +1,7 @@
+import { ResultsItemTemplate } from "@templates/search/ResultsItemTemplate";
```
slint: Missing header here

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/templates.js
```diff
@@ -0,0 +1,7 @@
...
+import { ResultsGridItemTemplate } from "@templates/search/ResultsGridItemTemplate";
```
slint: I guess something has to go in the Webpack config for this to be resolved right?

invenio_app_rdm/theme/templates/invenio_app_rdm/search.html
```diff
@@ -6,28 +6,22 @@
...
-  >
-</invenio-search-select-box>
-{%- endblock search_sort_select %}
-
-{%- block javascript %}
+<!DOCTYPE html>
+<html>
+  <head>
+    {%- block css %}
+    {{ webpack['vendor.css']}}
```
slint: I guess these can be extracted at some point into an RDM-specific base page.html template

fenekku:
@zzacharo this is marked as a draft PR, you want us to wait before looking at it?

zzacharo:
This PR needs:
* https://github.com/inveniosoftware/invenio-assets/pull/105/files
* https://github.com/inveniosoftware/react-searchkit/pull/91
* https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/42

All PRs are Draft as I want to first take your feedback on how it looks like :)

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchResults/SearchResults.js
```diff
@@ -0,0 +1,92 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2019 CERN.
+// Copyright (C) 2019 Northwestern University.
+//
+// Invenio App RDM is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import React, { Component } from "react";
+import { Container, Grid } from "semantic-ui-react";
+import _truncate from "lodash/truncate";
```
fenekku: discussion: if we only need `truncate`, my vote would be to implement it ourselves rather than bring lodash in, unless we do plan on using other lodash functions. Just want us to be careful about accruing dependencies and load size.

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchResults/SearchResults.js
```diff
@@ -0,0 +1,92 @@
...
+  EmptyResults,
+  Error,
+  ResultsLoader,
+  withState
+} from "react-searchkit/esm";
+
+import { ResultsGridItem, ResultsListItem } from "./components";
+import { config } from "../config";
+
+const SpanWithMargin = ({ text, margin }) => {
```
fenekku: minor: just a `function` is fine (wouldn't even need `let`)...

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchResults/SearchResults.js
```diff
@@ -0,0 +1,92 @@
...
+      break;
+    default:
+      style = { margin: `0 ${size}` };
+  }
+  return <span style={style}>{text}</span>;
+};
+
+class Results extends Component {
+  render() {
+    const { total } = this.props.currentResultsState.data;
```
fenekku: question: `this.props.currentResultsState` comes from `withState`? rather than using some form of inheritance? Just curious about choice.

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/Searchbar/Searchbar.js
```diff
@@ -0,0 +1,87 @@
...
+
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { Input } from "semantic-ui-react";
+import {
+  UrlHandlerApi,
+  onQueryChanged as triggeronQueryChanged
+} from "react-searchkit/esm";
+
+export class SearchBar extends Component {
```
fenekku: question: (these are all just questions for me to better understand the architecture: not meant as critiques) Seeing the need for us to define `onKeyPress` below seems like a lot... are we doing very specific customizations here?

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/index.js
```diff
@@ -0,0 +1,68 @@
...
+const aggregations = [
+  {
+    title: "Access Right",
+    agg: {
+      field: "access_right",
+      aggName: "access_right"
+    }
+  },
+  {
+    title: "Types",
```
fenekku: Let's call it "Resource Types" because "Types" is *very* generic.

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/index.js
```diff
@@ -0,0 +1,68 @@
...
+  timeout: 5000
+};
+
+export const config = {
+  logo,
+  searchApi,
+  aggregations,
+  sortValues,
+  resultsPerPageValues
+};
```
fenekku: If we can, how do we override these from the application's configuration (`config.py`)? `localhost:5000` is bound to require a change right?

invenio_app_rdm/theme/templates/invenio_app_rdm/search/header_search.html
```diff
@@ -0,0 +1,39 @@
+{# Copyright (C) 2019 CERN. Copyright (C) 2019 Northwestern University. Invenio
+App RDM is free software; you can redistribute it and/or modify it under the
+terms of the MIT License; see LICENSE file for more details. #}
```
fenekku: minor: missing new lines here?

invenio_app_rdm/theme/templates/invenio_app_rdm/search/search.html
```diff
@@ -0,0 +1,36 @@
+{#
+    Copyright (C) 2019 CERN.
+    Copyright (C) 2019 Northwestern University.
+
+    Invenio App RDM is free software; you can redistribute it and/or modify it
+    under the terms of the MIT License; see LICENSE file for more details.
+#}
+{%- extends "invenio_theme/frontpage.html" %}
```
fenekku: Shouldn't we inherit from the `config.BASE_TEMPLATE` ?

setup.py
```diff
@@ -64,9 +64,17 @@
...
 install_requires = [
-    'Invenio[base,auth,metadata,files]{}'.format(invenio_version),
+    'Invenio[base,auth,files]{}'.format(invenio_version),
+    'invenio-indexer>=1.1.1,<1.2.0',
+    'invenio-jsonschemas>=1.0.1,<1.1.0',
+    'invenio-oaiserver>=1.1.1,<1.2.0',
+    'invenio-pidstore>=1.1.0,<1.2.0',
+    'invenio-records-rest>=1.6.4,<1.7.0',
+    'invenio-records-ui>=1.0.1,<1.1.0',
+    'invenio-records>=1.3.0,<1.4.0',
```
fenekku: question: why did we have to explode the metadata bundle again?

invenio_app_rdm/config.py
```diff
@@ -333,3 +322,10 @@ def _(x):
 # =============
 INSTANCE_THEME_FILE = './scss/invenio_app_rdm/theme.scss'
 """Instance's theme file path to override the default theme."""
+
+INSTANCE_SEARCH_THEME_FILE = './scss/invenio_app_rdm/search/theme.scss'
+"""Instance's search theme file path to override the default search theme."""
+
+SEARCH_HEADER_TEMPLATE = 'invenio_app_rdm/search/header_search.html'
```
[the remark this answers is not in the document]
zzacharo: @fenekku yes if the directory is ok to stay then is enough I guess. I kept the name convention for header as it is https://github.com/inveniosoftware/invenio-theme/tree/master/invenio_theme/templates/invenio_theme for frontpage.

invenio_app_rdm/config.py
```diff
@@ -333,3 +322,10 @@ def _(x):
...
+
+SEARCH_BASE_TEMPLATE = 'invenio_app_rdm/search/search.html'
```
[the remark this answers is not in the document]
zzacharo: I put this code in a search module here but my intention for the longer run, and I think this is what probably would be done if I am not wrong, is to replace `invenio-search-ui` code with this one. I am not sure if `invenio-rdm-records` should be also responsible for the search but this can be discussed during our call today I guess!

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/App.js
```diff
@@ -0,0 +1,38 @@
...
+  render() {
+    return (
+      <ReactSearchKit searchApi={searchApi}>
+        <Container>
+          <Grid relaxed style={{ padding: "2em 0" }}>
+            <Grid.Row columns={2}>
+              <Grid.Column width={4} className="search-aggregations">
+                <SearchFacets />
+              </Grid.Column>
+              <Grid.Column width={12}>
```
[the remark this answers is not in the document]
zzacharo: Semantic-ui is 16 columns grid system https://semantic-ui.com/collections/grid.html

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchFacets/SearchFacets.js
```diff
@@ -0,0 +1,58 @@
...
+            label={<label>{key}</label>}
+            value={key}
+            onClick={() => onFilterClicked(key)}
+            checked={isSelected}
+          />
+          {childAggCmps}
+        </List.Content>
+      </List.Item>
+    );
+  };
```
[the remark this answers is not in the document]
zzacharo: Arrow functions are actual the solution on `this-bind` issue. The only caveat is that they are not inherited but in React you should not inherit other components as it can lead to some nasty errors. Did I get correctly your question?

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchFacets/index.js
```diff
@@ -0,0 +1 @@
+export { SearchFacets } from "./SearchFacets";
```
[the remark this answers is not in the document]
zzacharo: This is just a name convention it is used in React and we also followed in `invenio-app-ils`. It introduces a bit more boilerplate but from the other hand I think makes the structure more clear. Also, the `index.js` is used in some places to be the file in which you connect your component with redux state. This allows you test your components pure without being obliged to setup a redux store in order to test them.

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchResults/SearchResults.js
zzacharo: This import will only include the `truncate` function and the whole lodash library. On top of that lodash is coming already with searchkit :) But I agree to be careful with the dependencies we are using :)

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchResults/SearchResults.js
zzacharo: Yes `withState` is a high order function that gives you the chance to hook in searchkit's internal store and consume it in your component. Is defined [here](https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/HOC/withState.js)!

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/index.js
zzacharo: The whole config mechanism and how it is communicated from the backend is not part of this PR. I introduced the `config.js` just to start having one central place that my components are getting parametrized values and then this module will be the responsible to get the configuration from the backend.

invenio_app_rdm/theme/templates/invenio_app_rdm/search/search.html
zzacharo: This is open to discussion! I did that because is seemed the easiest to so I can have the same style as the frontpage. But in general part of the discussion today would be to see the template architecture and from where to inherit or which blocks should be defined so they can be overriden. Or even which templates should come through config so in my instance I can pass my own. Currently I just exposed the whole search page template plus the header :)

setup.py
zzacharo: Because of `invenio-search-ui`! But if we go for moving the `search` ui folder over then this can be imploded again :)

invenio_app_rdm/config.py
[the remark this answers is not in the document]
ppanero: I am with Zach on this. I think the search is also not as tight to the records (okay, u need a datamodel/API) but it might be independent of `rdm-records` somehow, its more `invenio-records` and its rest sibling.

I would leave it here for now (It's how it is done in the `cookiecutter-invenio-instance`) and in the long run it will go to `invenio-search-ui`.

invenio_app_rdm/config.py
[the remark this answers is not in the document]
ppanero: hmm, I see the "convention" but I also vote for `search/header.html`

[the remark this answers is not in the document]
ppanero: hmm, I see the "convention" but I also vote for `search/header.html`

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/index.js
```diff
@@ -0,0 +1,68 @@
...
+  {
+    text: "50",
+    value: 50
+  }
+];
+
+const logo = "/static/images/logo.svg";
+
+const searchApi = {
+  url: "https://localhost:5000/api/records/",
```
ppanero: Does this work also in case of containerized? (Wondering about ports, etc.) tho it should since uwsgi will still run in 5000.

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/Searchbar/index.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
```
ppanero: Missing copyright :)

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchFacets/index.js
ppanero: Missing copyright

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchResults/components/ResultsGridItem/index.js
```diff
@@ -0,0 +1 @@
+export { ResultsGridItem } from "./ResultsGridItem";
```
ppanero: Missing copyright

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchResults/components/ResultsListItem/index.js
```diff
@@ -0,0 +1 @@
+export { ResultsListItem } from "./ResultsListItem";
```
ppanero: Missing copyright. Ok in many files (mostly index.js files)  hehe just check they all have it :)

invenio_app_rdm/theme/assets/templates/search/ResultsGridItemTemplate.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Card, Image } from 'semantic-ui-react';
+import _truncate from 'lodash/truncate';
+
+export function ResultsGridItemTemplate(result, index) {
+    return (
+      <Card fluid key={index} href={`#`}>
+        <Card.Content>
+          <Card.Header>{result.metadata.title}</Card.Header>
```
ppanero: This changes with the new datamodel, just be aweare (or wait till its merged). It should be something like:

``` js
result.metadata.titles[0].title
```

invenio_app_rdm/theme/assets/templates/search/ResultsGridItemTemplate.jsx
```diff
@@ -0,0 +1,16 @@
...
+          <Card.Description>
+            {_truncate(result.metadata.description, { length: 200 })}
```
ppanero: This changes with the new datamodel, just be aweare (or wait till its merged). It should be something like:

``` js
result.metadata.descriptions[0].description
```

invenio_app_rdm/theme/assets/templates/search/ResultsItemTemplate.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import {Item} from 'semantic-ui-react';
+import _truncate from 'lodash/truncate';
+
+export function ResultsItemTemplate(result, index) {
+    return (
+      <Item key={index} href={`#`}>
+        <Item.Content>
+          <Item.Header>{result.metadata.title}</Item.Header>
```
ppanero: Same as above for both title and description

invenio_app_rdm/theme/templates/invenio_app_rdm/search/macros/searchbar.html
```diff
@@ -0,0 +1,3 @@
+{% macro render_searchbar() %}
+    <div id="search-bar"></div>
+{% endmacro %}
```
ppanero: missing new line?

invenio_app_rdm/theme/templates/invenio_app_rdm/search.html
zzacharo: @slint no for now because in the `vendor.css` semantic-ui is included while in the module specific css, e.g `invenio-app-rdm-theme.css` bootstrap is included. As long as we don't have a clear mechanism to separate these 2 currently, I need to include `vendor.css` in the dedicated search page. But in the long run I agree with you about the common rdm page.html.

invenio_app_rdm/theme/webpack.py
```diff
@@ -26,8 +26,25 @@ def theme():
         entry={
             'invenio-app-rdm-theme': current_app.config['INSTANCE_THEME_FILE'],
             'invenio-app-rdm-js': './js/invenio_app_rdm/inveniordm.js',
```
ppanero: Wasn't this file removed? If so, not sure of the implications (should be fine tho).

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/index.js
zzacharo: It should as everything run in the same port. But yes this should come from the backend in the longer run!

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/Searchbar/Searchbar.js
zzacharo: This is just to enable on Enter search :) I think is quite useful to trigger the search on enter no? :)

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchMain/SearchFacets/SearchFacets.js
```diff
@@ -0,0 +1,58 @@
+// This file is part of InvenioRDM
```
ppanero: Could you crosscheck with @lnielsen or @slint if this line is needed/affects something? is not there in many of the modules...

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/templates.js
ppanero: An alias in invenio-assets, latest PR :)

invenio_app_rdm/theme/assets/templates/search/ResultsGridItemTemplate.jsx
ppanero: Now the module is released, so if you rebase on master, this will actually **not work** :) need to update (suggestion above)

invenio_app_rdm/theme/views.py
```diff
@@ -67,3 +67,9 @@ def can_list_files(record):
     """
     PermissionPolicy = get_record_permission_policy()
     return PermissionPolicy(action='read_files', record=record).can()
+
+
+@blueprint.route('/search/')
```
ppanero: doubt: the previous does not have the final `/` and sometimes flask goes bananas with added/missing slashes.... tests this and make it consistent (ignore the comment if it is)

invenio_app_rdm/config.py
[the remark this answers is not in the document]
fenekku: Got it. Some of it will move to invenio-search-ui: https://github.com/inveniosoftware/invenio-app-rdm/issues/79

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/App.js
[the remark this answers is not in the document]
fenekku: Got it.

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchFacets/SearchFacets.js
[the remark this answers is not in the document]
fenekku: Yes. That's what I thought about the anonymous function. But I didn't know about how inheritance was not recommended in React. Thanks.

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/index.js
fenekku: Got it. Made the issue for it: https://github.com/inveniosoftware/invenio-app-rdm/issues/82

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchFacets/index.js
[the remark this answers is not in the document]
fenekku: Ok.

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchResults/SearchResults.js
fenekku: If react-searchkit already uses it that's fine :)

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchResults/SearchResults.js
fenekku: Haha, I am going to need to learn me some more react, 'cause that function is gibberish to me right now :smile:

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/Searchbar/Searchbar.js
fenekku: My puzzlement here, was exactly over *us* having to define the "search on enter" rather than react-searchkit doing it for us. I think this was answered by our discussion online: some of this will be moved to invenio-search-ui and *it* will be the one defining it for us. So all good.

invenio_app_rdm/theme/templates/invenio_app_rdm/search/search.html
fenekku: ... yeah, my vote is still to inherit from `config.BASE_TEMPLATE` ;)  (frontpage is just a specialized page)

setup.py
fenekku: Got it.

invenio_app_rdm/theme/assets/templates/search/ResultsGridItemTemplate.jsx
fenekku: Right, that's the situation we will get in with model and use of model being in separate modules...

Maybe a facade over the returned results is needed so that we can have an interface to the result that will be stable across version changes? Something like defining a javascript model for a search result that is fed from the search API. It would present a uniform interface in terms of fields/methods to JSX usage irrespective of the changes in the serialization: it would be duplication, but it would serve the practical goal of resilience. It would be coded to present defaults if search API changes to ensure resilience. (sort of the idea for Ember models I think)
Something to consider maybe if makes sense.

invenio_app_rdm/config.py
```diff
@@ -34,7 +34,7 @@
     IIIF_PREVIEW_TEMPLATE, PIDSTORE_RECID_FIELD, PREVIEWER_PREFERENCE, \
     RECORD_CONTRIBUTOR_TYPES, RECORD_CONTRIBUTOR_TYPES_LABELS, \
     RECORDS_FILES_REST_ENDPOINTS, RECORDS_PERMISSIONS_RECORD_POLICY, \
-    RECORDS_REST_ENDPOINTS, RECORDS_UI_ENDPOINTS
+    RECORDS_REST_ENDPOINTS, RECORDS_REST_FACETS, RECORDS_UI_ENDPOINTS
```
ppanero: Might be the sort also fails due to lack of import?

setup.py
```diff
@@ -64,9 +64,17 @@
...
-    'invenio-rdm-records>=0.9.0',
+    'Invenio[base,auth,files]{}'.format(invenio_version),
+    'invenio-indexer>=1.1.1,<1.2.0',
+    'invenio-jsonschemas>=1.0.1,<1.1.0',
+    'invenio-oaiserver>=1.1.1,<1.2.0',
+    'invenio-pidstore>=1.1.0,<1.2.0',
+    'invenio-records-rest>=1.6.4,<1.7.0',
+    'invenio-records-ui>=1.0.1,<1.1.0',
+    'invenio-records>=1.3.0,<1.4.0',
+    'invenio-rdm-records>=0.0.9',
```
ppanero: Careful! is 0.9.0

setup.py
```diff
@@ -64,9 +64,17 @@
...
     'invenio-records-permissions>=0.7.0',
+    'werkzeug>=0.16.1,<1.0.0',
```
ppanero: Is this needed? I think is fixed already in other packages if not just bump babelex and invenio-accounts.

setup.py
zzacharo: oups, thanks!

setup.py
zzacharo: I didn't know if it is fixed, is it?

PR #80, opened 2020-02-19
dependencies: decompose files bundle due to new files-rest requirement

ppanero:


setup.py
```diff
@@ -64,16 +64,21 @@
...
     'invenio-pidstore>=1.1.0,<1.2.0',
     'invenio-records-rest>=1.6.4,<1.7.0',
     'invenio-records-ui>=1.0.1,<1.1.0',
     'invenio-records>=1.3.0,<1.4.0',
-    'invenio-rdm-records>=0.9.0',
-    'invenio-records-permissions>=0.7.0',
+    # Due to files-rest
+    'invenio-files-rest>=1.1.0,<1.2.0',
+    'invenio-iiif>=1.0.0,<1.1.0',
+    'invenio-previewer>=1.1.0,<1.2.0',
```
fenekku: Completely approve. I wouldn't be surprised if we need to keep exploded packages around for the foreseeable future. Overtime maybe even auth will be exploded ;)

setup.py
```diff
@@ -64,16 +64,21 @@
     extras_require['all'].extend(reqs)
 
 install_requires = [
-    'Invenio[base,auth,files]{}'.format(invenio_version),
+    'Invenio[base,auth]{}'.format(invenio_version),
+    'invenio-rdm-records>=0.9.0',
+    'invenio-records-permissions>=0.7.0',
```
fenekku: (I know these dependencies have only been moved.) Why was it that we didn't cap these version specifiers? (Given the approach we have taken to dependencies so far, we may want to...)

setup.py
ppanero: totally true, its hard to stay with bundles. Maybe when we are more stable we will manage.

setup.py
ppanero: True that. Fixed :)

PR #84, opened 2020-02-21
search: fix Searchbar => SearchBar

fenekku:
Fix `invenio webpack build` failing. That was tricky to spot!

Another release will be needed.


ppanero:
Got included in https://github.com/inveniosoftware/invenio-app-rdm/pull/85

PR #86, opened 2020-02-26
search: fix absolute search url

zzacharo:


invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchMain/SearchResults/SearchResults.js
```diff
@@ -79,10 +79,18 @@ class Results extends Component {
 }
 
 const OnResults = withState(Results);
+const OnEmptyResults = () => (
```
zzacharo: Created this because by default `react-searchkit/EmptyResults` component is exposing a button to reset the query. Although this works and clears the query from the url parameter, the searchbar value is not updated because there is no way currently to communicate the change from the searchkit app to the searchbar one. So, I just override the render and removed the button so people don't end up to this situation.

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/index.js
```diff
@@ -62,7 +62,8 @@ const resultsPerPageValues = [
 const logo = "/static/images/logo.svg";
 
 const searchApi = {
-  url: "https://localhost:5000/api/records/",
+  baseURL: "",
+  url: "/api/records",
```
zzacharo: Tested the `containerize` and it works ;)

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/SearchMain/SearchResults/SearchResults.js
fenekku: Got it

invenio_app_rdm/theme/assets/js/invenio_app_rdm/search/config/index.js
fenekku: nice!

Issue #88, opened 2020-02-26
Same header and footer as site in search page [semantic-ui transition]


fenekku:
(we have purposefully deferred this task, but we need to not forget it)

As a user navigating the site, I want to have a consistent header / footer to help orient myself.
As a developer, I want to have 1 place to set the header/footer and know it is applied on all pages, so I am more efficient and address the above. 

By default, header/footer should be the same across pages, right now it isn't.

Now we have:

record page

![image](https://user-images.githubusercontent.com/1932651/75364255-09ffb580-5881-11ea-80f0-7f46773c93ed.png)

search page

![image](https://user-images.githubusercontent.com/1932651/75364300-1f74df80-5881-11ea-9110-9ecc6858d399.png)

(frontpage is different but that's fine)

lnielsen:
This will be solved by migrating to Semantic UI.

PR #98, opened 2020-03-30
ui: make footer stick to the bottom

### LINKED ISSUE #81: Make sticky footer
The footer doesn't stick to the bottom of the screen if the screen content is too small, so it looks weird:

image

The footer should stick to the bottom in that case and should flow with the content normally if the content of the page is large enough.

See https://css-tricks.com/couple-takes-sticky-footer/

slint:
- Closes #81.

### DIFF
--- a/invenio_app_rdm/theme/assets/scss/invenio_app_rdm/footer.scss
+++ b/invenio_app_rdm/theme/assets/scss/invenio_app_rdm/footer.scss
@@ -8,6 +8,7 @@
 
  #footer {
    color: #fff;
+   margin-top: auto;
 
   &.footer-bottom {
     background-color: #0377cd;
--- a/invenio_app_rdm/theme/assets/scss/invenio_app_rdm/theme.scss
+++ b/invenio_app_rdm/theme/assets/scss/invenio_app_rdm/theme.scss
@@ -17,10 +17,12 @@
 
 html,
 body {
-  height: 100%;
-  width: 100%;
   font-family: 'Open Sans', sans-serif;
   color: #000;
+  width: 100%;
+  height: 100vh;
+  display: flex;
+  flex-direction: column;
 }
 h1,
 h2,


slint:
Hm, don't merge yet I've across some issues on some pages... I'll update it accordingly

invenio_app_rdm/theme/assets/scss/invenio_app_rdm/footer.scss
```diff
@@ -8,6 +8,8 @@
 
  #footer {
    color: #fff;
+   position: absolute;
```
fenekku: Hahaha I don't think that's enough ! (I didn't mean for the footer to always be visible at the bottom, but rather to stick it to the bottom if not enough vertical space to make it seem to be sticking to be at the bottom.

I've implemented it in https://github.com/galterlibrary/InvenioRDM-at-NU .

ppanero: So you mean showing at the bottom always? So even if there is content to scroll, the footer will be visible and then u just scroll the content itself no?

slint: Hehe, I looked at Zenodo's CSS and thought that was it... 1.5h hour later the ugly truth revealed itself.

I fixed it now though, tried it out on both Zenodo RDM and InvenioRDM and seems to work for all cases/pages.

I think some of the troubles (or at least solving them) came from the fact that some high-level CSS rules coming from Bootstrap and Semantic UI... But there are plans to fix this, so I'm optimistic :)

slint:
Forgot to add, example from working page: https://zenodo-rdm.web.cern.ch/search?q=cole&sort=bestmatch&order=desc&p=1&s=10&l=list

fenekku:
I guess #88 is making this not work on the search page. But we will deal with it later. Otherwise works. Thanks!

PR #100, opened 2020-04-13
prototype: initialize deposit page [more]

fenekku:
- add DepositMain react code
- load data from Flask into React component
- create a Jinja template for deposit page

This is a tracer bullet to see how things should
be connected.


fenekku:
Killing this in favour of #102

PR #113, opened 2020-05-07
add icon pattern to fields

fenekku:
this adds an icon to the existing field and introduces `props` for label icons going forward:

`<MyComponent label={"Label text"} labelIcon={"<semantic-ui icon name>"} ... />`

### DIFF
--- a/invenio_app_rdm/theme/assets/js/react_invenio_deposit/components/ResourceTypeField.js
+++ b/invenio_app_rdm/theme/assets/js/react_invenio_deposit/components/ResourceTypeField.js
@@ -10,6 +10,7 @@ import PropTypes from "prop-types";
 import { getIn, Field } from "formik";
 import { Form, Icon } from "semantic-ui-react";
 
+import { FieldLabel } from "../../react_invenio_forms";
 import { SelectField } from "../../react_invenio_forms";
 
 
@@ -37,7 +38,7 @@ export class ResourceTypeField extends Component {
     );
     return (
       <div>
-        <label><Icon name={this.props.labelIcon} />{this.props.label}</label>
+        <FieldLabel htmlFor={this.props.fieldPath} icon={this.props.labelIcon} label={this.props.label} />
         <SelectField
           fieldPath={typeFieldPath}
           label={this.props.typeLabel}
@@ -69,7 +70,7 @@ export class ResourceTypeField extends Component {
 
 ResourceTypeField.propTypes = {
   fieldPath: PropTypes.string.isRequired,
-  label: PropTypes.string.isRequired,
+  label: PropTypes.string,
   labelIcon: PropTypes.string,
   options: PropTypes.shape({
     type: PropTypes.arrayOf(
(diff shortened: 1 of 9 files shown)


invenio_app_rdm/theme/assets/js/react_invenio_deposit/components/ResourceTypeField.js
```diff
@@ -8,45 +8,49 @@
...
+import { Form, Icon } from "semantic-ui-react";
 
 import { SelectField } from "../../react_invenio_forms";
 
+
 export class ResourceTypeField extends Component {
   renderResourceTypeField = (formikBag) => {
-    const resource_type_fieldpath = `${this.props.fieldPath}.type`;
-    const resource_subtype_fieldpath = `${this.props.fieldPath}.subtype`;
+    const typeFieldPath = `${this.props.fieldPath}.type`;
```
zzacharo, line 18: Nice :)

invenio_app_rdm/theme/assets/js/react_invenio_deposit/components/ResourceTypeField.js
```diff
@@ -8,45 +8,49 @@
...
+      formikBag.form.setFieldValue(subtypeFieldPath, "");
     };
 
-    const subtype_options = this.props.options.subtype.filter(
+    const subtypeOptions = this.props.options.subtype.filter(
       (e) => e["parent-value"] === resource_type
     );
     return (
       <div>
+        <label><Icon name={this.props.labelIcon} />{this.props.label}</label>
```
zzacharo, line 40: @fenekku Here, it seems that the icon is a required prop but not defined as one in the propTypes section. Should we require the use of an icon, do you think? Is this coming directly from the vocabulary endpoint?

fenekku: The icon shouldn't be required. I wasn't getting errors when `labelIcon` wasn't passed, but I now I see that the `Icon` component still renders which takes whitespace/looks strange. I will make it not appear if `labelIcon` is not passed to look better :).

fenekku:
- only display Icon if an icon prop is passed.
- made label a non-required prop to make the component more flexible
- added `FieldLabel` as a reusable component in react-invenio-forms to take care of the label presentation logic

invenio_app_rdm/theme/assets/js/react_invenio_forms/core/FieldLabel.js
```diff
@@ -0,0 +1,25 @@
...
+    const {htmlFor, icon, label} = this.props;
+    return (
+      <label htmlFor={htmlFor}>{icon ? <Icon name={icon} /> : null}{label}</label>
+    )
+  }
+}
+
+FieldLabel.propTypes = {
+  htmlFor: PropTypes.string,
+  label: PropTypes.string,
```
zzacharo, line 23: Just mark the props that are required here, e.g `label`, `htmlFor`. Also a nice trick for optional props that you want to give a default value is this https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values

fenekku: I thought `propTypes` was a nice way to define the interface for the component, but it should only be used for required props? So if a prop is not required, it should not be placed in the `propTypes`but rather only be placed in the `defaultProps`? Did I get this right?

What if there is no sensible default?: for icon there isn't a default - I guess `null` or `undefined` could be in the default, but kind of implied...

fenekku:
Closed via #114

PR #117, opened 2020-05-15
[follows from #114] Fix initial null controlled value

fenekku:
Closes https://github.com/inveniosoftware/react-invenio-deposit/issues/2


run-tests.sh
```diff
@@ -16,4 +16,4 @@ pydocstyle invenio_app_rdm tests docs && \
 isort -rc -c -df && \
 check-manifest --ignore ".travis-*" && \
 sphinx-build -qnNW docs docs/_build/html && \
-python setup.py test
+pytest
```
lnielsen, line 19: Comment: What prompted this change?

fenekku: Silly me. Ugh I've only put the explanation: https://discordapp.com/channels/692989811736182844/704629170470125578/708438917724241922
in discord but I thought I had put it here too :duck: 

This was explained in video chat, but for reference: python setup.py test:

- attempts to install test dependencies which can create incompatibilities
- is deprecated from setuptools

PR #128, opened 2020-05-19
deposit: refactor publish flow

### LINKED ISSUE #116: Frontend Publish Flow w/ current REST API
Implement frontend publish flow using the current REST API as defined in invenio-rdm-records (not the draft API yet)

- [ ] Prepare payload
- [ ] RDMAPIClient send
- [ ] Dealt with response: 
  if all good: redirect to record landing page
  else: error messages

zzacharo:
closes #116

### DIFF
--- a/invenio_app_rdm/theme/assets/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
@@ -36,7 +36,6 @@ class RecordPreviewer extends Component {
 }
 RecordPreviewer = connect(RecordPreviewer);
 
-
 export class FormRecordPreviewer extends Component {
   renderFormField = (formikBag) => {
     const values = formikBag.form.values;
@@ -50,10 +49,7 @@ export class FormRecordPreviewer extends Component {
 
   render() {
     return (
-      <Field
-        name="FormRecordPreviewer"
-        component={this.renderFormField}
-      />
+      <Field name="FormRecordPreviewer" component={this.renderFormField} />
     );
   }
 }
@@ -64,12 +60,7 @@ function fakeInitialRecord(backendRecord) {
    */
   // Experiment: Ignore backend for now and gradually fill what the frontend
   //             needs for a brand new record
-  const {
-    titles,
-    creators,
-    contributors,
-    ...fakedRecord
-  } = backendRecord;
+  const { creators, contributors, ...fakedRecord } = backendRecord;
   return fakedRecord;
 }
 
@@ -81,8 +72,8 @@ const defaultRecord = {
     {
       lang: "",
       title: "",
-      type: ""
-    }
+      type: "",
+    },
   ],
   creators: [
     {
@@ -91,19 +82,19 @@ const defaultRecord = {
           identifier: "",
           name: "",
           scheme: "",
-        }
+        },
       ],
       given_name: "",
       family_name: "",
       name: "",
-      type: "personal",
+      type: "Personal",
       identifiers: [
         {
           identifier: "",
           scheme: "",
-        }
-      ]
-    }
+        },
+      ],
+    },
   ],
   contributors: [
     {
@@ -112,22 +103,22 @@ const defaultRecord = {
           identifier: "",
           name: "",
           scheme: "",
-        }
+        },
       ],
       given_name: "",
       family_name: "",
       name: "",
-      type: "personal",
+      type: "Personal",
       identifiers: [
         {
           identifier: "",
           scheme: "",
-        }
+        },
       ],
-      role: ""
-    }
-  ]
-}
+      role: "",
+    },
+  ],
+};
 
 // NOTE: RDMDepositForm knows the meaning associated to the field.
 // i.e. it knows its datamodel is the one sent by invenio-rdm-records.
@@ -146,10 +137,16 @@ export class RDMDepositForm extends Component {
     this.config = props.config || {};
     this.controller = new RDMDepositController(new RDMDepositApiClient());
     // TODO: Remove when backend is better integrated
-    console.log("backend initial record", JSON.stringify(props.record, null, 2))
-    console.log("faked backend record", JSON.stringify(fakeInitialRecord(props.record), null, 2))
+    console.log(
+      "backend initial record",
+      JSON.stringify(props.record, null, 2)
+    );
+    console.log(
+      "faked backend record",
+      JSON.stringify(fakeInitialRecord(props.record), null, 2)
+    );
     const record = defaultize(fakeInitialRecord(props.record));
-    console.log("initial form record", JSON.stringify(record, null, 2))
+    console.log("initial form record", JSON.stringify(record, null, 2));
     this.state = {
       record: record || {},
     };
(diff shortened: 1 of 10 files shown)


invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositController.js
```diff
@@ -20,37 +24,48 @@ export class DepositController {
...
         window.history.replaceState(undefined, "", newURL);
       }
+      const response = await this.apiClient.save(payload.data);
+      store.dispatch({
+        type: SAVE_SUCCESS,
+        payload: response,
+      });
+      formik.setSubmitting(false);
+    } catch (error) {
+      store.dispatch(setFormErrorsFromResponse(error, formik));
```
fenekku, line 45: Normal:

1- We should be able to save a network call (when save is a network call) by doing:
```js
const response = null;
if (!this.exists(record)) {
  response = await this.apiClient.create(payload);
  // ...
} else {
  response = await this.apiClient.save(payload.data);
}
```

because we already pass the record in `create` .

2- the try-catch allows us to catch network errors, but not invalid metadata errors, right? We still need to parse the response to determine if there were errors and dispatch SAVE_SUCCESS accordingly.

3- Since we only deal with publish flow (in this issue and for the may release), we could just disable all the save section and focus on the publish section below only.

invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositController.js
```diff
@@ -20,37 +24,48 @@ export class DepositController {
     console.log("Validate record", record);
   }
 
-  async save_draft(record) {
+  async save_draft(record, { formik, store }) {
     // Saves a draft of the record
-    let payload = record;
+    const recordSerializer = store.config.recordSerializer;
+    let payload = recordSerializer.deserialize(record);
```
fenekku, line 30: Hmm the terms `deserialize` and `serialize` are perhaps confusing. I associate serialization to transforming an input into a format that can be sent over the network: frontend memory representation to represenation the backend can deserialize. See https://stackoverflow.com/questions/633402/what-is-serialization . So I would have expected this to be `serialize`. If it's too overloaded, we can also use:

serialize === marshal === dump
deserialize === unmarshal === load  

And if we need to convert from a json representation to another one (say more compatible with the expectations of the front-end form library), well, `convertToOtherOne` where `convert` is the action word is probably best. Synonyms like `transform`, `represent`, `reshape`... are good too.

invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositController.js
```diff
@@ -20,37 +24,48 @@ export class DepositController {
...
       }
+      const response = await this.apiClient.publish(payload.data);
+      store.dispatch({
+        type: PUBLISH_SUCCESS,
+        payload: response,
+      });
+      formik.setSubmitting(false);
+      window.location.replace(response.data.links.self.split("/api")[1]);
+    } catch (error) {
+      store.dispatch(setFormErrorsFromResponse(error, formik));
```
fenekku, line 68: same as points 1- and 2- above.

invenio_app_rdm/theme/assets/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
```diff
@@ -90,19 +81,19 @@ const defaultRecord = {
...
           name: "",
           scheme: "",
-        }
+        },
       ],
       given_name: "",
       family_name: "",
       name: "",
-      type: "personal",
+      type: "Personal",
```
fenekku, line 89: thanks for catching those! How did I miss them :thinking: ...

invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositRecordSerializer.js
```diff
@@ -38,6 +38,45 @@ export class DepositRecordSerializer {
   }
 
   deserialize(record) {
```
fenekku, line 40: Per remark above - rename to `serialize`

invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositRecordSerializer.js
```diff
@@ -38,6 +38,45 @@ export class DepositRecordSerializer {
...
-    return this.stripNullEquivalentFields(record);
+    let res = this.stripNullEquivalentFields(record);
```
fenekku, line 41: minor: `res` -> `stripped_record`

invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositRecordSerializer.js
```diff
@@ -38,6 +38,45 @@ export class DepositRecordSerializer {
...
+          type: "MainTitle",
+          title: res["titles"] ? res["titles"][0]["title"] : "",
+        },
+      ],
+      // TODO: Remove this when we fix the `Identifiers` schema
+      creators: [],
+      contributors: [],
+      // TODO: Remove these when fields are implemented
+      // also these fields are making the record landing page
+      // to fail if they don't exist
```
fenekku, line 64: :+1: Yes. We should be able to gradually clean this up before the may release

invenio_app_rdm/theme/assets/js/react_invenio_deposit/state/actions/deposit.js
```diff
@@ -28,41 +22,20 @@ export const setFormErrorsFromResponse = (response, formik) => {
...
-        payload: response,
-      });
-      formik.setSubmitting(false);
-    } catch (error) {
-      console.log("error", error);
-      dispatch(setFormErrorsFromResponse(error, formik));
-    }
+    controller.save_draft(record, {
+      formik,
+      store: { dispatch, getState, config },
```
fenekku, line 37: nice!

invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositController.js
zzacharo: @fenekku If the call resolves to a ValidationError then we catch it but if the response is 200, then yes we need to parse it. But shouldn't we avoid this?

zzacharo: @fenekku 
1- I don't see how we can save a network call as if the record doesn't exist we need to create it first before we save it. Currently, because we are using the `/records` API the save doesn't exist so we mock it but this will not happen when we integrate the `/drafts` API.
2- The `try/catch` is working properly if the network requests leads to an error response code. `ValidationError` gives a 400 so we are able to catch it.
3- I will disable in the UI the `save` button!

invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositApiClient.js
```diff
@@ -33,7 +28,7 @@ export class DepositApiClient {
     return new Promise((resolve, reject) => {
       setTimeout(() => {
         console.log("Record saved", record);
-        resolve({data: record});
+        resolve({ data: record });
```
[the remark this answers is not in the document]
zzacharo: Yes prettier. I set it up in the `react-invenio-forms` and `react-invenio-deposit` repositories so you can install it in VScode and will pick it up the config!

invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositController.js
zzacharo: I went for inverting `serialize`/`deserialize`. We can discuss in person if we need to rename the 2 methods :) Thanks for the detailed explanation on this :)

invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositApiClient.js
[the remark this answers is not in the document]
fenekku: Thanks!

invenio_app_rdm/theme/assets/js/react_invenio_deposit/DepositController.js
fenekku: 1- I interpreted https://codimd.web.cern.ch/w5jnQsJ3Sn6Ne5LLEQD0hg#Create-draft-initial-Deposit-page-Auto-save as: the API accepts creation of a draft by passing form values as initial values directly (no need to create something empty first). This applies because we are in the context of the draft (Save). In the context of publishing, we indeed need to make sure a record exists before. We would create it (with the values from the form as initial values) and then publish (no need to "save" it after it was created though). 

2- wow axios does this for us?! That's nice.

3- :heavy_check_mark:

zzacharo: @fenekku sorry I missed that context :) Yes in that case we can change a bit the workflow so we can save one request :)

Issue #125, opened 2020-05-19
Migrate templates to semantic-ui


ppanero:
Follow this [guide](https://codimd.web.cern.ch/bfDyh_ejQJSq-DZpGCSfMA)

PR #129, opened 2020-05-20
Semantic

ppanero:
Changes:
- Jinja templates div structure and classes according to https://codimd.web.cern.ch/bfDyh_ejQJSq-DZpGCSfMA and semanti-ui docs.
- Uses the main register form. Overwrites background using `cover-page` class.

Current status (minor UI issues signalled in red)
Screenshot 2020-05-22 at 13 53 16

Search page is as it was before (Might need some migration, since it is using the old react+semantic code)
Screenshot 2020-05-22 at 14 33 52

Screenshot 2020-05-22 at 16 37 07

Screenshot 2020-05-22 at 16 37 17

Issues encountered:

- The frontpage header is divided in two: a `nav` and a `div` (this last one coming from `header_frontpage.html`). Since the only element wrapping them is the `<header>`, changing background changes it for much more than the desirable space. Solution: overwrite both templates, because there is no block encapsulating these two.
- In order to simply change the sentence "Sign up for **a** InvenioRDM account" to "Sign up for **an** InvenioRDM account" the whole template needs to be overwritten.
Screenshot 2020-05-22 at 16 37 07
- The color of the button seems to be picked but for some reason is not changed:
Screenshot 2020-05-22 at 13 52 40

closes https://github.com/inveniosoftware/invenio-app-rdm/issues/125


ppanero:
I would vote for merging this and then try to extract things to `invenio-theme`. Mainly due to time constraints.

invenio_app_rdm/theme/templates/invenio_app_rdm/header.html
```diff
@@ -0,0 +1,85 @@
+{# -*- coding: utf-8 -*-
```
fenekku, line 1: minor: 2 space indentation. If you are using VSCode: https://github.com/editorconfig/editorconfig-vscode is your friend.

invenio_app_rdm/theme/templates/invenio_app_rdm/search/header.html
```diff
@@ -12,38 +12,36 @@
...
+          <div class="ui container">
+            <div class="ui buttons">
+              <a class="ui button" role="button" href="/deposits">Uploads</a><a class="ui button" role="button"
+                href="/coming-soon">Communities</a>
             </div>
+          </div>
+        </div>
         {% endblock %}
       </div>
     </div>
```
fenekku, line 47: [potential shelf]: Most of this header can now be the same as the header for the rest of the site. The only difference should be the searchbar itself which would be a react app... although ideally it would be a react app everywhere and only on the search page would it have a slightly different behaviour.

invenio_app_rdm/theme/webpack.py
```diff
@@ -20,39 +18,44 @@ def theme():
...
-            "semantic-ui-calendar-react": "^0.15.3",
-            "semantic-ui-css": "^2.4.1",
-            "semantic-ui-react": "^0.88.0"
+        default='semantic-ui',
+        themes={
+            'semantic-ui': dict(
+                entry={
+                    'invenio-app-rdm-theme': current_app.config['INSTANCE_THEME_FILE'],
+                    'invenio-app-rdm-search-theme': current_app.config[
+                        'INSTANCE_SEARCH_THEME_FILE'],
```
fenekku, line 30: very minor:
```suggestion
                    'invenio-app-rdm-search-theme': (
                        current_app.config['INSTANCE_SEARCH_THEME_FILE']
                    ),
```

invenio_app_rdm/theme/webpack.py
```diff
@@ -20,39 +18,44 @@ def theme():
...
+                    'invenio-app-rdm-search-js':
+                        './js/invenio_app_rdm/search/SearchMain/index.js',
+                    'invenio-app-rdm-searchbar-js':
+                        './js/invenio_app_rdm/search/SearchBar/index.js',
+                    'invenio-app-rdm-deposits-deposit-searchbar-js':
+                        './js/invenio_app_rdm/deposits/RDMDepositSearchBar/index.js',
```
fenekku, line 36: PEP8 will complain about these being too long. We can add `# noqa` at the end of the line.

invenio_app_rdm/theme/templates/invenio_app_rdm/search/header.html
ppanero: Yes, I agree with this. I just wanted to wait till deposit is cleaned from the folder. Then clean search. Then see how we can only have one header, potentially some things should go to invenio-assets/theme. We are needed too much overwriting atm. IMHO.

ppanero:
Update:

- Login button coloring errors seems to be a known issue in safari. Same happens to the frontpage search bar button margins. It works in Firefox/Chrome:

Screenshot 2020-05-25 at 15 10 50

- Sign up template text, I would personally go for a PR to invenio-theme... It is quite a simple customization.

invenio_app_rdm/theme/templates/invenio_app_rdm/search/header.html
fenekku: > We are needed too much overwriting atm. IMHO

yes.

I hope we don't get ourselves too muddied up with all of our temporary fixes...

fenekku:
agreed with the pr to invenio-theme

PR #158, opened 2020-05-29
signup: make signup button colour consistent

### LINKED ISSUE #144: semantic-ui: inconsistent signup button
Frontpage and signup page button do not have same color nor logo:
Screenshot 2020-05-28 at 15 59 10

zzacharo:
closes #144

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme.less
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme.less
@@ -37,8 +37,8 @@ h6 {
   border-color: transparent;
 
   .button[href^="/signup"] {
-    background-color: #44a244;
-    border-color: #44a244;
+    background-color: @signup_background_color;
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme.less
```diff
@@ -202,8 +202,8 @@ h6 {
 }
 
 .btn-warning {
```
ppanero, line 204: minor: we should be able to change this name:
- Semantic uses `button` instead of `btn`.
- It is not actually a "warning button". Could be "sign-up-button" class or something.

I'd say go ahead, merge and then open an issue to revist this in the future.

zzacharo: I opened an issue https://github.com/inveniosoftware/invenio-app-rdm/issues/160

PR #163, opened 2020-05-29
assets: fix header spacing

ppanero:
It is due to the `outer_nav` that was placed in the header overwrite. Therefore it is RDM specific. 

*Need more testing: I am on the process of doing so, my setup just broke, need to re-install from a fresh state.*


ppanero:
Closed in favour of #164

PR #174, opened 2020-06-01
theme: fix double importing of semantic ui theming

lnielsen:
DO NOT MERGE YET

 - Fixes a problem were the RDM and instane CSS would be included twice
   due to the Semantic UI theming. Depends on
   inveniosoftware/cookiecutter-invenio-rdm#85. We no longer need a
   dynamically generated theme bundle and we no longer need
   INSTANCE_THEME_FILE. This role is taken over by theme.config which
   is provided by cookiecutter-invenio-rdm.

- Removes use of the external font to not make external requests.

- Simplifies templates to use just Invenio-Theme files.


invenio_app_rdm/theme/webpack.py
```diff
@@ -12,42 +12,37 @@
...
-        }
-    )
+theme = WebpackThemeBundle(
+    __name__,
+    'assets',
+    default='semantic-ui',
+    themes={
+        'semantic-ui': dict(
+            entry={
+                'invenio-app-rdm-deposits-deposit-searchbar-js':
```
zzacharo, line 22: @lnielsen I think this can be deleted too, as now we use the main `header.html`. You can delete all occurrences of this bundle + plus the code.

PR #194, opened 2020-06-09
semantic-ui: namespace templates

### LINKED ISSUE #165: semantic-ui: templates should be nested inside "semantic-ui/"
This way, in the generated cookiecutter, one can be explicit and do:

```jinja
{%- extends 'semantic-ui/invenio_app_rdm/frontpage.html' %}
...
```

Also without the namespacing, this seems like this module provides by default Bootstrap3 templates (i.e. when no `APP_THEME` is set at all).

slint:
* Closes #165.


ppanero:
Don't the paths need to be updated in the config.py? Or would is it now working due to the fallback? 

From what I recall in documentation at migration time:

- Assets need to be `semantic-ui` prefixed (folder)
- Templates don't (Maybe because of the fall back?)

slint:
When templates are not prefixed, you basically rely on having `APP_THEME` not being set at all. Since in `invenio-app-rdm` though we're setting `APP_THEME = ['semantic-ui']` explicitly, I think it should be fine. 

This also goes well with the practice of being always explicit, and allows extending specific templates in an overlay:

https://github.com/zenodo/zenodo-rdm/blob/e1b4b37a6deaac1a0d2746a3f628603834143c0c/templates/zenodo/invenio_app_rdm/frontpage.html#L1

slint:
> It might be better to use the config vars, in case the user sets them it will actually make customization useful otherwise they'll need to overwrite the full template.

Agree and applied the suggestions.

There's a bigger topic here to discuss, since now there are three ways to override templates:

- Use config variables in `render_template`/`extends`/`include`.
  - This requires creating a new template in your instance and then changing also the respective config variable. It has the benefit of being very explicit (i.e. no magic names)
- Add your instance theme, like `APP_THEME = ['zenodo', 'semantic-ui']`, and then create the correct template filename
  - This requires a bit of "mental gymnastics" until you understand the templating lookup logic... But it can be very powerful in the sense that you can easily override any template you want with little effort, no matter if there are config variables defined in the original template to assist you.
-  In your instance's `templates` folder (the one that Flask/Jinja looks inside for the lookup), you can create a template with the same path/filename as the one you want to override
  - This is a bit "magic" overriding, since you rely on the fact that there is a special folder that gets taken into account before everything else

ppanero:
After discussing IRL, using config vars in the `frontpage.html` is not possible since `THEME_FRONTPAGE_TEMPLATE = 'invenio_app_rdm/frontpage.html'` and then there is a loop. This goes a long with the comment from @slint above. Maybe `invenio-theme` should have some sort of "BASE" templates.

PR #216, opened 2020-07-07
Frontpage translation

baldemir:
* marked strings for translation on frontpage.html and  header_frontpage.html
* changed transifex config file
* recreated .pot file

### DIFF
--- a/invenio_app_rdm/translations/messages.pot
+++ b/invenio_app_rdm/translations/messages.pot
@@ -1,57 +1,108 @@
 # Translations template for invenio-app-rdm.
-# Copyright (C) 2019 CERN
+# Copyright (C) 2020 CERN
 # This file is distributed under the same license as the invenio-app-rdm
 # project.
-# FIRST AUTHOR <EMAIL@ADDRESS>, 2019.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2020.
 #
 #, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: invenio-app-rdm 1.0.0a5\n"
+"Project-Id-Version: invenio-app-rdm 0.10.1\n"
 "Report-Msgid-Bugs-To: info@inveniosoftware.org\n"
-"POT-Creation-Date: 2019-12-19 08:33-0600\n"
+"POT-Creation-Date: 2020-07-06 13:24+0300\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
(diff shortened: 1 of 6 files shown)


invenio_app_rdm/translations/messages.pot
```diff
@@ -1,57 +1,108 @@
...
+msgstr ""
+
+#: invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/frontpage.html:29
+msgid "Customize"
+msgstr ""
+
+#: invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/frontpage.html:32
+#, python-format
+msgid ""
+"Customize\n"
```
zzacharo, line 66: @lnielsen do you think of a way to avoid all these newline symbols injected in the collected string?

baldemir:
> Very nice job! I have a minor comment but other than that it looks good to me! Have you tested also the transifex workflow i.e push/pull to/from transifex ?

Yes @zzacharo I have translated the last entry in Turkish translation file via Transifex workflow.

invenio_app_rdm/translations/messages.pot
```diff
@@ -1,57 +1,108 @@
 # Translations template for invenio-app-rdm.
-# Copyright (C) 2019 CERN
+# Copyright (C) 2020 CERN
```
fenekku, line 2: minor: you can add your institution's copyright here @baldemir since it is thanks to you we have translations setup correctly.

invenio_app_rdm/translations/messages.pot
fenekku: +1 yeah it's weird and seems fragile.

When working with transifex and translations in the past we had a similar situation without clear solution...

Another take is the Android approach: every string to be translated is just a placeholder/symbol. It makes it harder to understand the text in context (i.e. in the frontpage you would see `_("frontpage_step2_customize")` instead of this full text), but then changes in text are less brittle. I also suspect there might be vscode extensions that can translate/expand back in your editor.

invenio_app_rdm/translations/tr/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,113 @@
+# Translations template for invenio-app-rdm.
+# Copyright (C) 2020 CERN
+# This file is distributed under the same license as the invenio-app-rdm
```
fenekku, line 3: copyright here for your institution

PR #220, opened 2020-07-24
tests: test read flow using resources module

zzacharo:


tests/api/test_record_api.py
```diff
@@ -7,3 +7,131 @@
...
+
+        provider = CustomRecordIdProvider.create(
+            object_type='rec', object_uuid=record_uuid)
+        data['recid'] = provider.pid.pid_value
+        return provider.pid
+
+    current_pidstore.minters['recid_v2'] = custom_minter
+    yield app
+
+    current_pidstore.minters['recid_v2'] = recid_minter
```
fenekku, line 43: Hopefully we should soon be able to forego this!

setup.py
```diff
@@ -51,6 +51,7 @@
     ],
     # Extras
     'docs': [
+        'docutils>=0.10,<0.16',
```
ppanero, line 54: curiosity: is this fixing some issue that might appear in other CIs?

PR #222, opened 2020-07-29
resources: integrate drafts resource

ppanero:
requires https://github.com/inveniosoftware/invenio-rdm-records/pull/139

tests: 
- removed `test_record_read_unregistered_pid` since it is the same than create a draft and read it (i.e. the PID is still not registered since it is a draft).


tests/api/test_record_api.py
```diff
@@ -135,3 +192,24 @@ def test_read_deleted_record(client, location, minimal_record, users):
     response = client.get(SINGLE_RECORD_API_URL.format(recid), headers=HEADERS)
     assert response.status_code == 410
     assert response.json['message'] == "The record has been deleted."
+
+
+def test_record_search(client):
+    """Test draft creation."""
```
fenekku, line 198:
```suggestion
    """Test record search."""
```

PR #240, opened 2020-08-19
assets: dynamic import customized components

### LINKED ISSUE #234: global: use webpack dynamic imports to customize search pages
https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import

mvidalgarcia:
closes #234

Co-authored-by: Parth Shandilya <parth.shandilya@cern.ch>

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search_app_customizations.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search_app_customizations.js
@@ -5,83 +5,36 @@
 // Invenio App RDM is free software; you can redistribute it and/or modify it
 // under the terms of the MIT License; see LICENSE file for more details.
 
-import React from "react";
 import { overrideStore } from "react-overridable";
-import { Card, Input, Item } from "semantic-ui-react";
 import _ from "lodash";
 import _truncate from "lodash/truncate";
+import {
+  RDMRecordResultsListItem,
+  RDMRecordResultsGridItem,
+  RDMRecordSearchBarElement,
+} from "./search";
 
-const RDMRecordResultsListItem = ({ result, index }) => {
-  const description = _.get(
-    result,
-    "metadata.descriptions[0].description",
-    "No description"
-  );
+async function registerComponent(componentId, defaultComponent) {
+  let component;
+  try {
+    const module = await import(
+      /* webpackMode: "eager" */
+      `../../templates/search/${componentId}.jsx`
+    );
+    component = module.default;
+  } catch (error) {
+    component = defaultComponent;
+  } finally {
+    overrideStore.add(componentId, component);
+  }
+}
 
-  return (
-    <Item key={index} href={`/records/${result.id}`}>
-      <Item.Content>
-        <Item.Header>{result.metadata.titles[0].title}</Item.Header>
-        <Item.Description>
-          {_truncate(description, { length: 200 })}
-        </Item.Description>
-      </Item.Content>
-    </Item>
-  );
+const componentMapping = {
+  "ResultsList.item": RDMRecordResultsListItem,
+  "ResultsGrid.item": RDMRecordResultsGridItem,
+  "SearchBar.element": RDMRecordSearchBarElement,
 };
 
-overrideStore.add("ResultsList.item", RDMRecordResultsListItem);
-
-const RDMRecordResultsGridItem = ({ result, index }) => {
-  const description = _.get(
-    result,
-    "metadata.descriptions[0].description",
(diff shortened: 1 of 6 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/RDMRecordResultsGridItem.js
```diff
@@ -0,0 +1,31 @@
+// This file is part of InvenioRDM
```
lnielsen, line 1: Does the filename match the name which would be used by a user in ``my-site/assets/templates/search/<filename>``?

lnielsen (commented):
Looks great!

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/RDMRecordResultsGridItem.js
mvidalgarcia: Nope. Fixed now :)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search_app_customizations.js
```diff
@@ -5,83 +5,36 @@
...
+  RDMRecordResultsGridItem,
+  RDMRecordSearchBarElement,
+} from "./search";
+
+async function registerComponent(componentId, defaultComponent) {
+  let component;
+  try {
+    const module = await import(
+      /* webpackMode: "eager" */
+      `../../templates/search/${componentId}.jsx`
```
slint, line 22: **Minor:** There used to be (maybe still is) a `'@templates': 'assets/templates'` alias in the webpack config, that might help with easily accessing the templates

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search_app_customizations.js
```diff
@@ -5,83 +5,36 @@
...
+    component = defaultComponent;
+  } finally {
+    overrideStore.add(componentId, component);
+  }
+}
+
+const componentMapping = {
+  "ResultsList.item": RDMRecordResultsListItem,
+  "ResultsGrid.item": RDMRecordResultsGridItem,
+  "SearchBar.element": RDMRecordSearchBarElement,
```
slint, line 35: **Minor:** I find myself also in the need of defining such a listing to define default components in `invenio-search-ui`... Probably a full listing can start in `invenio-search-ui`, and then later get overwritten on the "module-level" by e.g. `invenio-app-rdm`, `invenio-communities`, etc. for the various search apps.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search_app_customizations.js
```diff
@@ -5,83 +5,36 @@
...
 };
 
-overrideStore.add("SearchBar.element", RDMRecordSearchBarElement);
+for (const [componentId, component] of Object.entries(componentMapping)) {
+  registerComponent(componentId, component);
```
slint, line 39: **Moderate:** There's a bit of a coordination issue here (probably what @ntarocco was worried about) when it comes to the complexity that a Promise/async-based solution would bring into the mix: since the imports are done asynchronously (although from the Webpack docs, the `eager` option, seems to change things in practice), the SearchApp can only be initialized once all of these promises are finished... This of course is an issue, since the initialization happens in a different JS file, so there's no `Promise.all(...)` that can be used reliably without some weird hack/convention (e.g. pushing promises in a `window.searchAppComponentRegistrations` array).

One way would be to actually do all of the component registrations in the JS code that does the search app initialization (but let's discuss that).

slint (commented):
LGTM, just left some comments here as discussion points for IRL

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search_app_customizations.js
mvidalgarcia: I think it [used to be there](https://github.com/inveniosoftware/invenio-assets/commit/51e696c774782f2500062380bcde06f1cad52181) but now it's importing the aliases from `./config`. Do we have to create a config file?

slint:
Changes introduced by this PR are included in https://github.com/inveniosoftware/invenio-theme/pull/201, https://github.com/inveniosoftware/invenio-search-ui/pull/96, and #260

PR #274, opened 2020-09-03
blueprint: make UI urls configurable

### LINKED ISSUE #232: Make UI URLs configurable
**NOTE**: This task has been much clarified over time so its definition has been rewritten to make it evident what it's about now

Make invenio-app-rdm UI URLs configurable via config.py variables. This then means percolating that knowledge across the codebase/modules using these same config variables.

- [x] in invenio-app-rdm
- [x] in invenio-records-resources
- [x] in react-invenio-deposit

We can use a callable blueprint. This way a ResourceConfig can be defined in the config.py and fetched with an application context because the application is created at time of blueprint registration.

The default URL values should be:

-    /uploads - List user records
-    /uploads/new - Create a new record (prior to having an ID, as soon as we have an ID by pressing save, the React app rewrites the URL to the one below).
-    /uploads/:id - Editable (draft) records
-    /search - Main search result page.
-    /records/:id - A published record

fenekku:
closes #232

### DIFF
--- a/invenio_app_rdm/theme/views.py
+++ b/invenio_app_rdm/theme/views.py
@@ -23,103 +23,108 @@ from invenio_rdm_records.resources import BibliographicDraftActionResource, \
 from invenio_rdm_records.services import BibliographicRecordService
 from invenio_rdm_records.vocabularies import Vocabularies
 
-blueprint = Blueprint(
-    'invenio_app_rdm',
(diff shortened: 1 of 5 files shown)


fenekku:
Some updates are still needed to make it pass the URL(s) to the react application.

PR #277, opened 2020-09-08
added isort installation

cenouralm:


setup.py
```diff
@@ -20,6 +20,7 @@
 
 tests_require = [
     'pytest-invenio>=1.3.2',
+    'isort>=4.2.2',
```
fenekku, line 23: pytest-invenio provides `isort` already... What was the issue here?

cenouralm: I was trying to run "./run-tests" but it kept saying "isort: command not found". After I added this line and running " pip install -e .[all]" it worked fine.

cenouralm: Just tested and it worked only running `pip install -e .[all]`.

fenekku: Ok. Closing.

PR #298, opened 2020-10-08
deposit: fixes

zzacharo:


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
```diff
@@ -101,16 +92,14 @@ const defaultRecord = {
 // However, it is RDMDepositForm that gets to decide what are the frontend
 // defaults for this datamodel
 function defaultize(initialRecord) {
-  let record = { ...defaultRecord };
-  record.metadata = { ...record.metadata, ...initialRecord.metadata };
-  return record;
+  return { ...defaultRecord, ...initialRecord };
```
fenekku, line 95: I am not sure this will work because the defaults from the backend for titles / creators / contributors were not right and it was easier to fix them on the frontend. This will take the backend ones... But things may have changed since!

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
```diff
@@ -40,6 +30,7 @@ const defaultRecord = {
     metadata_restricted: false,
     files_restricted: false,
     owners: [1],
+    access_right: "open",
```
fenekku, line 33: I removed that one because it didn't show up as selected (but was there) and so when the payload was sent it wasn't clear what value was being sent. Not providing it also allows for the error to show up.

zzacharo: Fixed that here https://github.com/inveniosoftware/react-invenio-deposit/commit/8cb590fb23185f537517c634f1866c8e91b93309 . It's a temporary solution as the `GroupField` component should pass down formik props :)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
zzacharo: We tested it and it seems working! What was before there, was causing problems...we should definitely remove all these transformations....

PR #320, opened 2020-10-19
search: show localized pub date

zzacharo:
Depends on https://github.com/inveniosoftware/invenio-rdm-records/pull/211
Depends on https://github.com/inveniosoftware/marshmallow-utils/pull/18

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
@@ -19,8 +19,8 @@ export const RDMRecordResultsListItem = ({ result, index }) => {
   );
   const publicationDate = _get(
     result,
-    "metadata.publication_date",
-    "No metadata"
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
```diff
@@ -19,8 +19,8 @@ export const RDMRecordResultsListItem = ({ result, index }) => {
   );
   const publicationDate = _get(
     result,
-    "metadata.publication_date",
-    "No metadata"
+    "ui.publication_date_l10n.value",
```
lnielsen, line 22: Question: Shouldn't this just be ``ui.publication_date_l10n``?

zzacharo: Yes I just kept the dictionary structure but I can change it yes :) But, wouldn't it better just to take from the `metadata.publication_date_l10n` instead of serializing in the `ui` object?

PR #345, opened 2020-10-22
release: v0.16.0

zzacharo:


fenekku:
Why the switch to 0.16.0? 0.15.1 would have been fine... 

We can wait for Lars' return for a larger discussion about versions, but my thoughts are:

- for each monthly release we bump changing modules 1 minor version up and subsequent changes are just patches. In this case, the module was bumped to 0.15.0 6 days ago, so anything else is just a patch.
- Isn't this ignoring semantic versioning and the notion of breaking change? Actually no. Everything goes before 1.0 with semantic versioning: https://semver.org/#spec-item-4
- Then what is the purpose or why is bumping minor all the time such a big deal? Not a "super" big deal, but it
   * gets high numbers very fast for no reason and people have negative biases towards these high numbers
   * makes it harder for the team (that's my take on it :stuck_out_tongue_winking_eye: ) to understand which version we are at and when to bump version (almost everything is a breaking change now, so hard to say what is not...)
   * allows us to say: we will release these versions this month with no ambiguity

Most of these are just about having a system in place for versioning though: another system would be good as well as long as it's clear.

fenekku (commented):
See comment above

PR #352, opened 2020-10-28
search: make aggregation buckets collapsible

Glignos:


Glignos:
Screenshot 2020-10-28 at 15 56 41

PR #361, opened 2020-10-30
add tests through gh actions

ppanero:


run-tests.sh
```diff
@@ -12,8 +12,13 @@
...
 # released.
 
-docker-services-cli up postgresql es redis
-python -m check_manifest --ignore ".travis-*" && \
+# Quit on errors
+set -o errexit
+
+# Quit on unbound symbols
+set -o nounset
+
```
fenekku, line 20: Could `set -e` be enough? Why was this added now?

fenekku (approved):
Just wandering about those `set` calls... :thinking:

run-tests.sh
ppanero: `set -e` would replace the first, `set -u` would replace the second. I was just trying to follow what was agreed on expanding command-line arguments to be more explicit.

Why? Because it was not enough as it was. For example, if the manifest check was wrong it would continue to run tests. We want it to fail fast and stop in the first command that errors.

fenekku: ok!

PR #376, opened 2020-11-06
deposit: add FundingField

fenekku:
- closes https://github.com/inveniosoftware/react-invenio-deposit/issues/16
- requires: https://github.com/inveniosoftware/react-invenio-deposit/pull/80

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
@@ -15,6 +15,7 @@ import {
   DatesField,
   DepositFormApp,
   DescriptionsField,
+  FundingField,
   IdentifiersField,
   LanguagesField,
   PublishButton,
@@ -113,6 +114,49 @@ export class RDMDepositForm extends Component {
             { text: "Other", value: "other" },
           ],
         },
+
+        // TODO: Replace with an API backend
+        funding: {
+          funder: [
+            {
+              name: 'National Institutes of Health (US)',
+              identifier: 'funder1',
+              scheme: 'funderScheme1',
+            },
+            {
+              name: 'European Commission (EU)',
+              identifier: 'funder2',
+              scheme: 'funderScheme2',
+            },
+          ],
+          award: [
+            {
+              title: 'CANCER &AIDS DRUGS--PRECLIN PHARMACOL/TOXICOLOGY',
+              number: 'N01CM037835-016',
+              identifier: 'awardA',
+              scheme: 'awardSchemeA',
+              parentScheme: 'funderScheme1',
+              parentIdentifier: 'funder1',
+            },
+            {
+              title: 'Beyond the Standard Model at the LHC and with Atom Interferometers.',
+              number: '228169',
+              identifier: 'awardB1',
+              scheme: 'awardSchemeB',
+              parentScheme: 'funderScheme2',
+              parentIdentifier: 'funder2',
+            },
+            {
+              title: 'ENvironmental COnditions in GLAucoma Patients',
+              number: '747441',
+              identifier: 'awardB2',
+              scheme: 'awardSchemeB',
+              parentScheme: 'funderScheme2',
+              parentIdentifier: 'funder2',
+            },
+          ],
+        },
+
         related_identifiers: {
           resource_type: this.config.vocabularies.resource_type,
           scheme: [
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
```diff
@@ -113,6 +114,49 @@ export class RDMDepositForm extends Component {
             { text: "Other", value: "other" },
           ],
         },
+
+        // TODO: Replace with an API backend
```
ppanero, line 118: minor: even if I agree with the TODO, in my experience they tend to fall in the forgotten limbo... Openining issues seems more effective.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
```diff
@@ -113,6 +114,49 @@ export class RDMDepositForm extends Component {
...
+        funding: {
```
ppanero, line 119: moderate: This does not match the inner data model (maybe intended?). Here there is a `funding` dictionary with a list of `funder` and a list of `award`, while in the data model there is a list of `funding` (and each funding can have one `funder and/or one `award`).

ppanero (commented):
just need to crosscheck the data model change, if intended or not. Otherwise LGTM.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
fenekku: You're right, we keep forgetting them. I made https://github.com/inveniosoftware/react-invenio-deposit/issues/82 for the more general case.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
fenekku: Yes this is intended. This is the (placeholder) vocabulary, not the actually stored values in the field. When the field is filled we store an array of funding objects:` [{ funder: {...}, award: {...} }, ... ]`.

I was wondering about the schema in invenio-rdm-records, so it's good you brought it up. There can be just a funder? or just an award? The frontend assumes that an award necessarily has a funder. We can go over this on Monday.

tmorrell: It's important to be able to support a Funder without an Award. A lot of times we don't have information about what specific award is associated with a work, or the funding is from an organization that doesn't use awards/grants. 

Having an award require a funder seems like a reasonable requirement, but I'm not sure whether that will cause troubles for Zenodo (I think they accept only award numbers, then map separately)

fenekku: Good to know!

ppanero: Thanks @tmorrell! So If I understand correctly the `funder` is always required but the `award` might or might not be present (@fenekku we need to crosscheck this since it's not consistent with the data model atm). In addition, @fenekku since the front end separates them, how will is associate one funder to award? Assuming positions (i.e. [0] and [0] etc. if so need to add an empty award when there's not so keep the index consistent).

ppanero:
Discussed IRL. It is fins as it is now - being able to link a funder and award. To be revisited (along with data model of this part) when dealing with vocabs (December)

ppanero (approved):

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
fenekku: After discussion, this is just a placeholder. It will be revisited in next Invenio Sprint: https://github.com/inveniosoftware/invenio-rdm-records/issues/264 . :+1:

Issue #256, opened 2020-08-24
styles: add a footer similar to InvenioRDM


lnielsen:
## Describe the solution you'd like

Create an improved footer for InvenioRDM.

1. Two layers. Bottom layer darkens the main header color.
2. Notice the shadow in between the two footers (see also #334)
3. Create Jinja template blocks according to screenshot below.
4. Notice link vs text color. Very slight difference.
5. Exact content to be defined, but we could start out with some of the links from http://inveniosoftware.org/products/rdm/

![Screenshot 2020-08-24 at 16 23 32](https://user-images.githubusercontent.com/1698163/91056541-67a16e80-e626-11ea-9d56-00f22dbec76c.png)

![Screenshot 2020-08-24 at 16 21 32](https://user-images.githubusercontent.com/1698163/91056319-1ee9b580-e626-11ea-9487-1368075b83ee.png)

The idea is that it should be easy for user to add menu, logos and text in the footer, custom to each instance.

fenekku:
This is a more detailed version of https://github.com/inveniosoftware/invenio-app-rdm/issues/152 . Will close the latter.

PR #377, opened 2020-11-06
ui: make two level footer

ppanero:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/256

Current code:
Screenshot 2020-11-06 at 17 03 56

Example with according to issue:
Screenshot 2020-11-06 at 17 19 38

More zoom to display the shadow:
Screenshot 2020-11-06 at 17 20 00

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/footer.less
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/footer.less
@@ -6,26 +6,35 @@
  * under the terms of the MIT License; see LICENSE file for more details.
  */
 
- #footer {
-   color: #fff;
-   margin-top: auto;
+.footer-global {
+  color: @footerTextDarkColor;
+  font-weight: 300;
+  margin-top: 0rem; // Default was `6rem`
+  font-size: @font-size-base;
 
-  &.footer-bottom {
-    background-color: #0377cd;
-    padding-top: 15px;
-    padding-bottom: 15px;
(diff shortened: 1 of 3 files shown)


lnielsen:
Looks good! Here my comments:

- Colors:
   - Use blue from the header (left side) as the background color of the top footer
   - Darken that blue (see Zenodo SCSS for a good value to darken it with)
- Shadows looks good
- Would put the language in the bottom footer bar, but smaller size.
- Menu:
   - Less spacing between menu items (note the menu header is fine.
   - Colors of the menu items you could do with specifying white and make the color a bit transparent.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/footer.less
```diff
@@ -6,26 +6,31 @@
  * under the terms of the MIT License; see LICENSE file for more details.
  */
 
- #footer {
-   color: #fff;
-   margin-top: auto;
+.footer-global {
+  color: #ffffff;
```
zzacharo, line 10: maybe define also here and a variable e.g `footerTextColor` and set it in the `variables.less`? Similar to what you've done for the `footerDarkColor`...

ppanero:
Applied changes:

- Made footer text colors into variables
- Changed footer colors, made light equal than the top and used Zenodo's dark blue
- Moved language to lower footer
- Reduced spacing between menu items
- Made link color white with transparency

Screenshot 2020-11-09 at 09 39 26

PR #398, opened 2020-11-11
search: fix searchbar issue

### LINKED ISSUE #395: search page: query text not taken into account
Package version (if known): <0.16.4> `Containerize` flow

## Describe the bug
When you search for a text, the url parameter is updated but, neither the searchbar is showing the searched value nor the executed search takes into account the querysrtring!

## Steps to Reproduce
1. Go to '/search'
2. Search for something e.g  'test'
3. See that the `?q=` parameter is URL is updated **BUT** the searchbar doesn't display anything and the results are not filtered.

## Expected behavior

- The searchbar should display the searched text after the search.
- The results should be filtered by the querystring value

## Screenshots (if applicable)
Screenshot 2020-11-11 at 10 02 36

zzacharo:
closes #395

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
@@ -14,19 +14,16 @@ import {
   Input,
   Item,
   Label,
-  List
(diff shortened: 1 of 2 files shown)


slint (approved):
LGTM and tested locally and works :shipit:

Issue #392, opened 2020-11-10
ux: search page: search result issues


fenekku:
Package version (if known): 0.16.4

## Describe the bug
- All of the search result is clickable and directs to the record landing page. This is unexpected when clicking on a badge or a creator. Do we need to have the "View" button then? 
- the search result shows escaped HTML for the description.

## Screenshots (if applicable)

![initial_search_page_result](https://user-images.githubusercontent.com/1932651/98729225-36e8ef80-2360-11eb-87f2-afa05ad7a0c8.png)

fenekku:
This was fixed to some extent, so marking it as close. `&nbsp;` are still displayed in the search but there is an issue for that already: https://github.com/inveniosoftware/invenio-app-rdm/issues/592

PR #399, opened 2020-11-11
search: renders description html in results

Glignos:
Addresses the second part of https://github.com/inveniosoftware/invenio-app-rdm/issues/392

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
@@ -82,7 +82,7 @@ export const RDMRecordResultsListItem = ({ result, index }) => {
           ))}
         </Item.Meta>
         <Item.Description>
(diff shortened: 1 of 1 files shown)


Glignos:
Screenshot 2020-11-11 at 14 10 13

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
```diff
@@ -82,7 +82,8 @@ export const RDMRecordResultsListItem = ({ result, index }) => {
           ))}
         </Item.Meta>
         <Item.Description>
-          {_truncate(description, { length: 350 })}
+          <div dangerouslySetInnerHTML={{__html: _truncate(description, { length: 350 })}}>
```
lnielsen, line 85: We should strip HTML tags, and afterwards truncate instead of rendering the HTML - people will be able to screw up the frontpage/search quite a lot if we allow it.

Issue #407, opened 2020-11-17
Move frontpage default code to cookiecutter


fenekku:
Having the frontpage's html, less and assets in this repo, means it is always part of the webpack build process making it slower and generating bigger assets that are eventually unused anyway. Placing the frontpage related things in the cookiecutter gives an immediate customization point to installers and is more efficient.

zzacharo:
@fenekku I am closing this and maybe we could open a more specific issue if it's still needed for improving customization :)

lnielsen:
Log ideas to https://github.com/inveniosoftware/product-rdm/issues/14

fenekku:
Oh I thought this was quite specific: move the frontpage html/js/less from app-rdm into the cookiecutter project, so that the frontpage. which will always be overwritten, doesn't stay in app-rdm. Otherwise, more configuration needed / webpack still builds its assets (pretty sure) and app-rdm is laden with dead code. I will put that in the product one.

PR #408, opened 2020-11-17
ui: improve some customization aspects [+]

fenekku:
- remove specialized header templates {DEPOSITS,SEARCH_UI}_HEADER_TEMPLATE
- fix header.html's html

Longer term things to look into:
- move all frontpage html + less + images into cookiecutter: made an issue for it: https://github.com/inveniosoftware/invenio-app-rdm/issues/407
- we can reduce html nesting (I've collapsed a few layers in my instance)
- `invenio-cli watch` can be slow
- Had to `invenio-cli assets update` and/or restart even if running `invenio-cli watch`
  (don't know what we can do, but a better update experience would be nice)

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -279,17 +279,12 @@ WSGI_PROXIES = 2
 
 # Invenio-APP-RDM
 # =============
-SEARCH_UI_HEADER_TEMPLATE = 'invenio_app_rdm/header.html'
-"""Search page's header template."""
 
 SEARCH_UI_SEARCH_TEMPLATE = 'invenio_app_rdm/search.html'
 """Search page's base template."""
 
 DEPOSITS_FORMS_BASE_TEMPLATE = 'invenio_app_rdm/deposits/forms/deposits.html'
 """Deposits form page's base template."""
-
-DEPOSITS_HEADER_TEMPLATE = 'invenio_app_rdm/header.html'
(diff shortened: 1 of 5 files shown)


ppanero:
@fenekku just to fully understand, for:
> Had to invenio-cli assets update and/or restart even if running invenio-cli watch
(don't know what we can do, but a better update experience would be nice)

Did you use `--development` when installing, and/or when updating assets?

ppanero (commented):
I think the different headers where necessary because:
1- We display a different one in the frontpage (but this could be still ok).
2- There was an issue with the search bar apps which needed to be different? cc @zzacharo since I do not fully recall.

If this two are not an issue, it LGTM! Feel free to merge.

fenekku:
@ppanero

> Had to invenio-cli assets update and/or restart even if running invenio-cli watch

Here it's really just that the experience feels a little clunky even though we know why (and I don't really have a good solution). We have to run `assets update` to symlink/copy; we have to restart to get invenio.cfg changes and we can run `watch` to get updates to already symlinked files. Perhaps ideally `invenio-cli run` can be improved to account for all that, but it's complicated.

> different headers

We are now using the same header across all pages, so removing those is fine. Issue for being able to override the header template of specific files is here: https://github.com/inveniosoftware/invenio-app-rdm/issues/410

fenekku:
Merging in!

Issue #259, opened 2020-08-25
styles: mobile version has many layout faults


lnielsen:
## Describe the bug

If you reduce the browser window size, you'll see many different layout issues on the frontpage (marked with red arrows in the screeenshots).

## Screenshots (if applicable)

![Screenshot 2020-08-25 at 15 56 55](https://user-images.githubusercontent.com/1698163/91183580-ef04e580-e6eb-11ea-8833-fa07c9f17c3f.png)

![Screenshot 2020-08-25 at 15 57 44](https://user-images.githubusercontent.com/1698163/91183587-f1673f80-e6eb-11ea-9117-3fa80019de74.png)

fenekku:
Linking this issue: https://github.com/inveniosoftware/invenio-app-rdm/issues/170

jrcastro2:
Linking this PR: https://github.com/inveniosoftware/invenio-theme/pull/218

Issue #170, opened 2020-06-01
theme: fix mobile layout


lnielsen:
View InvenioRDM on a mobile is currently not possible:

![Screenshot 2020-06-01 at 12 01 43](https://user-images.githubusercontent.com/1698163/83398448-a9dc5a80-a3ff-11ea-9271-263062d67071.png)

lnielsen:
It's better, not perfect, but good enough to close the issue. The rest should be handled via specifc issues.

![Screenshot 2021-03-05 at 09 44 15](https://user-images.githubusercontent.com/1698163/110090244-5be1f200-7d97-11eb-81a1-0d260e8198a7.png)

PR #427, opened 2020-12-03
styles: Fixes mobile layout

jrcastro2:
* Adds responsive navbar, footer and frontpage,
* partially addresses #259 and #170 

### Frontpage
image
### Navbar extended
image

### Frontpage
127 - Full - iPad - 2020-3-12 at 1 36 42 PM
### Navbar extended
image

## Generic Laptop (1280x950)
127 - Full - Generic Laptop - 2020-3-12 at 1 58 11 PM

### DIFF
--- a/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/header.html
+++ b/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/header.html
@@ -7,81 +7,83 @@
     under the terms of the MIT License; see LICENSE file for more details.
 #}
 
-
-
 <header class="theme header">
   <div class="outer-navbar">
     {%- block navbar %}
     <nav id="invenio-nav" class="ui inverted menu borderless">
       <div class="ui grid container">
-        {%- block navbar_header %}
-        {%- block brand %}
-          {%- if config.THEME_LOGO %}
-          <a href="/">
-            <img class="ui medium image rdm-logo"
-                src="{{ url_for('static', filename=config.THEME_LOGO) }}"/>
-          </a>
-          {%- else %}
-          <a class="logo" href="/">{{ _(config.THEME_SITENAME) }}</a>
-          {%- endif %}
-        {%- endblock brand %}
-        {%- endblock navbar_header %}
-
-
-        {%- block navbar_nav %}
-          {%- if config.THEME_SEARCHBAR %}
-            {%- block navbar_search %}
-              <div class="item" id="header-search-bar">
-                <form class="form" action="{{ config.THEME_SEARCH_ENDPOINT }}" role="search">
-                  <div class="ui action icon input">
-                    <input type="text" name="q" placeholder="{{ _('Search') }}">
-                    <button type="submit" class="ui icon search button">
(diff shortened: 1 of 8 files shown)


ppanero (approved):
Amazing! RDM loves you for this!

Doubt: I wonder if in tablet view the login and signup buttons should be right-aligned instead of centred? or even included in the navbar? @lnielsen might have a preference. To me, it looks a bit weird centred.


Question: I see you had to change many CSS classes from (e.g. `ten wide column`) to a custom made one (e.g. `rdm-footer`). This basically means, that SUI is not 100% mobile compatible?

jrcastro2:
> Question: I see you had to change many CSS classes from (e.g. `ten wide column`) to a custom made one (e.g. `rdm-footer`). Does this basically mean that SUI is not 100% mobile compatible?

It's compatible with mobile version but with the grid system you are limited. When defining a `ten wide column` it means that we will have in every screen a concrete width size (defined in percentage), for example; for `ten column` it's 62.5%, but in our case, in the mobile version, we want to have a width of 100% to display the footer nicely, that's the motivation behind the custom clases.

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/header.html
```diff
@@ -7,81 +7,96 @@
...
-  {%- endblock breadcrumbs %}
+{%- block flashmessages %}
+{%- from "invenio_theme/macros/messages.html" import flashed_messages with context -%}
+{{ flashed_messages() }}
+{%- endblock %}
+{%- block breadcrumbs %}
+{%- include "invenio_theme/breadcrumbs.html" %}
+{%- endblock breadcrumbs %}
 </header>
+<script language="javascript">
```
lnielsen, line 90: Question: What is this needed for?

jrcastro2: It's needed for expand or collapse the navbar menu when clicking on the hamburger icon, basically it sets the `ul` tag to active and all the elements from that "active" list change the `display` property from `none` to `block`.

lnielsen: Ok, I think we need to move it to the ``{% block javascript %}`` template block so that it is evaluated at the end of the page (otherwise it's blocking page rendering). I think ideally, it should be in it's own file like this example from Invenio-I18N:

https://github.com/inveniosoftware/invenio-i18n/blob/master/invenio_i18n/assets/semantic-ui/js/invenio_i18n/app.js

https://github.com/inveniosoftware/invenio-i18n/blob/master/invenio_i18n/webpack.py#L29

lnielsen: or perhaps it just has to be added here: https://github.com/inveniosoftware/invenio-theme/blob/master/invenio_theme/assets/semantic-ui/js/invenio_theme/theme.js

ppanero:
Will be available with latest invenio-theme

PR #430, opened 2020-12-09
Bump invenio-rdm-records

slint:


fenekku:
@slint just bumping invdenio-rdm-records is good enough. @zzacharo just bumped this module to the december release line `0.18.X` 2 days ago so we are good on that front.

PR #438, opened 2020-12-11
deposit: add top level form feedback [+]

fenekku:
- remove error handling (moved to react-invenio-deposit)


fenekku:
In the interest of time, I am merging and releasing this one. It just integrates the FormFeedback from react-invenio-deposit and gets rid of the error handling that now resides in react-invenio-deposit.

This will give us a common ground to test the release.

PR #447, opened 2021-01-13
global: project cleanup

lnielsen:
* Removes old scripts and docker files from
  cookiecutter-invenio-instance which is no longer needed due to
  Invenio-CLI.

* Removes outdated installation instructions and updates documentation
  links to point to https://inveniordm.docs.cern.ch.

* Bump version of Invenio Framework to v3.4.0 and updates setup.py
  classifiers.

* Removes __future__ imports.


fenekku (approved):
:fire: code is always good.

Issue #452, opened 2021-01-13
[form][ux] Add Delete/Discard button


fenekku:
Per mockup, add button with text: 
- Delete for unpublished draft
- Discard for draft of published record

PR #464, opened 2021-01-15
deposit: add DeleteButton + fix record serialization [+]

fenekku:
- requires: react-invenio-deposit v0.10.0 (https://github.com/inveniosoftware/react-invenio-deposit/pull/153)
- the initial record was serialized without its links
  * we have to be careful about our interfaces
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/452

### DIFF
--- a/invenio_app_rdm/theme/views.py
+++ b/invenio_app_rdm/theme/views.py
@@ -105,23 +105,22 @@ def ui_blueprint(app):
         # Dereference relations (languages, licenses, etc.)
         draft._record.relations.dereference()
         serializer = UIJSONSerializer()
-        _record = serializer.serialize_object_to_dict(draft._record)
+        record = serializer.serialize_object_to_dict(draft.to_dict())
+
         # TODO: get the `is_published` field when reading the draft
         from invenio_pidstore.errors import PIDUnregistered
         try:
-            _ = service.draft_cls.pid.resolve(
-                    pid_value, registered_only=True)
-            _record["is_published"] = True
+            service.draft_cls.pid.resolve(pid_value, registered_only=True)
+            record["is_published"] = True
         except PIDUnregistered:
-            _record["is_published"] = False
+            record["is_published"] = False
 
-        searchbar_config = dict(searchUrl=search_url)
         return render_template(
             current_app.config['DEPOSITS_FORMS_BASE_TEMPLATE'],
             forms_config=forms_config,
-            record=_record,
+            record=record,
             files=files_list.to_dict(),
-            searchbar_config=searchbar_config
+            searchbar_config=dict(searchUrl=search_url)
         )
 
     @blueprint.route(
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/theme/views.py
```diff
@@ -105,23 +105,22 @@ def deposits_edit(pid_value):
         # Dereference relations (languages, licenses, etc.)
         draft._record.relations.dereference()
         serializer = UIJSONSerializer()
-        _record = serializer.serialize_object_to_dict(draft._record)
+        record = serializer.serialize_object_to_dict(draft.to_dict())
```
fenekku, line 108: Just a friendly note: we should start being more careful about object interaction and focus on using our public interfaces.

ppanero: Nice catch! How to serialize records is a bit messy. Actually the use of jinja and views are breaking the full flow since records do not pass by the service serializers... (Which is one of the things we wanted to fix with the *-resources modules). This was raised on the last release on two of the issues I worked on (deposit search and deposit page nor rendering due to lack of some attributes, UI serializer ones, the line you change was the fix... sorry I didnt notice about the links).

This issue, IIRC, will be tackled in this release? Is part of what Lars called "backend mess"

invenio_app_rdm/theme/views.py
```diff
@@ -105,23 +105,22 @@ def deposits_edit(pid_value):
...
+
         # TODO: get the `is_published` field when reading the draft
         from invenio_pidstore.errors import PIDUnregistered
         try:
-            _ = service.draft_cls.pid.resolve(
-                    pid_value, registered_only=True)
-            _record["is_published"] = True
+            service.draft_cls.pid.resolve(pid_value, registered_only=True)
```
ppanero, line 113: minor: I personally think is useful to have the `_` when it returns something and it is not used. Also, the fact that the return from the resolver is not used might turn into inconsistencies:

- Do something
- Resolve
- Return the original record/draft (not the resolved one)
- The user sees the original (Correct), but the DB/ES one for some reason did not register the action/change.
- Next time the resolved one is taken... There was a change not saved/performed

Maybe too much of a corner case...

ppanero (approved):

invenio_app_rdm/theme/views.py
fenekku: Yes, I think it might be what Lars was talking about, but he was probably focusing on vocabularies for now. It's not clear to me either how to converge the serialization in API response and serialization in Jinja rendering response yet...

invenio_app_rdm/theme/views.py
fenekku: My take is when a variable is assigned, it's so it can be reused. Since it's not reused it doesn't need to be there (minimalist value) and it's less confusing (was it supposed to be reused? was it supposed to be called `_` because of internationalization? ...). Not sure I follow the corner case :sweat_smile:  - assigning the variable doesn't change what resolve returns...

ppanero: well... my point was that actually should be reused to avoid the corner case... but ok... let's hope we don't bump into it :)

PR #465, opened 2021-01-21
deposit: creators component refactor

zzacharo:
closes inveniosoftware/react-invenio-deposit#154

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
@@ -17,6 +17,7 @@ import {
   List,
 } from "semantic-ui-react";
 import { BucketAggregation, Toggle } from "react-searchkit";
+import _find from "lodash/find";
 import _get from "lodash/get";
 import _truncate from "lodash/truncate";
 import Overridable from "react-overridable";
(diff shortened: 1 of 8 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
```diff
@@ -73,23 +74,27 @@ export const RDMRecordResultsListItem = ({ result, index }) => {
...
-              {_get(creator, "identifiers.orcid") ? (
+              {_find(
+                _get(creator, "person_or_org.identifiers", []),
+                (identifier) => identifier?.scheme === "orcid"
+              ) ? (
                 <img className="inline-orcid" src="/static/images/orcid.svg" />
               ) : null}
-              {creator.name}
+              {creator.person_or_org?.name}
               {creators.length > 1 && index != creators.length - 1 ? "," : null}
```
fenekku, line 84:
```js
{ 
  _get(creator, "person_or_org.identifiers", []).find((identifier) => identifier.scheme === "orcid") &&
  <img className="inline-orcid" src="/static/images/orcid.svg" />
}
{creator.person_or_org.name}
{(index + 1 < creators.length) && ","}
```

I am assuming we have polifylling setup with Babel so that we can use features that are unequally supported (looking at you IE!).
But the bigger discussion is: do we want to?

Here we can omit optional chaining: We have to assume knowledge of the structure/content of what we are getting. Otherwise, the code becomes obscured by optional chaining / we can't code new things without writing checks every line if what we hold is what we think we hold... 
In other words we can safely admit the backend gives us non-undefined identifiers.

On "usage of javascript features". I tend to be conservative on the issue:
- use as much as possible the built-in features, so we don't have to rely on external libraries
- prioritize the features that are more widely supported first (if you can achieve what you want with them, do so)
- carefully consider the implications of the more cutting edge features we do decide to use (there must be a pretty good reason at this point)

Obviously, balancing those is an art :smile: . My guiding stars are then: does this make the code more readable/expressive?", "how many javascript concepts are needed in order to understand the code?".

fenekku (approved):
Like other approving for time. Overall, we shouldn't have to try to bridge the gap in data format change: we can assume data is structured according to changes for this release. We haven't released a stable version so we don't have to worry about it :smile_cat:

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
zzacharo: Damn IE.....I was happy this feature got included in the javascript core 😅  I agree with you about the assumption of the data structure. This was a leftover from when the creators were mixed (old demo data plus newly added). I think is safe enough to remove it! 

In general, this feature improves code readability I think. In the cases, we cannot assume the structure, I think is a good option to use it now that is included in the core!

PR #471, opened 2021-01-21
deposit: stickify deposit form sidebar

### LINKED ISSUE #468: ui: sticky sidebar
The side bar should be stick

 Screenshot 2021-01-21 at 14 53 28

Example in https://react.semantic-ui.com/modules/sticky/

fenekku:
- closes #468 
- fixes a missing `Grid,Row` that was causing some subtle layout issues for a long time

**Screenshots&&

image

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
@@ -15,6 +15,7 @@ import {
   DepositFormApp,
   FormFeedback,
   DeleteButton,
+  DepositFormTitle,
   DescriptionsField,
   FileUploader,
   FundingField,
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposits/RDMDepositForm/RDMDepositForm.js
```diff
@@ -250,160 +251,170 @@ export class RDMDepositForm extends Component {
     };
   }
 
+  contextRef = createRef()
```
zzacharo, line 254: According to [this](https://stackoverflow.com/questions/54620698/whats-the-difference-between-useref-and-createref) I would put the `contextRef` initialization in the constructor to avoid recreating it in every render. Althoug h, `RDMDepositForm` component shouldn't rerender, it doesn't depend on dynamic state, I would move it to the constructor, just to be safe :) I will merge this and I can include it in one of my upcoming PRs!

zzacharo (approved):

PR #502, opened 2021-02-02
style: helptext label style

mb-wali:
closes inveniosoftware/react-invenio-deposit/issues/186


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -46,6 +47,13 @@
     font-size: 0.9em;
   }
 
+  .ui.form label.helptext{
```
zzacharo, line 50: I think this rule can be merged with the above if not the above already applies...Can you double check?

mb-wali: I pushed the change, how come its not ```outdated```?

zzacharo (approved):

PR #538, opened 2021-02-03
search: improve search items + refactor

### LINKED ISSUE #516: list uploads: item result template updates
1. Remove the statistics
2. Add creators
3. Remove the "by <user>" because I think it is currently display creators instead of the user who created the item.

Screenshot 2021-02-02 at 12 00 22

fenekku:
- closes #516
- make user_records_search items and search items converge on very similar implementation
  * it would be possible in a future refactor to merge the two
    (since we only have 2 instances, DRYing up is not necessary yet)
- :warning:  you will need to rebuild your assets to see the changes because of file name changes (seen as new files by the manifest.json)
- also fixes some JS console errors on the record landing page. 
- also fixes access label size (was slightly bigger than the other ones because of icon)
- One of the changes to make them converge is to use floated buttons for actions in both cases. This goes back to the look in a previous mockup. Feel free to indicate if this derogates too much from intention @lnielsen .

**Screenshots**
*Before*

Search item
search_item

User record search item
user_records_item

*After*

Search item
new_search_item

User record search item
new_user_records_item


zzacharo (approved):

lnielsen:
> One of the changes to make them converge is to use floated buttons for actions in both cases. This goes back to the look in a previous mockup. Feel free to indicate if this derogates too much from intention @lnielsen .

👍 That works......

Two tiny comments for the user records:

1. Can we move up the status icon to be top aligned. 
2. Make sure that we can later add back the name of the submitter (marked box)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_records_search/index.js
```diff
@@ -15,14 +16,13 @@ import {
...
   RDMRecordFacetsValues,
-  RDMToggleComponent,
-} from "../../search/components";
-import {
   RDMRecordSearchBarElement,
-  RDMCountComponent,
-} from "../../search/components";
+  RDMToggleComponent,
+} from "../search/components";
+import { RDMRecordSearchBarElement } from "../search/components";
```
zzacharo, line 25: Here I guess this import is unnecessary right? It is already imported in line 22...

zzacharo:
@fenekku I will merge this so you don't have a lot of conflicts when you are back online! Please create another PR addressing the comments from @lnielsen !

PR #543, opened 2021-02-04
form: sub field labels font-weight to normal

mb-wali:
closes inveniosoftware/react-invenio-deposit#88
not sure about the size of top-level fields - if its too big, see screenshots.
*screenshot*
font-size

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
@@ -21,6 +21,20 @@
     }
   }
 
+  /*Field Label*/
+  .ui.form .field .form-group label{
+    font-size: 13px;
+    font-weight: normal;
+  }
+
+  label.field-level-class {
+    font-size: 15px;
+    &.small {
+      font-size: 13px
+    }
+  }
+  /**/
+
   .save-button {
     margin-bottom: 0.5em;
   }


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -21,6 +21,19 @@
...
+
+  /**targets the top-level fields, id=top-level-fields is injected in react.*/
+  label#top-level-fields{
+    font-size: 14px;
+  }
+  /**targets the top-level resource_type field,
+  this is because this field (label) does not have the id.*/
+  .required.field label[for="metadata.resource_type"]{
+    font-size: 14px;
+  }
```
fenekku, line 36: See https://github.com/inveniosoftware/react-invenio-deposit/pull/197#discussion_r570997932 if we can get a way for us not to repeat these one-offs...  If the strategies there don't work, then let's go with this (but with classes :wink: )

mb-wali: yes- you are right, i was thinking the same while working on it. Mybe i can hack around the css and target each and every ```for=""``` attributes of the labels and add the styles - would that make sense?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -21,6 +21,23 @@
...
+  /*Field Label*/
+  .ui.form .field .form-group label{
+    font-weight: normal;
+  }
+  // top level field
+  label.top-level-fields{
+    font-size: 14px;
+  }
+  // tob level field explicitly
+  label.top-level-fields[for="metadata.languages"]{
```
zzacharo, line 33: Is the `for` clause needed here?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -21,6 +21,23 @@
...
+    font-size: 14px;
+  }
+  // label without top-level field class
+  .required.field label[for="metadata.resource_type"]{
```
zzacharo, line 37: Also here maybe we can apply a specific class like the above `top-level-fields`?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -21,6 +21,23 @@
     }
   }
 
+  /*Field Label*/
+  .ui.form .field .form-group label{
```
zzacharo, line 25: Also here, I think is better to apply a class to the fields you want to use that css rule! Also, now that the rdm theme is merged maybe you can put these rules in the `form.overrides` file instead of here! That way, we can start changing the `.ui.xxx` rules...

zzacharo (approved):

PR #548, opened 2021-02-05
523 export as json

### LINKED ISSUE #523: landing page: make JSON export return the record's JSON
Now that only JSON is exported on the landing page, a JSON representation of the record should actually be exported.

max-moser:
closes #523 

Introduced a new configuration item for registering record serializers, e.g.:
```
APP_RDM_RECORD_EXPORTERS = {
    "json": {
        "name": "JSON",
        "serializer": ("invenio_rdm_records.resources.serializers:"
                       "UIJSONSerializer")
    }
}
```
The key is used for the lookup and should be URL-safe, while the `"name"` property is used for display purposes.
The `"serializer"` property is an import string for (or direct reference to) a factory function that can be called without arguments which creates an instance of the serializer (e.g. an argument-less constructor).

*Side note*: This configuration item is also used in the side-bar template, to only display links to export formats of registered serializers.

The URL to the export UI of a record can also be configured via a new configuration item (which needs to include both `<pid_value>` and `<export_format>`):
`APP_RDM_RECORDS_EXPORT_URL = "/records/<pid_value>/export/<export_format>"`

*Side note*: This URL is rendered in the jinja template for the side-bar via `url_for(...)`

Also, the `views` have been refactored a bit to split up general and record-related view functions.

**Screenshot of the export page**:
image

### DIFF
--- a/invenio_app_rdm/theme/views/__init__.py
+++ b/invenio_app_rdm/theme/views/__init__.py
@@ -7,7 +7,36 @@
 
 """Blueprints and views for Invenio-App-RDM."""
 
-from .general import ui_blueprint
-from .records_ui import file_download_ui, records_ui_blueprint
+from flask import Blueprint
+from flask_menu import current_menu
 
-__all__ = ("file_download_ui", "ui_blueprint", "records_ui_blueprint")
+from .filters import register_filters
+from .general import register_general_ui_routes
+from .records import file_download_ui, register_records_ui_routes
+
+
+def ui_blueprint(app):
+    """Blueprint for the routes and resources provided by Invenio-App-RDM."""
+    blueprint = Blueprint(
+        "invenio_app_rdm",
+        __name__,
+        template_folder="../templates",
+        static_folder="../static",
+    )
+
+    register_filters(blueprint)
+    register_general_ui_routes(app, blueprint)
+    register_records_ui_routes(app, blueprint)
+
+    @blueprint.before_app_first_request
+    def init_menu():
+        """Initialize menu before first request."""
+        item = current_menu.submenu("main.deposit")
+        item.register(
+            "invenio_app_rdm.deposits_user", "Uploads", order=1
+        )
+
+    return blueprint
+
+
+__all__ = ("file_download_ui", "ui_blueprint")
--- a/invenio_app_rdm/theme/views/__init__.py
+++ b/invenio_app_rdm/theme/views/__init__.py
@@ -10,6 +10,7 @@
 from flask import Blueprint
 from flask_menu import current_menu
 
+from .deposits import register_deposits_ui_routes
 from .filters import register_filters
 from .general import register_general_ui_routes
 from .records import file_download_ui, register_records_ui_routes
@@ -27,14 +28,13 @@ def ui_blueprint(app):
     register_filters(blueprint)
     register_general_ui_routes(app, blueprint)
     register_records_ui_routes(app, blueprint)
+    register_deposits_ui_routes(app, blueprint)
 
     @blueprint.before_app_first_request
     def init_menu():
         """Initialize menu before first request."""
         item = current_menu.submenu("main.deposit")
-        item.register(
-            "invenio_app_rdm.deposits_user", "Uploads", order=1
-        )
+        item.register("invenio_app_rdm.deposits_user", "Uploads", order=1)
 
     return blueprint
 
--- a/invenio_app_rdm/theme/views/__init__.py
+++ b/invenio_app_rdm/theme/views/__init__.py
@@ -1,0 +1,13 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 TU Wien.
+#
+# Invenio App RDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Blueprints and views for Invenio-App-RDM."""
+
+from .general import ui_blueprint
+from .records_ui import file_download_ui, records_ui_blueprint
+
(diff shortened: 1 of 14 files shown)


max-moser:
idea: if we have the `UIJSONSerializer` from RDM-Records to accept arguments that are passed to to the `json.dumps(...)` calls, a pretty-printing `UIJSONSerializer` could be configured like so:

```
APP_RDM_RECORD_EXPORTERS = {
    "json": {
        "name": "JSON",
        "serializer": lambda: UIJSONSerializer(indent=2)
    }
}
```

invenio_app_rdm/theme/views/records.py
```diff
@@ -0,0 +1,100 @@
...
+        resource = current_rdm_records.records_resource
+        service = current_rdm_records.records_service
+        links_config = resource.config.links_config
+
+        record = service.read(
+            id_=pid_value, identity=g.identity, links_config=links_config
+        )._record
+
+        # get the configured serializer
+        temp = app.config.get("APP_RDM_RECORD_EXPORTERS", {}).get(
```
fenekku, line 43: `temp` -> `exporter`

invenio_app_rdm/theme/views/records.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+        if temp is None:
+            raise Exception(
+                "no exporter for the specified format registered: {}".format(
+                    export_format
+                )
+            )
+
+        format_name = temp.get("name", export_format)
+        serializer = obj_or_import_string(temp["serializer"])()
```
fenekku, line 54: Yeah... UIJSONSerializer isn't currently implemented to allow the usage of the pretty_print options we talked about :disappointed: ... I made an issue outlining the problem and solution here: https://github.com/inveniosoftware/invenio-rdm-records/issues/367 . If you want to take it separately, it would help :+1: .

invenio_app_rdm/theme/views/records.py
```diff
@@ -0,0 +1,100 @@
...
+            "template", "invenio_app_rdm/landing_page/export_page.html"
+        )
+
+        resource = current_rdm_records.records_resource
+        service = current_rdm_records.records_service
+        links_config = resource.config.links_config
+
+        record = service.read(
+            id_=pid_value, identity=g.identity, links_config=links_config
+        )._record
```
fenekku, line 40:
```suggestion
        record = service.read(
            id_=pid_value, identity=g.identity, links_config=links_config
        ).to_dict()
```
This should be enough to get a serializable result. Access to `_record` was probably done in another view, but we shouldn't do it since it's private.

setup.py
```diff
@@ -95,7 +96,7 @@
             'invenio-app-rdm = invenio_app.cli:cli',
         ],
         'invenio_base.blueprints': [
-            'invenio_app_rdm = invenio_app_rdm.theme.views:ui_blueprint',
+            "invenio_app_rdm = invenio_app_rdm.theme.views:ui_blueprint",
```
fenekku, line 99: probably unnecessary ;)

invenio_app_rdm/theme/views/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+    @blueprint.before_app_first_request
+    def init_menu():
+        """Initialize menu before first request."""
+        item = current_menu.submenu("main.deposit")
+        item.register("invenio_app_rdm.deposits_user", "Uploads", order=1)
+
+    return blueprint
+
+
+__all__ = ("file_download_ui", "ui_blueprint")
```
fenekku, line 42: :100: Thanks for the split into files. Much nicer this way.

fenekku (commented):
This is good. Mostly minor comments. After you make adjustments I will merge.

setup.py
max-moser: :smile: 
yes, that's still a left-over from experiments with multiple blueprints, and now it's just there looking awesome

invenio_app_rdm/theme/views/records.py
max-moser: I got the `record._record` out, but had to tweak a couple things in the Jinja template to get it running.
From what I see there's no problem, but maybe there's stuff breaking somewhere under the hood that I don't see:

* had to throw out the `record.data|dereference_record` in `export_page.html`, because `record.data` (dict) doesn't have a `relations` attribute (https://github.com/inveniosoftware/invenio-app-rdm/pull/548/commits/423752e444722d6c85bc584bfa76891bfec472f4#diff-4f71f75b58c44d01d2135bfec1a989b083a16321f4833c7c997c4e1e1e75ff2eL17)
* had to comment out the access-right icon, because the `record.ui` doesn't have an `access_right` property when done this way (https://github.com/inveniosoftware/invenio-app-rdm/pull/548/commits/423752e444722d6c85bc584bfa76891bfec472f4#diff-4f71f75b58c44d01d2135bfec1a989b083a16321f4833c7c997c4e1e1e75ff2eR36)
* to preserve compatibility with existing `render_template` calls, I added a code-path that handles the `record` being specified as `RDMRecord`, or as `RecordItem` together with separate `pid_value` argument (https://github.com/inveniosoftware/invenio-app-rdm/pull/548/commits/423752e444722d6c85bc584bfa76891bfec472f4#diff-124d51df0ed3415b56175f047c630953e5df5b5e312406e74f7cb9c121493736L59)

fenekku: Thanks for highlighting these issues It's fine for now Those are all points that we need to take into account when improving that response flow / representation. Personally, I am not happy yet with the squishiness of some abstractions and we will be making improvements.

fenekku (approved):

PR #552, opened 2021-02-08
theme: move license styling to rdm theme

zzacharo:
closes https://github.com/inveniosoftware/react-invenio-deposit/issues/194

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/list.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/list.overrides
@@ -1,0 +1,33 @@
+/*******************************
+    User Variable Overrides
+*******************************/
+
+@import "@less/invenio_theme/theme/elements/list.overrides";
+
+/* Deposit license field */
+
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/list.overrides
```diff
@@ -0,0 +1,33 @@
...
+
+.item.deposit-drag-listitem {
+    margin-bottom: 5px;
+    &.hidden {
+      opacity: 0;
+    }
+    .drag-anchor {
+        cursor: move;
+    }
+}
```
mb-wali, line 17: If some one wants to add a new style or class for deposit license files, it should be done here?

mb-wali (approved):

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/list.overrides
zzacharo: I would say yes :) That way we organize the custom classes per field!

PR #556, opened 2021-02-10
deposit: add support for default values in the deposit form

max-moser:
closes inveniosoftware/invenio-rdm-records#358

in the screenshot: default values for **rights** (licenses), **publisher**, and **publication_date** (via a callable).
image

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -28,7 +28,7 @@ WARNING: An instance should NOT install multiple flavour extensions since
          there would be no guarantee of priority anymore.
 """
 
-from datetime import timedelta
+from datetime import datetime, timedelta
 
 from .theme.utils import previewer_record_file_factory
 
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/theme/views/deposits.py
```diff
@@ -43,10 +45,17 @@ def deposits_create():
             vocabularies=Vocabularies.dump(),
             current_locale=str(current_i18n.locale),
         )
+
+        new_record = dump_empty(RDMRecordSchema)
+        defaults = app.config.get("APP_RDM_DEPOSIT_FORM_DEFAULTS") or {}
+        for key in defaults:
```
zzacharo, line 51: minor: here you could do `for key, value in defaults.items()`

zzacharo (approved):

invenio_app_rdm/config.py
```diff
@@ -362,3 +362,26 @@ def _(x):
 }
 
 APP_RDM_RECORDS_EXPORT_URL = "/records/<pid_value>/export/<export_format>"
+
+APP_RDM_DEPOSIT_FORM_DEFAULTS = {
+    "publication_date": lambda: datetime.now().strftime("%Y-%m-%d"),
+    "rights": [
+        {
+            "id": "cc-by-4.0",
+            "title": "Creative Commons Attribution 4.0 International",
```
lnielsen, line 371: Minor: This data should somehow be retrieved from the license vocabulary. Perhaps via a callable:

```
from invenio_vocabularies.proxies import current_service

def default_license(id_):
    def callable():
         item = current_service.read(('licenses', id_), system_identity).data
         return {"id": item.id, ...}
   return callable

APP_RDM_DEPOSIT_FORM_DEFAULTS = {
     "publication_date": lambda: datetime.now().strftime("%Y-%m-%d"),
     "rights": default_license('cc-by-4.0'),
```

PR #557, opened 2021-02-11
deposit: sticky error messages

### LINKED ISSUE #534: form: display errors with consistent message styling
The "flash messages" used for display errors should be used consistently throughout the side.

The deposit form error messages are currently shown as below:
Screenshot 2021-02-03 at 14 12 12

# Expected behavior

The error message should be displayed in a similar style yo below:

Screenshot 2021-02-03 at 14 12 39

The warning message should be shown below the header, and if possible sticky, so that when you scroll down it stays on the top. It has to be seen how this plays with the sticky right sidebar with the buttons

mb-wali:
when setting the header (flash error messages) sticky on top, it interact with sidebar sticky bar (protection widget), for the time being  the increase to `offset` for sidebar Sticky will do it. but should find a better way to deal with this.

closes #534 

*screenshot*
solution2

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
@@ -8,7 +8,15 @@
 
 import _get from "lodash/get";
 import React, { Component, createRef } from "react";
-import { Button, Card, Grid, Icon, Ref, Sticky } from "semantic-ui-react";
+import {
+  Button,
+  Card,
+  Container,
+  Grid,
+  Icon,
+  Ref,
+  Sticky,
+} from "semantic-ui-react";
 import {
   AccessRightField,
   ComingSoonField,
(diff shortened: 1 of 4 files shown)


lnielsen:
1. Is the message from side to side, similar to the screenshot in the issue? It might not be too clear from the screenshot now that I look at it again. But e.g. if you try to send a verification email again to yourself, then the style of the message should match.

2. The sticky menu? How does it look if there's no error message?

mb-wali:
> 1. Is the message from side to side, similar to the screenshot in the issue? It might not be too clear from the screenshot now that I look at it again. But e.g. if you try to send a verification email again to yourself, then the style of the message should match.
> 2. The sticky menu? How does it look if there's no error message?

Here is some different screenshots.
**Without Error scrolled**
without-error

**Width of error messages**
full-width

zzacharo:
Latest screenshot
Screenshot 2021-02-15 at 15 19 53

zzacharo:
Created another issue for making the message sticky https://github.com/inveniosoftware/invenio-app-rdm/issues/558 . Currently only the styling is aligned with the other `flashed` messages.

PR #566, opened 2021-02-17
deposit: format disabled in sidebar

fenekku:
part of closing https://github.com/inveniosoftware/react-invenio-deposit/issues/218


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -68,4 +68,9 @@
     width: unset;  // Allows the sidebar cards to take the full space of their column
   }
 
+  .ui.list .disabled {
```
zzacharo, line 71: This can be easily moved to `list.overrides` so list components can use it along the instance :)

fenekku: done!

PR #575, opened 2021-02-22
bugfix: label whitespace

### LINKED ISSUE #572: deposit form: help text whitespace issue
The help text is not placed close enough to it's associated field. Also, there's too little margin to the next field.,

Screenshot 2021-02-19 at 15 16 54

mb-wali:
closes #572 

*screenshot*
label-whitespace


fenekku (approved):
This is not related but I noticed there is this strange "\" in the identifiers global help text:

image

We should get rid of it.

Issue #595, opened 2021-02-22
Landing page access control 


lnielsen:
A landing page display a single record. This includes for instance:
    
- Landing page: https://inveniordm.web.cern.ch/records/rny5x-zt515
- Export format pages: https://inveniordm.web.cern.ch/records/rny5x-zt515/export/dc
   
**Authorization**

Any landing page should require ``read`` permission to the entire record.
    
The record MUST be fetched through the service layer to ensure to ensure that field-level permissions have been applied (i.e. metadata that is not allowed to be access by a given identity is filtered out, e.g. specifically that files have been filtered out or not).
        
**Field-level permissions**

- Files: The ``files`` field should require ``read-files`` permission.
   
**Conditional rendering**
    
- [x] Manage panel: Each landing page should display a special "manage" section which contains e.g. an Edit button. This section MUST require ``update`` permission to be rendered. Subparts MAY require other permissions (e.g. delete button MAY require ``admin`` permission).
- [ ] Files box:
    - Record without files:
        - Files box show that it is metadata only record.
    - Record with files
        - Files included in metadata (i.e. user has access):
            - If public: normal rendering.
            - If protected and files are embargoed: render with yellowish file box.
            - If protected and files are not embargoed: render with redish file box.
        - Files not included in metadata (i.e. user has no access):
            - If files are embargoed: render embargo information
            - If files are protected: render information that files are protected
-  [ ] Preview:
    - Same rules as for the files box.

lnielsen:
```
{
"permissions": {
    "can_manage": True,
   "can_read_files": True,
}
}
```

PR #603, opened 2021-02-23
details: start re-implementing the file preview and download

### LINKED ISSUE #579: landing page: make preview work again
In #560 the views for the landing page were refactored. The following were left out:

- File preview
- File download

## File download

⚠️  Before starting we have to check if https://inveniordm.docs.cern.ch/reference/rest_api/#record-files should be used instead.

- [View that was removed](https://github.com/inveniosoftware/invenio-app-rdm/pull/560/files#diff-988eb51a0bc0a8bf6efd0b04d1bb6b6d8f24cc80504ea36699a202904a20af4dL183-L188)
- A new view was added [here](https://github.com/inveniosoftware/invenio-app-rdm/pull/560/files#diff-f8f8f03c590caf2e424ff8ba99e652f600b0f4fb973dc1e8172365daa1e28bfaR52-R55) and [here](https://github.com/inveniosoftware/invenio-app-rdm/pull/560/files#diff-2fb602eed821cbd7d07f7c61560b3e8ed50e165259da085c113fff47274b8496R64-R91)

The goal of this view is simply to have nice URLs like ``/records/12345-abcde/files/<key>`` and provide a proper human-friendly error when a file is not found (instead of a REST API 404 with a JSON message.

The task is to reimplement the view:

- The ``@pass_record`` decorator takes care of reading the record via the service (i.e. record permission level checks will have been applied). 
- Probably ``record.files`` has to be used to get the files (instead of the approach being used in the current method). 
- The call to ``ObjectResource.send_object`` The permission check ``ObjectResource.check_object_permission(obj)`` might need to be done directly in the view (since we have the record readily available and i'm afraid that ``check_object_permission`` will try to retrieve the record from the database.

Things to check:
- A public record with restricted files. If the user cannot access the files, the files should not be downloadable.

## File preview

- [View that was removed](https://github.com/inveniosoftware/invenio-app-rdm/pull/560/files#diff-988eb51a0bc0a8bf6efd0b04d1bb6b6d8f24cc80504ea36699a202904a20af4dL189-L194)
- [Stub view](https://github.com/inveniosoftware/invenio-app-rdm/pull/560/files#diff-2fb602eed821cbd7d07f7c61560b3e8ed50e165259da085c113fff47274b8496R58)

The file preview is similar to the file download. The goal is that the record file preview is passed through the service so that proper access restrictions have been applied.

max-moser:
closes #579 

some of the work is already done, but there's still some issues...

image

# the file name on the previewer isn't updated, when another file is selected for preview
image

# FIXED: when none of the files are previewable, the preview box should be hidden
image

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -60,12 +60,20 @@
         {%- endblock record_content -%}
         {# files #}
         {%- block record_files -%}
-        {%- if record.files and record|can_list_files %}
-          {%- set files = record.files|make_files_preview_compatible %}
-          {%- set default_preview_file = files|select_preview_file(default_preview=record.files.default_preview) %}
-          {{ preview_file_box(default_preview_file, pid) }}
-          {{ file_list_box(files, pid) }}
-        {% endif %}
+        {%- if files is not none -%}
+          {%- if files.enabled and record|can_list_files %}
+            {%- set default_preview = files.default_preview %}
+            {%- set files = files|order_entries %}
+            {%- if files|has_previewable_files -%}
+              {%- set default_preview_file = files|select_preview_file(default_preview=default_preview) %}
+              {{ preview_file_box(default_preview_file, pid) }}
+            {%- endif -%}
+            {{ file_list_box(files, pid) }}
+          {% endif %}
+        {%- else -%}
+          {# TODO: come up with a proper display #}
+          <div>Sorry, the files are restricted!</div>
+        {%- endif -%}
         {%- endblock record_files -%}
         {%- block record_details -%}
         {# More details #}
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
@@ -1,13 +1,24 @@
 {#
   Copyright (C) 2020 CERN.
   Copyright (C) 2020 Northwestern University.
-  Copyright (C) 2021 Graz University of Technology
+  Copyright (C) 2021 Graz University of Technology.
+  Copyright (C) 2021 TU Wien.
 
   Invenio RDM Records is free software; you can redistribute it and/or modify
   it under the terms of the MIT License; see LICENSE file for more details.
 #}
 
-{% from 'invenio_previewer/macros.html' import file_list, preview_file %}
+{%- macro preview_file(preview_endpoint, pid_value, filename, id='preview-iframe', width='100%', height='400' ) %}
+  <iframe
+    class="preview-iframe"
+    id="{{id}}"
+    name="{{id}}"
+    width="{{width}}"
+    height="{{height}}"
+    src="{{ url_for(preview_endpoint, pid_value=pid_value, filename=filename) }}">
+  </iframe>
+{%- endmacro %}
+
 
 {% macro preview_file_box(file, pid) %}
 <div class="panel-spacing">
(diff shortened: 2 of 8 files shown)


invenio_app_rdm/records_ui/views/records.py
```diff
@@ -9,26 +9,61 @@
...
+from invenio_previewer.proxies import current_previewer
+from invenio_rdm_records.proxies import current_rdm_records
 from invenio_rdm_records.resources.serializers import UIJSONSerializer
 
-from ..utils import previewer_record_file_factory
-from .decorators import pass_record
+from .decorators import pass_file_metadata, pass_record_files, pass_record
+
+
+class PreviewFile:
```
max-moser, line 24: this one replaces `invenio_previewer.api:PreviewFile`, which sometimes has problems when accessing `.size` (because often, `self.file` would not support dict-like access)

this should probably find a better home

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -9,26 +9,61 @@
...
+
+    def has_extensions(self, *exts):
+        """Check if file has one of the extensions."""
+        file_ext = splitext(self.data["key"])[1]
+        file_ext = file_ext.lower()
+        return file_ext.lower() in exts
+
+    def open(self):
+        """Open the file."""
+        return self.file._file.file.storage().open()
```
max-moser, line 52: not 100% sure if this is really required, and it's kind of a wonky operation

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -55,41 +94,66 @@ def record_export(record=None, export_format=None, pid_value=None):
...
+                current_app.logger.warning(
+                    (
+                        "Preview failed for {}, in {}:{}".format(
+                            fileobj.file.key,
+                            pid_type,
+                            pid_value,
+                        )
+                    ),
+                    exc_info=True,
+                )
```
max-moser, line 132: this just feels outdated, can probably be removed too?

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
```diff
@@ -17,13 +28,57 @@
...
     <div id="collapsablePreview" class="active content">
-      {{ preview_file('invenio_records_ui.recid_previewer', pid=pid, filename=file.key) }}
+      {{ preview_file('invenio_app_rdm_records.record_file_preview', pid_value=pid, filename=file.key) }}
     </div>
   </div>
 </div>
 {%- endmacro %}
 
 
+{# TODO respect the 'order', if it's specified #}
```
max-moser, line 38: TODO

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -60,9 +60,9 @@ <h1>{{ metadata.title }}</h1>
         {%- endblock record_content -%}
         {# files #}
         {%- block record_files -%}
-        {%- if record.files and record|can_list_files %}
-          {%- set files = record.files|make_files_preview_compatible %}
-          {%- set default_preview_file = files|select_preview_file(default_preview=record.files.default_preview) %}
+        {%- if files.enabled and record|can_list_files %}
+          {%- set files = files.entries %}
+          {%- set default_preview_file = files|select_preview_file(default_preview=files.default_preview) %}
           {{ preview_file_box(default_preview_file, pid) }}
```
max-moser, line 66: i think this should be made conditional, something along the lines of

`{% if any([file is previewable for file in files]) %}`

... just in a valid fashion, using the `previewable` test and filters - does anybody know how to do this?

max-moser: fixed with a new filter, `has_previewable_files`

max-moser:
issue: apparently, explicitly set "preview files" aren't honored (if set)

tests/api/test_record_api.py
```diff
@@ -147,13 +153,11 @@ def test_read_record_with_redirected_pid(client, location, minimal_record,
     assert response.json["message"] == "Moved Permanently."
 
 
-def test_read_deleted_record(client, location, minimal_record, users,
-                             es_clear):
+def test_read_deleted_record(
+    client_with_login, location, minimal_record, es_clear, admin_user
+):
```
max-moser, line 158: this test fails b/c the permission for record deletion is currently `Disabled()`, i think

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -60,12 +60,19 @@ <h1>{{ metadata.title }}</h1>
...
+            {%- set files = files|order_entries %}
+            {%- if files|has_previewable_files -%}
+              {%- set default_preview_file = files|select_preview_file(default_preview=files.default_preview) %}
+              {{ preview_file_box(default_preview_file, pid) }}
+            {%- endif -%}
+            {{ file_list_box(files, pid) }}
+          {% endif %}
+        {%- else -%}
+          {# TODO: come up with a proper display #}
+          <div>Sorry, the files are restricted!</div>
```
max-moser, line 74: should the message be included here, or rather in the view via `flash`ing messages?

max-moser:
issue: i think it would make for nicer UX if clicking a file's "Download" button would always simply start the file download, instead of opening the file in the current browser tab in some cases (e.g. txt files)

tests/conftest.py
```diff
@@ -105,5 +108,45 @@ def dump_user(user):
     db.session.commit()
     return {
         'user1': dump_user(user1),
-        'user2': dump_user(user2)
+        'user1_raw': user1,
+        'user2': dump_user(user2),
+        'user2_raw': user2,
```
max-moser, line 113: i think the `user1: dump_user(user1)` are never actually used, should we remove them instead?

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
lnielsen: This is part of #595 to have a nice page so we can simply integrate like this. Ping @zzacharo

tests/conftest.py
lnielsen: Yes, if not used, then remove :-)

tests/api/test_record_api.py
lnielsen: :-) It's good, it should not be possible to delete the record (or at least it's a super admin task). I would remove the test complete. We need proper tombstone handling, etc, so we'll handle a deleted record at that time.

invenio_app_rdm/records_ui/views/records.py
lnielsen: Yeah, back in the days it was because we had some super weird files that should be previewable but wasn't.

invenio_app_rdm/records_ui/views/records.py
lnielsen: Ok for now...... we may need to rethink all the file previewing a bit.....not very pretty to download a 20MB PDF in the browser :-)

PR #605, opened 2021-02-24
landing-page: add permissions to record view

zzacharo:
* Show manage section only if user has `update_draft` permission

Partially closes https://github.com/inveniosoftware/invenio-app-rdm/issues/595

### DIFF
--- a/invenio_app_rdm/records_ui/views/decorators.py
+++ b/invenio_app_rdm/records_ui/views/decorators.py
@@ -55,9 +55,7 @@ def pass_draft(f):
     def view(**kwargs):
         pid_value = kwargs.get('pid_value')
         draft = service().read_draft(
-            id_=pid_value,
-            identity=g.identity,
-            links_config=draft_links_config()
+            id_=pid_value, identity=g.identity, links_config=links_config()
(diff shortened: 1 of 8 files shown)


invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -34,25 +36,53 @@ def pass_record(f):
...
-        kwargs['draft'] = draft
-        return f(**kwargs)
+        try:
+            pid_value = kwargs.get('pid_value')
+            draft = service().read_draft(
+                id_=pid_value, identity=g.identity, links_config=links_config()
+            )
+            kwargs['draft'] = draft
+            return f(**kwargs)
+        except PermissionDeniedError:
```
lnielsen, line 64: You can use ``blueprint.error_handler()``  ([see example](https://github.com/inveniosoftware/invenio-records-ui/blob/master/invenio_records_ui/views.py#L66-L72)). This way we'll handle it on all views in the same manner. We'll anyway also need it for PIDDoesNotExists error etc.

lnielsen (approved):

PR #610, opened 2021-02-24
deposit: fix styling in user records search

zzacharo:
Screenshot 2021-02-25 at 00 17 59


fenekku (approved):
I don't recall how it looked before (apart from button changes), but this looks as it should :art:

PR #616, opened 2021-02-25
records ui: fix undefined error on export-as page

max-moser:


invenio_app_rdm/records_ui/views/records.py
```diff
@@ -73,7 +72,9 @@ def record_detail(record=None, files=None, pid_value=None, permissions=None):
 
 
 @pass_record
```
max-moser, line 74: if the "manage" ribbon should be shown on the export page as well, the `@user_permissions(actions=['update_draft'])` decorator could be set here

lnielsen: I would say yes, it should show up.

lnielsen (approved):

PR #635, opened 2021-02-25
js: on window scroll show btn

mb-wali:
closes https://github.com/orgs/inveniosoftware/projects/63#card-55675686

unscroll
scroll

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/index.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/index.js
@@ -36,8 +36,22 @@ if (document.getElementById("recordManagement")) {
     $("#preview-iframe").attr("src", $(event.target).data("url"));
   });
 
-  $("#jump-btn").on("click", function (event) {
-    document.documentElement.scrollTop = 0;
+
+  // Show `Jump up` button window scroll
+  $(function() {
+    var jumpbtn = document.getElementById("jump-btn");
+    var prevScrollTop = $(window).scrollTop()
+    $(window).on('scroll', function(e) {
+      var $src = $('.main-div');
+      var currentScrollTop = $(this).scrollTop()
+      if (currentScrollTop >= prevScrollTop && currentScrollTop > 44) {
+        jumpbtn.style.display = "block";
+      }
+    });
+    $("#jump-btn").on("click", function (event) {
+      document.documentElement.scrollTop = 0;
+    });
+
   });
 
   // func to toggle the icon class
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/index.js
```diff
@@ -36,8 +36,22 @@ if (document.getElementById("recordManagement")) {
...
+      var $src = $('.main-div');
+      var currentScrollTop = $(this).scrollTop()
+      if (currentScrollTop >= prevScrollTop && currentScrollTop > 44) {
+        jumpbtn.style.display = "block";
+      }
+    });
+    $("#jump-btn").on("click", function (event) {
+      document.documentElement.scrollTop = 0;
+    });
+
```
fenekku, line 54: The goal here is to show the jump button only once we have scrolled down the page (right?).

:thinking: 
- why do we need to compute the `prevScrollTop = $(window).scrollTop()` ? Shouldn't we always compare to 0 since it's for the top of the page?
- that hardcoded 44 is for...? (the size of the button I think?)

fenekku (approved):
It works, but it seems complicated? :sweat_smile:

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/index.js
mb-wali: yes,
* some times you find your self inside a page already scrolled half way for example [#develop](https://inveniordm.docs.cern.ch/#develop)
* 44 is the pixels scrolled from the top.

fenekku: Whenever we arrive on a page half scrolled, then `document.documentElement.scrollTop > 0` is true, and the moment this is the case, it means we want to display the jump button. I guess 44 can replace 0 for the threshold So would something like this work?

```suggestion
  $(function() {
    var jumpbtn = document.getElementById("jump-btn");
    var threshold = 44;
    if (document.documentElement.scrollTop > threshold) {
      jumpbtn.style.display = "block";
    }
  });
  
  $("#jump-btn").on("click", function (event) {
      document.documentElement.scrollTop = 0;
  });
```

```

fenekku: I guess that wouldn't work because we are not checking if the window is scrollable...

fenekku:
We will revisit the UX for this in March. We are putting it in blocked temporarily.

PR #638, opened 2021-02-25
templates: make them configurable

fenekku:
- closes https://github.com/inveniosoftware/docs-invenio-rdm/issues/109
- documented in https://github.com/inveniosoftware/docs-invenio-rdm/pull/110


fenekku:
blocked / delayed and potentially removed if we want to keep the overridding mechanism

Issue #649, opened 2021-03-02
List uploads: split in  unpublished, published, draft


zzacharo:
We want to split the record/draft status per search page as follows:

- Search page
  - Records should have a `status == published` and `is_published == True`.
- Uploads page - records status
  - `published`: records that are published and have no draft. (`is_published == True`)
  - `draft`: records that are published but have an active draft i.e being edited. (`is_published == True`
  - `unpublished`: new drafts with no associated published records. `is_published == False`

fenekku:
Exchanging ideas here about the uploads page...

What about keeping draft and unpublished together? I am wondering about the benefit of keeping them separate. There seems to be more work implementation-wise to separate them and then that's more to display to the user / more decision for them to make/remember.

What about making the published records just be all records that are... published (period)? That tweak was notoriously confusing to our users when using the framework directly. They would think their published record was back to a draft only. 

Drafts would just be drafts. 

 An indicator and/or link to draft/published record would make clear the respective published record/draft.

zzacharo:
Eventually, after discussing with @lnielsen we will go for:

- `Published` records or drafts or published records
- `New` drafts of not published records i.e new drafts

PR #679, opened 2021-03-09
wip split uploads search record statuses

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/649

requires https://github.com/inveniosoftware/invenio-rdm-records/pull/433
requires https://github.com/inveniosoftware/invenio-records-resources/pull/209/files

# Screenshots

Screenshot 2021-03-10 at 14 28 51

Screenshot 2021-03-10 at 15 38 51

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_records_search/components.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_records_search/components.js
@@ -6,13 +6,14 @@
 // Invenio App RDM is free software; you can redistribute it and/or modify it
 // under the terms of the MIT License; see LICENSE file for more details.
 
-import React from "react";
+import React, { useState } from "react";
 import {
   Icon,
   Card,
   Container,
   Grid,
   Label,
+  Modal,
   Item,
   Button,
   Segment,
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_records_search/components.js
```diff
@@ -37,6 +38,49 @@ import {
...
+      <Button compact floated="right" color="red" onClick={handleOpen}>
+        <Icon name="trash alternate outline" />
+        Delete
+      </Button>
+
+      <Modal open={modalOpen} onClose={handleClose} size="tiny">
+        <Modal.Content>
+          <h3>Are you sure you want to delete this draft?</h3>
+        </Modal.Content>
+        <Modal.Actions>
```
lnielsen, line 71: Question: Do we have any kind of feedback to the user on these buttons? Seems to be that we'll have a lot of buttons which sends an API request, and we should be able to show a spinner and if something went wrong. Perhaps not for this PR, but we need something similar for the "New version" button.

lnielsen (approved):
Good for me to go.

lnielsen (approved):

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_records_search/components.js
zzacharo: We don't have at the moment but yes we need a common way to present all the API errors coming from action buttons e.g the delete, new version etc.

Issue #684, opened 2021-03-10
landing page: add "New version" button


lnielsen:
Next the current "Edit" button we should display a "New version button"

- Action:
   - Send an API request ``POST /records/:id/versions``. The redirect to deposit form page for the new version
   - The form should show a progress while sending the API reqeust, and show if an error happened.

- Permissions: 
   - can_edit: button is shown disabled, and explains that they don't have manage permissions.
   - can_manage: button shown enabled.

A similar button is added to the deposit form, so think about reuse.

Issue #683, opened 2021-03-10
landing page: add new version alert


lnielsen:
See Zenodo

- Alert on the top "There is a **newer version** of this record available.".
- Link should go to ``/records/:id/latest`` (which doesn't yet exists).

![Screenshot 2021-03-10 at 11 24 22](https://user-images.githubusercontent.com/1698163/110616960-6f28ff00-8195-11eb-9383-68c3ac25d1f2.png)

lnielsen:
https://zenodo.org/record/4570815

PR #689, opened 2021-03-11
New Version available/button

fenekku:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/683
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/684
- reduce DOM calls
- fix jump button bug (and others) by making semantic ui
  javascript independent of presence of recordManagement
  div (which could be absent if viewer is not owner)

Banner with newer version available, New Version button and error response displayed.

banner_w_error

NOTE: 
- Because check on `can_manage` permission is done and this permission is not provided, the New Version Button will not show up by default for now.
- On the other hand, the "newer_version_exists" test has been set to True so that the "Newer version available" banner is always shown for now.

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -19,6 +19,13 @@
 
 {%- block page_body %}
 <div class="banners">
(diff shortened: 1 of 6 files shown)


zzacharo:
After discussing with @lnielsen I pushed 2 changes:

- New version button should be green similar to `New upload` button
- The "There is a newer version" link moved to bottom of the banner so that it's closer to the record metadata

zzacharo (approved):

PR #690, opened 2021-03-12
run-tests: forward supplied options and arguments to pytest

### LINKED ISSUE #675: [tooling] Allow running one test with run-tests.sh
When working on any of the RDM repositories, it is slow to run all tests locally all the time via `run-tests.sh`, but it has become necessary because it sets up isolated services / some tests are not passing otherwise. Being able to run:

```
pipenv run ./run-tests.sh </path/to/test/file>
``` 
would be better (it's a large overhead but at least it doesn't run everything which would be even longer)

Simply editing `run-tests.sh` to allow command line arguments to be forwarded to pytest.

## Alternatives
Delving deeper into how not needing services for local tests would work. (this seems much harder and messy because on CI we do want the services).  

## Additional context
I used to run my local tests via pytest directly and a local ES container that I would spin up and down quickly. It was fast and targeted: I could just run the tests for the files I was working on multiple times. This setup doesn't work anymore `Database locked` errors with sqlite and redis is required for app-rdm but a generic container (like for ES) doesn't do it... I have been pushed to edit my local run-tests.sh manually :( 

With the potential influx of new developers, removing friction will have large benefits.

max-moser:
closes #675 
since the main goal of `run-tests.sh` is to call `pytest`, i think it makes sense to just forward the given CLI options/arguments to the `pytest` call

this enables running the tests in specific files (e.g. `./run-tests.sh tests/test_version.py`), as well as selecting tests on a pattern-matching basis (e.g. `./run-tests -k test_set_default_value__value`), etc.

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -3,9 +3,20 @@
 #
 # Copyright (C) 2020 CERN.
 # Copyright (C) 2020 Northwestern University.
+# Copyright (C) 2021 TU Wien.
 #
 # Invenio App RDM is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
+
+# Usage:
+#   ./run-tests.sh [pytest options and args...]
+#
+# Note: the DB, SEARCH and CACHE services to use are determined by corresponding environment
+#       variables if they are set -- otherwise, the following defaults are used:
+#       DB=postgresql, SEARCH=elasticsearch and CACHE=redis
+#
+# Example for using mysql instead of postgresql:
+#    DB=mysql ./run-tests.sh
 
 # Quit on errors
 set -o errexit
@@ -19,11 +30,10 @@ function cleanup() {
 }
 trap cleanup EXIT
 
-
 python -m check_manifest --ignore ".*-requirements.txt"
 python -m sphinx.cmd.build -qnNW docs docs/_build/html
 eval "$(docker-services-cli up --db ${DB:-postgresql} --search ${SEARCH:-elasticsearch} --cache ${CACHE:-redis} --env)"
-python -m pytest
+python -m pytest $@
 tests_exit_code=$?
 python -m sphinx.cmd.build -qnNW -b doctest docs docs/_build/doctest
 exit "$tests_exit_code"


max-moser:
another improvement could be to be smart about the `docker-services-cli`:
for instance, we could add a flag that tells the script not to shut down the services after the tests ran through, which would make subsequent runs of `run-tests.sh` faster

run-tests.sh
```diff
@@ -3,10 +3,14 @@
 #
 # Copyright (C) 2020 CERN.
 # Copyright (C) 2020 Northwestern University.
+# Copyright (C) 2021 TU Wien.
 #
 # Invenio App RDM is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
+# Usage:
+#   env DB=postgresql ./run-tests.sh [pytest options and args...]
```
fenekku, line 12: Why `env DB=postgresql` ? Works without... :thinking:

fenekku (approved):
Thanks that's what I had in mind.

run-tests.sh
max-moser: ah yeah, that's just an example
the magic happens in line 28: `${DB:-postgresql}` will look for the environment variable `${DB}` and use it if it's defined - and if it's not defined, it'll use `"postgresql"` as fallback

fenekku: Right and postgresql is the default anyway. `Usage` sounded like a "must", but it's a "may".

max-moser: i'll update the comment accordingly

run-tests.sh
```diff
@@ -3,10 +3,21 @@
...
 
+# Usage:
+#   ./run-tests.sh [pytest options and args...]
+#
+# Note: the DB, SEARCH and CACHE services to use are determined by corresponding environment
+#       variables if they are set -- otherwise, the following defaults are used:
+#       DB=postgresql, SEARCH=elasticsearch and CACHE=redis
+#
+# Example for using mysql instead of postgresql:
+#    DB=mysql CACHE=redis ./run-tests.sh
```
fenekku, line 19: minor: remove the `CACHE=redis` so that the example is the most clear.

Issue #735, opened 2021-03-28
landing page: caption of the preview box does not change


lnielsen:
Package version (if known): <version>

## Describe the bug

Preview: The caption of the preview box does not change when selecting another file for the preview.

## Steps to Reproduce

1. Upload and publish a new record with two files.
2. On the landing page, click preview of the first file, then the next. 
3. Observe that the filename in the previewbox does not change.

## Expected behavior

The filename in the previewbox should display the file being previewed.

fenekku:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/642 ;
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/74

PR #742, opened 2021-03-29
landing-page: change caption in the preview box on file change

audrium:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/735
result:
pr


fenekku:
Just a note here: this loses us the angled down arrow icon. Maybe we want to keep it :)

PR #743, opened 2021-03-29
landing-page: share record access link modal

### LINKED ISSUE #703: landing page: get a link user interface
### Tasks

- [ ] Add "Share" button to landing page manage section. The button requires ``can_manage`` permission. If the user has ``update_draft`` permission, the button should be in disabled state with a tool tip ``You do not have permissions to share this record.``

Screenshot 2021-03-18 at 09 16 01

- [ ] Landing page should make sure that the link is not injected intot the template if ther user does not have permissions.

- [ ] Clicking on "Share" opens a dialog if you have permissions. See mockups below.

Screenshot 2021-03-18 at 10 26 46

mvidalgarcia:
closes #703

image

#### To test

- Create a record
- As creator, you should be able to see the Share button
- Create an access link
- Visit `https://127.0.0.1:5000/api/records/<recid>/access/links` (API) to check if the access link has been created properly with the right permission
- Edit share mode (permission) and verify in the API.
- Delete link and verify in the API.

#### TBD

- One record can have multiple links (1:n), how should the UI reflect this? So far only accessing the element 0 of the list (emulating a 1:1 relationship).
  - > Only one link initially......later we may have more.....
- The generated access links are the same for all the users? No reference to users when creating access links.

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareButton.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareButton.js
@@ -1,0 +1,25 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2021 Northwestern University.
+//
+// Invenio RDM Records is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import React from "react";
+import { Icon, Button, Popup } from "semantic-ui-react";
+
+export const ShareButton = (props) => {
+  return (
+    <Popup
+      content="You don't have permissions to share this record."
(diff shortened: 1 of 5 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareButton.js
```diff
@@ -0,0 +1,42 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2021 Northwestern University.
```
lnielsen, line 2: Minor: CERN should be here

lnielsen (approved):
Looks good!Just if you can update the copyright, then we can merge.

Issue #669, opened 2021-03-02
deposit: create a draft preview


zzacharo:
It would be useful to have a preview of an "in progress" draft before the user clicks publish. A way to implement this is to create a new simple view or adapt the landing page one, to be able to render a draft record!

lnielsen:
For the version box:

![Screenshot 2021-03-30 at 14 33 55](https://user-images.githubusercontent.com/1698163/112989318-11b91a00-9165-11eb-8539-0e3b784314eb.png)

---

For the preview header message there's three different variations:

```record.is_published is True```

![Screenshot 2021-03-30 at 14 26 12](https://user-images.githubusercontent.com/1698163/112989377-2695ad80-9165-11eb-8195-536fbef3fcf1.png)

```record.is_published is False and record.versions.index == 1```

![Screenshot 2021-03-30 at 14 26 28](https://user-images.githubusercontent.com/1698163/112989378-272e4400-9165-11eb-8401-6077fb3b6d57.png)

```record.is_published is False and record.versions.index >= 1```

![Screenshot 2021-03-30 at 14 26 50](https://user-images.githubusercontent.com/1698163/112989379-27c6da80-9165-11eb-828a-c1a9c27581a4.png)

PR #751, opened 2021-03-31
deposit: draft preview

jrcastro2:
* Adds draft preview
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/669

## Details page
image

## Export page
image

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -18,14 +18,42 @@
 
 {%- block page_body %}
 <div class="banners">
-  {% if permissions is defined and permissions.can_edit %}
+  {% if is_preview %}
+  <div class="ui info flashed top-attached manage message">
+    <div class="ui container">
+      <div class="ui relaxed grid">
+        <div class="column">
+          <div class="row">
+            <div class="header">
+              <i class="eye icon"></i> Preview
+            </div>
+            <p>
+              {% if record.is_published %}
+                You are previewing changes that have not yet been published.
+              {% endif %}
+
+              {% if not record.is_published and record.versions.index == 1 %}
+                You are previewing a new record that has not yet been published.
+              {% endif %}
+
+              {% if not record.is_published and record.versions.index > 1 %}
+                You are previewing a new record version that has not yet been published.
+              {% endif %}
+            </p>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+  {% endif %}
+  {% if permissions is defined and permissions.can_edit and not is_preview %}
   <div class="ui warning flashed top-attached manage message">
     <div class="ui container">
       <div id="recordManagement" data-record='{{ record | tojson }}' data-permissions='{{ permissions | tojson }}'></div>
     </div>
   </div>
   {% endif %}
-  {% if not record.versions.is_latest %}
+  {% if (not record.versions.is_latest and not is_preview) or (is_preview and not record.versions.is_latest_draft) %}
   <div class="ui warning flashed top-attached manage message">
     <div class="ui container">
       <div class="ui relaxed grid">
@@ -89,9 +117,9 @@
             {%- set files = files|order_entries %}
             {%- if files|has_previewable_files -%}
               {%- set default_preview_file = files|select_preview_file(default_preview=default_preview) %}
-              {{ preview_file_box(default_preview_file, pid) }}
+              {{ preview_file_box(default_preview_file, pid, is_preview) }}
             {%- endif -%}
-            {{ file_list_box(files, pid) }}
+            {{ file_list_box(files, pid, is_preview) }}
           {% endif %}
         {%- else -%}
           {# TODO: come up with a proper display #}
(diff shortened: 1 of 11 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -18,14 +18,14 @@
...
-  {% if permissions is defined and permissions.can_edit %}
+  {% if permissions is defined and permissions.can_edit and not is_preview %}
   <div class="ui warning flashed top-attached manage message">
     <div class="ui container">
       <div id="recordManagement" data-record='{{ record | tojson }}' data-permissions='{{ permissions | tojson }}'></div>
     </div>
   </div>
   {% endif %}
-  {% if not record.versions.is_latest %}
+  {% if not record.versions.is_latest and not is_preview %}
```
lnielsen, line 28: Minor: I think we should still show the "newer" version message on the preview, so that they'll see how it looks like

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -42,6 +42,30 @@
...
+          <div class="row">
+            <div class="header">
+              <i class="eye icon"></i> Preview
+            </div>
+            <p>
+              {% if record.is_published %}
+                You are previewing changes that has not yet been published.
+              {% endif %}
+
+              {% if not record.is_published and record.versions.index >= 1 %}
```
lnielsen, line 59: Minor: The third case is missing here - ``if not record.is_published and record.versions.index == 1``

lnielsen (approved):
Just two minor things, otherwise looks very good

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -77,18 +64,48 @@ def view(**kwargs):
     return view
 
 
+def pass_record_or_draft(f):
+    """Decorate to retrieve the record or draft using the record service."""
+    @wraps(f)
+    def view(**kwargs):
+        pid_value = kwargs.get('pid_value')
+        preview = request.args.get('preview')
+        if preview == '1':
```
fenekku, line 73: What about another decorator `pass_preview` that sticks `is_preview` into `kwargs` (like `record` is here)? This decorator would make it so that this line becomes:

```suggestion
        preview = kwargs.get('is_preview')
        if is_preview:
```

Then that `preview == '1'` check is centralized which would position us to collapse the conditional urls easily... If that actually works :thinking:

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js
```diff
@@ -85,9 +102,10 @@ export const RecordVersionsList = (props) => {
   }, []);
 
   return loading ? (
-    <PlaceholderLoader />
+    <>{isPreview ? <PreviewMessage /> : <PlaceholderLoader />}</>
```
fenekku, line 105: Sure... it will show up when loading is done so it's not a big deal if `PreviewMessage` is absent here (or place it before the conditional so not to have to repeat it).

fenekku (approved):

fenekku:
That's a cool addition!

invenio_app_rdm/records_ui/views/decorators.py
jrcastro2: I like this idea! Thanks a lot!

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js
jrcastro2: It's needed there because if the call we do just before (check [here](https://github.com/inveniosoftware/invenio-app-rdm/blob/9f78032067aca5a0b3d2f5a8533de88b4a1df5f0/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js#L86-L94)) doesn't retrieve any results it will stay with the loading state for ever. That's the case for the drafts.

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/side_bar.html
```diff
@@ -42,7 +42,11 @@
     {# dynamically create the list of export formats #}
     {%- for fmt, val in config.get("APP_RDM_RECORD_EXPORTERS", {}).items() -%}
       {%- set name = val.get("name", fmt) -%}
-      {%- set export_url = url_for('invenio_app_rdm_records.record_export', pid_value=record.id, export_format=fmt) -%}
+      {% if is_preview %}
+        {%- set export_url = url_for('invenio_app_rdm_records.record_export', pid_value=record.id, export_format=fmt, preview=1) -%}
+      {% else %}
+        {%- set export_url = url_for('invenio_app_rdm_records.record_export', pid_value=record.id, export_format=fmt) -%}
+      {% endif %}
```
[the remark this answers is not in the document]
jrcastro2: Unfortunately not, `is_preview` is a boolean and we need to pass ''1" as value.

PR #752, opened 2021-03-31
ui: fix overlapping btn in upload list

### LINKED ISSUE #720: list uploads: buttons interfere with text and text not clickable
Screenshot 2021-03-24 at 13 59 36

## Expected behavior

1) That the title does not overlap with the buttons, but instead wraps around
2) That the title of a search result item is clickable similar to search results page.

mvidalgarcia:
closes #720

image


fenekku (approved):
We might look into placing "search styles" into `search.less` eventually, but the style is fine where it is now.

PR #753, opened 2021-04-01
deposit: language field

jrcastro2:
* Rewrites the score calculation to
  properly rank the results


jrcastro2:
Not needed anymore, rewrite param is added in the backend

Issue #667, opened 2021-03-02
global: access status label on search and landing pages


zzacharo:
This task is about adding an access status badge back to records.

⚠️ The easiest in this task is adding the badge. The hard part is about vocabularies.

The badge should be displayed on:

1. Landing page
2. Search results - each result item
3. List uploads - each result item

Examples:

![Screenshot 2021-04-06 at 07 33 27](https://user-images.githubusercontent.com/1698163/113663715-ce056980-96aa-11eb-8e6f-0fbac68bfc00.png)
![Screenshot 2021-04-06 at 07 35 06](https://user-images.githubusercontent.com/1698163/113663719-cf369680-96aa-11eb-857a-a3fb55331f4e.png)

### Vocabulary

The badge should be defined as a vocabulary. See e.g.:

- [Vocabularies definitions](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/data/vocabularies.yaml)
- [Example of licenses vocabulary](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/data/vocabularies/licenses.csv)

The following properties is already available in the vocabulary:

1. id 
2. icon
3. title (i18n)

You'll have to add another property:

4. color in the ``props`` section of the vocabulary

### Vocabulary items:

You will have the following items:

- Open: a public records with public files
- Embargoed: 1) a public record with restricted files under embargo + 2) a restricted record under embargo
- Metadata-only: 1) a public record without files and 2) a public record with restricted files (not under embargo)
- Restricted: Restricted records (not under embargo)

### UISeralizer

The information should be added to the UISerializer in Invenio-RDM-Records and be used from there in 

```json
{
  "ui": {
    "access_status": {
      "id": "",
      "icon: "",
      "color": "",
      "title_l10n": "",
      "description_l10n": ""
    }
  }
}
```

### Facet in search

The search results should also has a facet "Access status"  that allow filtering records by the 4 options:

- [x] Open
- [x] Embargoed
- [x] Metadata-only
- [x] Restricted

PR #762, opened 2021-04-08
global: access status label on search and landing pages

diegodelemos:
* Includes search results, user search results, and record landing pages.

Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/667.

Depends on:
- https://github.com/inveniosoftware/invenio-rdm-records/pull/499 (backend implementation)
- https://github.com/inveniosoftware/invenio-records/pull/252 (adds needed hooks to record base api)
- https://github.com/inveniosoftware/invenio-records-resources/pull/233 (adds access status to service schema dump so it is accessible in the UI serialisers)

**Status**
Mostly done see screen recordings.

https://user-images.githubusercontent.com/9035606/114845595-30154b80-9ddc-11eb-9459-02e62861510b.mov

https://user-images.githubusercontent.com/9035606/114845605-33103c00-9ddc-11eb-8d9b-1d5db57e09bd.mov

**Missing**

- Move `AccessStatusEnum` to a more suitable location
    - moving it to `records/api.py` made more sense so we expose it in the record module's external api, however, this doesn't work because there is a circular import (fields are imported in `api.py`).

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -87,9 +87,10 @@
             </div>
             <div class="right floated right aligned column">
               <span class="ui label small grey">{{ record.ui.resource_type }}</span>
-              <!--TODO: Re-enable in next releases-->
-              <!-- <span class="ui label small access-right {{ 'unlock' }}">
-                <i class="icon {{ 'unlock' }}"></i>{{ _("Open Access") }}</span> -->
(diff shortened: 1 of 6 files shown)


diegodelemos:
@lnielsen could you please have a first look at this PR group? Especially at fact that [the access status is placed in the root of the record dict](https://github.com/inveniosoftware/invenio-rdm-records/pull/499/files#diff-7748f03a5b3b5d20f5948609435dd46e364c684cfa8770860deaad584533cb26R211). The reason is explained [here](https://github.com/inveniosoftware/invenio-rdm-records/pull/499/commits/6c52c393f4bfe0f1d06da4d327cbd2434ea62aec), I didn't want to change the dumper behaviour since it is there for backwards compatibility. An option to keep status inside `RecordAccess` object would be to change the dumper to do a recursive `update` on the dictionary (`.update` on every value that is a dict), let me know if you have some advice.

lnielsen (approved):

PR #770, opened 2021-04-12
records-ui: Re-enable the DOI badge in the landing page

### LINKED ISSUE #760: re-enable DOI badge in the landing page
The DOI badge has been disabled since it was a mere placeholder. The DOI badge is displayed on the right *details* column in Zenodo, e.g.:

Screenshot 2021-04-08 at 09 12 19

In RDM it should be displayed in the main landing page. The HTML for this page can be found in the [invenio_app_rdm/records_ui/.../records/detail.html](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html), be careful not to confuse it with the _.../records/details/details.html_.

How to show the DOI badge has been already implemented in the [doi.html](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/doi.html).

The goal of the task is to obtain the DOI (if exists) from the record and show it in the landing page (*details.html*.). This can be done with the `include` primitive (see the *details.html* for examples).

Road to the goal:

- [ ] Get an instance up an running.
- [ ] Fork and clone locally the `invenio_app_rdm` package.
- [ ] Install `invenio_app_rdm` for local development [invenio-cli install packages ...](https://inveniordm.docs.cern.ch/reference/cli/#packages-install).
- [ ] Edit the *details.html* file with some dummy HTML code (e.g. add an h1) to see the changes on the site.
- [ ] Include the *doi.html*, for example below right holders and  above rights/subjects. It might fail, since the records might not have a proper doi set up. Hopefully it will *just* show "No minted DOI" as text 🚀 .
- [ ] Add the `pids` field to the demo data. This might imply a change in `invenio-rdm-records/cli.py`, therefore you will need to fork and clone said repo, install for local development (`invenio-cli install packages...`). Ask Pablo for the structure of this field, but it should be as described in the [RFC](https://codimd.web.cern.ch/WGzPHxxZTI-kFj-BNTrZoQ).
- [ ] Then you might need to modify the [Jinja filter here](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/views/filters.py#L94) to obtain it properly. Ideally it should be as easy as `record.pids.doi.value`, performing existance checks etc.
- [ ] Crosscheck that the modal shows correctly.

Tips: When doing a PR add sreenshots to make the review process lighter :)

Pineirin:
Closes #760 
Screenshot from 2021-04-12 12-55-30

Update: DOI title added before the DOI badge.
Screenshot from 2021-04-12 15-22-12

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -97,10 +97,12 @@
         {%- block record_title -%}
         <h1>{{ metadata.title }}</h1>
         <p>{%- include "invenio_app_rdm/records/details/creators.html" %}</p>
+
         {%- endblock record_title -%}
+
         {%- block record_content -%}
         <p>{%- include "invenio_app_rdm/records/details/contributors.html" %}</p>
-
+        <p>{%- include "invenio_app_rdm/records/details/doi.html" %}</p>
         {%- include "invenio_app_rdm/records/details/subjects.html" %}
         <div class="top-padded">
         {%- include "invenio_app_rdm/records/details/description.html" %}
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -97,9 +97,7 @@
         {%- block record_title -%}
         <h1>{{ metadata.title }}</h1>
         <p>{%- include "invenio_app_rdm/records/details/creators.html" %}</p>
-
         {%- endblock record_title -%}
-
         {%- block record_content -%}
         <p>{%- include "invenio_app_rdm/records/details/contributors.html" %}</p>
         <p>{%- include "invenio_app_rdm/records/details/doi.html" %}</p>
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/doi.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/doi.html
@@ -7,8 +7,9 @@
 #}
 
 {%- set id_doi = record.pids|doi_identifier %}
-<dd>
 {% if id_doi %}
+<dt><b>{{_('DOI')}}</b></dt>
+<dd class="top-bottom-padded">
   <span class="get-badge" data-toggle="tooltip" data-placement="bottom" style="cursor: pointer;"
     title="Get the DOI badge!">
     <img id="record-doi-badge" data-target="[data-modal='{{ id_doi }}']"
@@ -28,7 +29,5 @@
       {{ badges_formats_list(url_for('invenio_formatter_badges.badge', title='DOI', value=id_doi, ext='svg', _external=True, _scheme='https'), id_doi|pid_url(scheme='doi')) }}
     </div>
   </div>
-{% else %}
-  {{ _("No minted DOI") }}
-{% endif %}
 </dd>
+{% endif %}
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/doi.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/doi.html
@@ -6,7 +6,7 @@
   it under the terms of the MIT License; see LICENSE file for more details.
 #}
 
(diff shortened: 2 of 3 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/doi.html
```diff
@@ -6,7 +6,8 @@
   it under the terms of the MIT License; see LICENSE file for more details.
 #}
 
-{%- set id_doi = metadata.identifiers|doi_identifier %}
+{%- set id_doi = record.pids|doi_identifier %}
+<dt><b>{{_('DOI')}}</b></dt>
 <dd>
```
ppanero, line 11: There is a tiny space that does not show here. See e.g. the subjects title gets a bit of space in between title and content. It is because it lacks the css class (it was already "broken", not coming from your changes but we can fix it now since you are touching the file).

```suggestion
<dd class="top-bottom-padded">
```

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -97,10 +97,12 @@
         {%- block record_title -%}
         <h1>{{ metadata.title }}</h1>
         <p>{%- include "invenio_app_rdm/records/details/creators.html" %}</p>
+
```
ppanero, line 100: minor: Are this two spaces needed?

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/doi.html
```diff
@@ -6,7 +6,8 @@
   it under the terms of the MIT License; see LICENSE file for more details.
 #}
 
-{%- set id_doi = metadata.identifiers|doi_identifier %}
+{%- set id_doi = record.pids|doi_identifier %}
+<dt><b>{{_('DOI')}}</b></dt>
```
ntarocco, line 10: I would have added a `if doi is set` (pseudcode): DOI can be optional and might not be there. What happens in that case?
Maybe it is already handled somewhere else...

ppanero: Needs discussion IRL after tomorrow standup. Not sure we understand the expected end goal/result

invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -96,7 +96,7 @@ def doi_identifier(identifiers):
     for identifier in identifiers:
         # TODO: extract this "DOI" constant to a registry?
         if identifier == "doi":
-            return identifiers[identifier]
+            return identifiers[identifier]['identifier']
```
ppanero, line 99: moderate: Since the `identifiers` is a dict. Would this work:

```
return identifiers.get("pids")
```

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/doi.html
ppanero: Discussed IRL: If no DOI show **nothing** (no title, no sentence)

Pineirin:
Feedback implemented in last commit (merged with the original).

ntarocco:
@Pineirin I still don't see how the `DOI` label will be not visible when the doi is not present. Can you please demo the change in both cases, when DOI is set and when DOI is not set?
I would also recommend you to use branches even in your fork and not directly `master`, in case you have to work on 2 features at same time in the same repo :)

invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -96,7 +96,7 @@ def doi_identifier(identifiers):
     for identifier in identifiers:
```
ppanero, line 96: moderate: I think there was a missunderstanding here... if you use the `get` there is no need to loop on top of it :)

Pineirin:
The changes weren't pushed (I was pushing the wrong folder as I duplicated it).
I have included the new feedback, also here there are some screenshots to illustrate the behavior with and without DOI.

With
Screenshot from 2021-04-16 10-33-04
Screenshot from 2021-04-16 10-33-13

Without
Screenshot from 2021-04-16 11-05-14

ppanero (approved):
Nice job man 🚀

PR #798, opened 2021-04-23
[WIP]language: add language field back to additional_{title,desc}

fenekku:


fenekku:
This was dealt with recently.

PR #801, opened 2021-04-27
views: add pid resolving redirection ui endpoint

ppanero:
requires https://github.com/inveniosoftware/invenio-rdm-records/pull/565
Move endpoint from RDM-Records

Screenshot 2021-04-27 at 12 58 46
Screenshot 2021-04-27 at 12 59 18

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -435,6 +435,7 @@ APP_RDM_ROUTES = {
     "record_export": "/records/<pid_value>/export/<export_format>",
     "record_file_preview": "/records/<pid_value>/preview/<path:filename>",
     "record_file_download": "/records/<pid_value>/files/<path:filename>",
+    "record_from_pid": "/<any({schemes}):pid_scheme>/<path:pid_value>",
     "record_latest": "/records/<pid_value>/latest",
     "deposit_search": "/uploads",
     "deposit_create": "/uploads/new",
--- a/invenio_app_rdm/records_ui/views/__init__.py
+++ b/invenio_app_rdm/records_ui/views/__init__.py
@@ -19,8 +19,8 @@ from .filters import can_list_files, dereference_record, doi_identifier, \
     has_previewable_files, make_files_preview_compatible, order_entries, \
     pid_url, select_preview_file, to_previewer_files, vocabulary_title
 from .records import not_found_error, record_detail, record_export, \
-    record_file_download, record_file_preview, record_latest, \
-    record_permission_denied_error, record_tombstone_error
+    record_file_download, record_file_preview, record_from_pid, \
+    record_latest, record_permission_denied_error, record_tombstone_error
 
 
 #
@@ -46,6 +46,16 @@ def create_blueprint(app):
         routes["record_latest"],
         view_func=record_latest,
     )
+
+    app_ext = app.extensions['invenio-rdm-records']
+    schemes = app_ext.records_service.config.pids_providers.keys()
+    schemes = ','.join(schemes)
+
+    if schemes:
+        blueprint.add_url_rule(
+            routes["record_from_pid"].format(schemes=schemes),
+            view_func=record_from_pid,
+        )
 
     blueprint.add_url_rule(
         routes["record_export"],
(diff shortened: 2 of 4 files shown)


invenio_app_rdm/config.py
```diff
@@ -435,6 +435,7 @@ def files_rest_permission_factory(obj, action):
     "record_export": "/records/<pid_value>/export/<export_format>",
     "record_file_preview": "/records/<pid_value>/preview/<path:filename>",
     "record_file_download": "/records/<pid_value>/files/<path:filename>",
+    "record_from_pid": "/pids/<pid_type>/<path:pid_value>",
```
lnielsen, line 438: Question: it's not possible to find a way to make them root-level instead?

```/doi/....```

Comment (lnielsen) [reply invenio_app_rdm/config.py:438]: Right now we only need for DOIs, so we could hard-coded the DOI prefix.

File: invenio_app_rdm/records_ui/views/__init__.py [hunk 2]
```diff
@@ -47,6 +47,11 @@ def create_blueprint(app):
         view_func=record_latest,
     )
 
+    blueprint.add_url_rule(
```
lnielsen, line 50: The service ought to be defined at this moment, hence you could do simply string formatting on the route here.

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -79,6 +79,26 @@ def view(**kwargs):
     return view
 
 
+def pass_record_from_pid(f):
+    """Decorate a view to pass the latest version of a record."""
+    @wraps(f)
+    def view(**kwargs):
+        pid_type = kwargs.get('pid_type')
+        pid_value = kwargs.get('pid_value')
+        pid_client = kwargs.get('client')
```
lnielsen, line 88: Where is the ``client`` coming from? I don't see it in the URL pattern?

invenio_app_rdm/config.py
ppanero: We can I just thought that having it generic was better... and we can not put a root level variable becuase it will conflict with some `/records/<pid_value>`...

I will hardcode doi then.

invenio_app_rdm/records_ui/views/__init__.py
ppanero: What do you mean, with what purpose? I do not follow... sorry

invenio_app_rdm/records_ui/views/decorators.py
ppanero: you are right... it was meant for multi-client, and should be something like `/doi/provider.1234-abdce?client=zenodo`... However, for that I would need to access `requests.args`, which is not super clean. Any other ideas?

invenio_app_rdm/config.py
ppanero: Done

invenio_app_rdm/records_ui/views/__init__.py
lnielsen: It was if you needed the list of configured pid providers to put them in the URL pattern.

Now you have ``"/doi/<path:pid_value>"``, but it could be something like ``"/{schemes}/<path:pid_value>".format(schemes="|".join(pidschemes))

Let's talk IRL

invenio_app_rdm/records_ui/views/decorators.py
lnielsen: The PID resolution i think should be independent of the provider.

ppanero:
Changes implemented, depends on https://github.com/inveniosoftware/invenio-rdm-records/pull/560 to be tested

ppanero:
Working!
Screenshot 2021-04-28 at 13 50 39

invenio_app_rdm/records_ui/views/__init__.py
```diff
@@ -47,6 +47,19 @@ def create_blueprint(app):
         view_func=record_latest,
     )
 
+    with app.app_context():
```
lnielsen, line 50: ``app.extensions['invenio-rdm-records'].records_service.config`` to avoid pushing an app context

lnielsen (approved):

PR #804, opened 2021-04-27
views: add quick-create (plus) menu

slint:
Closes https://github.com/inveniosoftware/invenio-communities/issues/229.


slint:
image

slint:
We should review what kind of Jinja blocks/files could be refactored in order to minimize code duplication for this small menu. I can imagine moving this to `invenio-theme` (since registering a menu under the `plus.` namespace is optional, and only done in RDM right now).

PR #805, opened 2021-04-27
Add doi in versions

ines-cruz:
* add doi in the ui in the versions section

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js
@@ -7,9 +7,9 @@
 // under the terms of the MIT License; see LICENSE file for more details.
 
 import axios from "axios";
+import _get from "lodash/get";
 import React, { useEffect, useState } from "react";
 import { Divider, Grid, Icon, Message, Placeholder } from "semantic-ui-react";
-
 const deserializeRecord = (record) => ({
   id: record.id,
   parent_id: record.parent.id,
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js
```diff
@@ -36,6 +37,11 @@ const RecordVersionItem = ({ item, activeVersion }) => {
             {item.publication_date}
           </p>
         </Grid.Column>
+        <Grid.Column width={12}>
```
ntarocco, line 40: In semantic-ui, columns are max `16`. I suspect you wanted to go to a new row here.
In that case, you have to:
```
</Grid.Row>
<Grid.Row columns={1}>
...
</Grid.Row>

File: invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js [hunk 2]
```diff
@@ -36,6 +37,11 @@ const RecordVersionItem = ({ item, activeVersion }) => {
...
+          <p className="text-muted font-small">
+            {_.get(item.pids, "doi.identifier", "")}
```
ntarocco, line 42: I don't see the import of `get`: you should always import specific function whenever possible to take advantage of the [tree-shaking feature](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking).
```
import _get from 'lodash/get';
```

ntarocco (commented):
Good job! It would be great if you add a screenshot to see the changes :)

ines-cruz:
screenshot
In the versions section the doi is visible according to the version

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js
ines-cruz: Thank you. I changed it in the other commit

ppanero:
Styling-wise, basing on Zenodo, I think the DOI should be shown in the same `Row` component, but in a different line. See screenshot :)

Screenshot 2021-04-28 at 10 31 36

ines-cruz:
test

ntarocco (approved):

PR #806, opened 2021-04-27
deposit: inject PIDs field config

ntarocco:
Blocked by release of https://github.com/inveniosoftware/react-invenio-deposit/pull/273

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -12,6 +12,7 @@
 from flask import current_app, render_template
 from flask_login import login_required
 from invenio_i18n.ext import current_i18n
+from invenio_rdm_records.proxies import current_rdm_records
 from invenio_rdm_records.resources.serializers import UIJSONSerializer
 from invenio_rdm_records.services.schemas import RDMRecordSchema
 from invenio_rdm_records.services.schemas.utils import dump_empty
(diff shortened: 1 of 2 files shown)


ntarocco:
Any idea how to override such config in `my-site`?

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -24,11 +25,53 @@
...
+        # all providers disabled for this scheme
+        if not provider_enabled:
+            continue
+
+        u_scheme = scheme.upper()
+
+        pids_provider = {
+            "scheme": scheme,
+            "pid_label": u_scheme,
+            "pid_placeholder": "10.1234/rdm.123456",
```
ppanero, line 56: minor: the dafault provider is called "datacite", and I think is nice to know the DOI will come form there?

```suggestion
            "pid_placeholder": "10.1234/datacite.123456",
```

ppanero, line 56: minor: the dafault provider is called "datacite", and I think is nice to know the DOI will come from there?

```suggestion
            "pid_placeholder": "10.1234/datacite.123456",
```

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -24,11 +25,53 @@
...
+            "can_be_managed": can_be_managed,
+            "can_be_unmanaged": can_be_unmanaged,
+            "btn_label_get_pid": f"Get a {u_scheme} now!",
+            "managed_help_text": f"Reserve a {u_scheme} or leave this " \
```
ppanero, line 60: minor: This should be "managed" and "required", right? i.e. even if a pid type is managed if there is no required provider, it wont automatically get a pid.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -392,26 +375,28 @@ export class RDMDepositForm extends Component {
...
+                        pidType={pid.scheme}
+                        unmanagedHelpText={pid.unmanaged_help_text}
                       />
                       <Divider />
                     </Fragment>
                   ))}
                   <IdentifiersField
                     fieldPath="metadata.identifiers"
-                    label="Identifier(s)"
+                    label="Alternate identifier(s)"
```
ppanero, line 399: 🚀

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -24,11 +25,53 @@
...
+            if provider_attrs["system_managed"]:
+                can_be_managed = True
+            else:
+                can_be_unmanaged = True
+
+        # all providers disabled for this scheme
+        if not provider_enabled:
+            continue
+
+        u_scheme = scheme.upper()
```
ppanero, line 51: minor: This is good for now, but I think there might be cases when is not full upper case (e.g. `PubMed`).

ppanero (approved):
LGTM, however, the fact that `get_form_pids_config` is quite cumbersome is telling me that we might need to revisit the config in the future.

invenio_app_rdm/records_ui/views/deposits.py
ntarocco: 👍🏼

invenio_app_rdm/records_ui/views/deposits.py
ntarocco: For the time being, I did not add a `required` to the UI: if the user will not insert/reserve anything, and the backend has the managed required, it will automatically create a DOI.

invenio_app_rdm/records_ui/views/deposits.py
ntarocco: True... in any case this requires some more work to override such config...

PR #808, opened 2021-04-28
WIP: Add datacite to export section

ines-cruz:
export2

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -448,7 +448,7 @@ APP_RDM_RECORD_EXPORTERS = {
                        "UIJSONSerializer")
     },
     "datacite": {
-       "name": _("DataCite"),
+       "name": _("DataCite JSON"),
        "serializer": ("invenio_rdm_records.resources.serializers:"
                       "DataCite43JSONSerializer")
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/config.py
```diff
@@ -446,9 +446,15 @@ def files_rest_permission_factory(obj, action):
         "name": _("JSON"),
         "serializer": ("invenio_rdm_records.resources.serializers:"
                        "UIJSONSerializer")
-    }
+    },
+    "datacite": {
+       "name": _("DataCite"),
```
lnielsen, line 451:
```suggestion
       "name": _("DataCite JSON"),
```

ines-cruz: I wrote DataCite so it would be the same as in Zenodo, should i still change it to DataCite JSON?

lnielsen: It's because we'll have both a DataCite XML and DataCite JSON whereas Zenodo only has a DataCite XML.

ppanero:
closed in favour of https://github.com/inveniosoftware/invenio-app-rdm/pull/831

PR #813, opened 2021-04-28
upgrade: provide upgrade script from 2.0 to 3.0

fenekku:
Script to upgrade from InvenioRDM 2.0 to 3.0

Everything seemed to work fine after following the new upgrade instruction (PR in docs pending). (See comment in line for only note).

### DIFF
--- a/invenio_app_rdm/upgrade_scripts/migrate_2_0_to_3_0.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_2_0_to_3_0.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+# Copyright (C) 2021 Northwestern University.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""Record migration script from InvenioRDM 2.0 to 3.0.
+
+Disclaimer: This script is intended to be executed *only once*, namely when
+upgrading from InvenioRDM 2.0 to 3.0!
+If this script is executed at any other time, probably the best case scenario
+is that nothing happens!
+"""
+
+from invenio_db import db
+from invenio_rdm_records.records.api import RDMDraft, RDMParent, RDMRecord
+
+
+def execute_upgrade():
+    """Execute the upgrade from InvenioRDM 2.0 to 3.0.
+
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/upgrade_scripts/migrate_2_0_to_3_0.py
```diff
@@ -0,0 +1,66 @@
...
+    """Execute the upgrade from InvenioRDM 2.0 to 3.0.
+
+    Please read the disclaimer on this module before thinking about executing
+    this function!
+    """
+    for record_metadata in RDMRecord.model_cls.query.all():
+        record = RDMRecord(record_metadata.data, model=record_metadata)
+
+        # Updating to new $schema: commit() will take care of adding it
+        record.pop("$schema", None)  # so that new schema replaces it
```
fenekku, line 30: The previous upgrade used this method to update the `$schema`, so I kept with it... It's a little strange/iffy: we wipe the `$schema` out and rely on any future changes to the records to require a save draft (and publish) first to update the schema. Until then, there is no schema in the record's json column in the db. 

Because `schema` is a `ConstantField`, we can't do `record.schema = "<new schema"` either. *Maybe* we could do `record["$schema"] = "<new schema>"`, but I didn't go there since the above works and I didn't encounter any problem related to it yet.

invenio_app_rdm/upgrade_scripts/migrate_2_0_to_3_0.py
```diff
@@ -0,0 +1,66 @@
...
+    """Execute the upgrade from InvenioRDM 2.0 to 3.0.
+
+    Please read the disclaimer on this module before thinking about executing
+    this function!
+    """
+    for record_metadata in RDMRecord.model_cls.query.all():
+        record = RDMRecord(record_metadata.data, model=record_metadata)
+
+        # Updating to new $schema when eventually saved
+        record.pop("$schema", None)
```
fenekku, line 30: The previous upgrade used this method to update the `$schema`, so I kept with it... 

It's a little strange/iffy: we wipe the `$schema` out and rely on any future changes to the records to require a save draft (and publish) first which will update the `$schema` back. Until then, there is no `$schema` in the record's json column in the db. 

Because the `schema` systemfield is a `ConstantField`, we can't do `record.schema = "<new schema"` either. *Maybe* we could do `record["$schema"] = "<new schema>"`, but I didn't go there since the above works and I didn't encounter any problem related to it yet.

Issue #834, opened 2021-05-06
ui: header breaks when email is long and window size small


ppanero:
Package version (if known): v3.0.0

## Describe the bug

Header breaks when email is long and window size small

## Steps to Reproduce
1. Register with a long email (e.g. ppanero@awesome-rdm.com)
2. Go to the main page
3. Reduce screen size width up to the point of having the menu (uploads, communities) in a buger manu

## Screenshots (if applicable)

<img width="897" alt="Screenshot 2021-05-06 at 10 44 27" src="https://user-images.githubusercontent.com/6756943/117269174-59178200-ae58-11eb-8ce7-7b7e070e5350.png">

PR #835, opened 2021-05-06
Fix header break with long email

ines-cruz:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/834
Changed the position of the plus button when the window is resized so it wouldn't stay on top of the user email and make it go below the header. Added some spacing between the plus button and the user email so they wouldn't overlap

header1
header2


ppanero (commented):
Nice work! However, I think we should aim for a more generic "@media" fix, which is more complex and will require more time... Should we tackle all this RDM related in another time?

Mainly to have a centralized point of CSS (more maintainable in the long run)... otherwise we might not know where styling is coming from...

mitsosf:
This is addressed in #902

PR #839, opened 2021-05-11
Fix German help-search translation

chriz-uniba:
Based on the comments that I made [here](https://github.com/tu-graz-library/invenio-config-tugraz/pull/61/files) (unfortunately after the PR was already merged), I propose these fixes and improvements to the German help-search translation.


mb-wali (approved):
Thanks! for fixing these typos :)
looks much better.

Issue #856, opened 2021-05-25
link subject pills to filtered searches


ines-cruz:
Related to https://github.com/inveniosoftware/invenio-app-rdm/issues/415

Click on subject pills and query by that subject

lnielsen:
Replaced by #1226

PR #858, opened 2021-05-26
link creator's name with search 

### LINKED ISSUE #415: link pills to filtered searches
From Dan (Northwestern also wants this)

Would be nice if you could click on an author name (again in detail view, possibly elsewhere) to query by that author. Same for subject pills. [editor: and other pills too]

ines-cruz:
closes #415


ines-cruz:
part of the task ( the linking the subjects) will be done in this task: #856

ines-cruz:
search1
search2
search3

I added a record with both a person and an organization.
( both of them have spaces in their name and the names are unique to all other records.)
If I click the organization it leads me to the second screenshot where just this record is shown, but if i click on the name other records are shown. 
Should I change the search parameters?

PR #862, opened 2021-05-27
 ux: overlapping subject pills

ines-cruz:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/296


ines-cruz:
With different screen sizes:
sub1
sub2
sub3

ppanero (commented):
It looks good... however, I would try to avoid putting style in the components (I would go for css classes + style in css files)...

I also recall an issue for long email addresses that was not properly displayed.... I am wondering if we should tie all these issues together and address along with responsivenes...

mitsosf:
This is addressed in #902

PR #885, opened 2021-06-04
upgrade script: Adds migration of languages

jrcastro2:


ppanero (approved):
Works for me, thanks a lot!

Created 2 drafts and 3 records with several add titles and descs with languages then migrated

Before the upgrade
Screenshot 2021-06-04 at 12 52 30
Screenshot 2021-06-04 at 12 52 15

After the upgrade
Screenshot 2021-06-04 at 13 36 44

ntarocco (approved):

PR #887, opened 2021-06-07
celery: update expired embargos

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/400

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -30,6 +30,7 @@ WARNING: An instance should NOT install multiple flavour extensions since
 
 from datetime import datetime, timedelta
 
+from celery.schedules import crontab
 from flask_principal import Denial
 from invenio_access.permissions import any_user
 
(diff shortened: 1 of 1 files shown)


ppanero (approved):

invenio_app_rdm/config.py
```diff
@@ -265,6 +266,10 @@ def files_rest_permission_factory(obj, action):
     'draft_resources': {
         'task': 'invenio_draft_resources.tasks.cleanup_drafts',
         'schedule': timedelta(minutes=60),
+    },
+    'rdm_records': {
+        'task': 'invenio_rdm_records.tasks.update_expired_embargos',
+        'schedule': crontab(minute=2, hour=0),
```
ppanero, line 272: nice! 🚀

ntarocco, line 272: Silly question: does it mean at midnight and 2 minutes? Was the minute `2` chosen for some reason?

jrcastro2: > does it mean at midnight and 2 minutes?

Yes

>  Was the minute `2` chosen for some reason?

No reason at all (Copied from [Zenodo](https://github.com/zenodo/zenodo/blob/master/zenodo/config.py#L181-L185))

ppanero: I *think* is because embargoes will expire right after midnight? e.g. they set a date.

PR #902, opened 2021-06-14
ui: overlap in subject pills & long email fix

mitsosf:


mitsosf:
References #862 & #834

mitsosf:
The subjects are now properly aligned (1) and long emails don't break the style anymore (2):
image

ntarocco:
I tested it out and it looks much better thank you!
There is a point where it breaks, maybe worth it to slightly change the media query?
See:
Screenshot 2021-06-18 at 17 32 16

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/navbar.less
```diff
@@ -46,6 +46,17 @@ nav {
   padding-bottom: 8px;
 }
 
+@media only screen and (min-width: 745px) and (max-width: 1000px){
```
ntarocco, line 49: I tested with `1200px` and it looks like is fixes the issues:
```suggestion
@media only screen and (min-width: 745px) and (max-width: 1200px){
```
WDYT?

mitsosf:
@ntarocco 
I also thought that someone may have a very long email, even longer than the one you used for the review. In that case, it may make sense to cut the last part of the email address or something eg.

superlongemailthatwouldtakealotofspace@longemail.com

Could be 
superlongemailthatwouldtakealoto...
or
superlongemail...@longe....com

Not sure if it's worth the efforts/resources, was just thinking about limits

ntarocco:
> @ntarocco
> I also thought that someone may have a very long email, even longer than the one you used for the review. In that case, it may make sense to cut the last part of the email address or something eg.
> 
> [superlongemailthatwouldtakealotofspace@longemail.com](mailto:superlongemailthatwouldtakealotofspace@longemail.com)
> 
> Could be
> superlongemailthatwouldtakealoto...
> or
> superlongemail...@longe....com
> 
> Not sure if it's worth the efforts/resources, was just thinking about limits

You are right indeed. I would keep for the moment like this, and if we have issues reported we will probably do what you suggest! :)

Issue #909, opened 2021-06-17
share: can_preview link mentions unpublished changes


ppanero:
From discussion on https://github.com/inveniosoftware/invenio-app-rdm/issues/876

*can_preview* link should not display a banner mentioning unpublished changes (Since they are published).

![](https://user-images.githubusercontent.com/39839586/122281201-9f2c2080-ceaf-11eb-93fb-7b1101297c71.png)

**Steps**
1- Go to uploads, save and publish a record
2- Share --> can_preview link
3- Go to that link (not-logged in user)
4- See that the banner says changes are not published

**Context**
It is most likely due to it using the same Jinja template than the desposit form *preview* functionality.

PR #911, opened 2021-06-17
preview: Bug fixes

jrcastro2:
* Back to edit button only available
  if user has edit permissions

* Updated message if previewing a
  published record without a draft

* Version banner is only displaying if
  we are not previewing the latest version

* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/909

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -28,15 +28,13 @@
               <i class="eye icon"></i> {{_ ("Preview")}}
             </div>
             <p>
-              {% if record.is_published %}
+              {% if not is_draft %}
+               You are previewing a published record.
+              {% elif record.is_published %}
                 You are previewing changes that have not yet been published.
-              {% endif %}
-
-              {% if not record.is_published and record.versions.index == 1 %}
+              {% elif not record.is_published and record.versions.index == 1 %}
                 You are previewing a new record that has not yet been published.
-              {% endif %}
-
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -28,15 +28,13 @@
               <i class="eye icon"></i> {{_ ("Preview")}}
             </div>
             <p>
-              {% if record.is_published %}
+              {% if not is_draft %}
+               You are previewing a published record.
```
fenekku, line 32: It seems like if we give a `can_view` permission link for a private record to someone, then that person will see this preview banner even though they are not "previewing", but viewing...

Honestly I think the problem lies in the overlapping semantics of can_preview / can_view. The semantics should be: 
- `can_preview` is about being able to see a draft on the landing page
- `can_view` is about being able to see a record on the landing page

To me, the simple solution would be changing the below from:

```
You are previewing changes that have not yet been published.
``` 

to 

```
You are previewing changes that may not have been published yet.
```
i.e. we are not showing a visual diff, so we have no way of telling people *what* about the draft is different from the published record. At best we can tell them that some of the draft may not have been published.

As an aside, personally I am not super convinced about the need of `can_preview` if a depositor can simply publish as private and then give `can_view` link to others.

jrcastro2: After discussion, new issue created <https://github.com/inveniosoftware/invenio-app-rdm/issues/912>.

lnielsen:
As agreed IRL, we're merging this one, and have opened #912 to fix larger issues with preview sharing.

PR #917, opened 2021-06-23
vocab: adds role vocab to form config

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/670
* needs https://github.com/inveniosoftware/invenio-app-rdm/pull/920 (for tests)


ppanero:
Futuristically speaking we should remove the underscores, right? `resource_type` --> `resourcetype` and `title_type` --> `titletype`?

jrcastro2:
> Futuristically speaking we should remove the underscores, right? `resource_type` --> `resourcetype` and `title_type` --> `titletype`?

Yes, was planning to do this in a separate PR or maybe I can add it to next one (since it needs changes in rdm-records too)

ppanero:
@jrcastro2 I'm merging this then, let's open an issue and treat it separatly so we dont polute another PR with refactors

PR #937, opened 2021-07-07
views: use proper locale negotiation

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/709


ppanero (approved):
Is it the only appearance we have of this? I couldn't find it anywhere else, so I think we are covered

jrcastro2:
> Is it the only appearance we have of this? I couldn't find it anywhere else, so I think we are covered

Yes, I only could find this one.

PR #949, opened 2021-07-15
landing page: update licenses to use UIObject

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/718

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -22,7 +22,6 @@ from invenio_vocabularies.records.models import VocabularyScheme
 from marshmallow_utils.fields.babel import gettext_from_dict
 from sqlalchemy.orm import load_only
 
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -169,6 +169,7 @@ def get_form_config(**kwargs):
     return dict(
         vocabularies=vocabularies,
         current_locale=str(current_i18n.locale),
+        default_locale=BABEL_DEFAULT_LOCALE,
```
lnielsen, line 172: This needs to come from ``current_app.config.get('BABEL_DEFAULT_LOCALE', 'en')`` instead, and I can see it's a bug that was introduced earlier so you probably also have to fix it higher up in the file.

Otherwise instances cannot change the default language.

jrcastro2: Ah. yes it makes sense, thanks!

PR #950, opened 2021-07-15
records: remove use of old vocabularies

lnielsen:


lnielsen:
Needs https://github.com/inveniosoftware/invenio-rdm-records/pull/745

PR #954, opened 2021-07-15
style: fixes landingpage & search page header margin

### LINKED ISSUE #951: ux: inconsistent spacing to header
The spacing marked on the two screenshots differs. The landing page has a ``15px`` top margin which shouldn't be there.

Screenshot 2021-07-15 at 11 57 15
Screenshot 2021-07-15 at 11 57 47

mb-wali:
closes #951 

**Searchpage**
querypage

**Langingpage**
landing

The reason behind this margin is to allow enough space between content and notification bar: required at least `2px` to seperate them see also - screenshot
2px


lnielsen:
> The reason behind this margin is to allow enough space between content and notification bar: required at least 2px to seperate them see also - screenshot

Got it, so then likely the margin should be a bottom margin on the notification bar instead so. With the reduced margin, the notification looks strange.

mb-wali:
> > The reason behind this margin is to allow enough space between content and notification bar: required at least 2px to seperate them see also - screenshot
> 
> Got it, so then likely the margin should be a bottom margin on the notification bar instead so. With the reduced margin, the notification looks strange.

without-notification
with-notification

Issue #953, opened 2021-07-15
JSON export should *not* use UI json serialised


lnielsen:
Expect instead that it uses the normal JSON serializer without the ``ui`` key.

![Screenshot 2021-07-15 at 12 46 34](https://user-images.githubusercontent.com/1698163/125775975-581d9a72-71c8-4dc5-8e29-d0b27b4c84c7.png)

PR #955, opened 2021-07-15
exports: json should not use ui serializer

ppanero:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/953

Uses bare JSONSerializer from Flask-Resources

Screenshot 2021-07-15 at 14 07 10

The description shows the HTML tags. I think this is correct?
```
    "description": "<h2>&lt;h1&gt;Normal description&lt;/h1&gt;</h2><p><strong>this should be bold</strong></p>"
```

**Open Question**

Why does the UISerializer not use MarshmallowJSONSerializer as parent since it is anyway defining a scheme? Logic could be simplified to add the UI key with Marshmallow schemes.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,7 +20,9 @@ from invenio_accounts.proxies import current_datastore
 from invenio_accounts.testutils import login_user_via_session
 from invenio_db import db
 from invenio_rdm_records.proxies import current_rdm_records
+from invenio_vocabularies.contrib.subjects.api import Subject
 from invenio_vocabularies.proxies import current_service as vocabulary_service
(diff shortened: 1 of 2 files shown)


lnielsen:
Probably it just needs a rebase on latest master.

tests/conftest.py
```diff
@@ -133,7 +135,7 @@ def resource_type_type(app):
 @pytest.fixture(scope="module")
 def resource_type_item(app, resource_type_type):
```
ppanero, line 136: Should we expose this fixtures in rdm-records and use them here?

Issue #962, opened 2021-07-16
deposit form: scheme should use correct captilization


lnielsen:
Schemes do not use correct captilization in drop-downs and errors (e.g. alternate identifiers).

## Expected behavior

ARXIV -> arXiv
BIBCODE -> Bibcode
HANDLE -> Handle

There might be more. Check IDUtils 

## Screenshots (if applicable)

![Screenshot 2021-07-16 at 06 59 27](https://user-images.githubusercontent.com/1698163/125894214-a6b533b9-b17c-4aad-a6b3-7dafc5681698.png)

Also in errors it's lower-cased instead:

![Screenshot 2021-07-16 at 07 04 44](https://user-images.githubusercontent.com/1698163/125894699-95381a43-152d-4925-8c13-a9ec2e8496dc.png)

PR #971, opened 2021-07-16
Identifiers

ppanero:
- requires https://github.com/inveniosoftware/invenio-rdm-records/pull/753
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/962

*Open Questions* (EDIT: Answered in comments, went for the alternative below)
- The labels (for all vocabularies) will be a problem if the user customizes the vocabularies (e.g. roles). We might want the components to do a `read_many/all` and do some logic.
- The error message comes from the backend (rdm-records --> marshmallow-utils), therefore we need to propagate labesl. Shall this become a vocabulary? We will have the same problem than the point above. Just the propagation would be solved by making it into a vocabulary.

**Alternative**
Make them into environment/config variables that can be changed by the user so labels are always found.


lnielsen:
The identifier schemes are not vocabularies so they shouldn't be affected?

Probably the backend should define valid values, and the values should be used by Invenio-App-RDM to be dumped here?

ppanero:
@lnielsen so basically we go for the "alternative":

> Make them into environment/config variables that can be changed by the user so labels are always found.

That way they can be defined in rdm-records (the validators can use them), be overwritten by users, and also be injected into the UI via config.

ppanero:
Scheme labels are taken from config
Screenshot 2021-07-19 at 13 54 16

**Open questions**:
- Would it be preferable just adding identifier labels to the config and let the frontend build the full dict (i.e. relations + resource type + identifiers schemes) or is it ok in the backend
- As you can see in the photo some widths overlap the `x` for deletion with the label (see the `Bibcode` in the bottom of the previous image). This was already there, should be tackled in a separate issue.

lnielsen (approved):

PR #978, opened 2021-07-19
i18n: adds german translations

mb-wali:


invenio_app_rdm/theme/assets/semantic-ui/translations/invenio_app_rdm/i18next.js
```diff
@@ -23,6 +23,9 @@ const options = {
     // cache user language off
     caches: [],
   },
+  react: {
+    transKeepBasicHtmlNodesFor: []
```
zzacharo, line 27: Maybe add a comment here on why we set it to empty array?

PR #989, opened 2021-07-20
styling: fix txt and img alignment

### LINKED ISSUE #957: detail page: orcid icon misaligned
The author name should be vertically centered on the ORCID name:

Screenshot 2021-07-15 at 19 03 54

mb-wali:
did not use the existing class - since it's used other places - just to be sure we are not creating any other bug.

closes #957 

**screenshot**
inline

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/landing_page.less
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/landing_page.less
@@ -62,6 +62,10 @@
      height: @font-size-base;
  }
 
+ .inline-orcid-icon {
+    margin-bottom: 3px;
+  }
+
  .thin-line {
      margin: 0rem 0rem 0.5rem 0rem !important;
  }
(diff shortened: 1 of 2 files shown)


fenekku (commented):
Thanks for this, I know these alignments are a pain. You'll forgive me if these are 3 nitpicks :smile: :
- could we rename `inline-text` to `inline-orcid-icon` or `text-aligned-orcid-icon` or something like this since it's the icon that is aligned and not the text
- can we place the rule in `landing_page.less` so it's tied closer to where it's used
- (harder) could we reduce the horizontal spacing between the icon and the author. They are a bit further apart than expected for nice grouping.

If the last one is too involved, let me know and we can take it up separately/we can merge the PR without it.

mb-wali:
> Thanks for this, I know these alignments are a pain. You'll forgive me if these are 3 nitpicks  :
> 
> * could we rename `inline-text` to `inline-orcid-icon` or `text-aligned-orcid-icon` or something like this since it's the icon that is aligned and not the text
> * can we place the rule in `landing_page.less` so it's tied closer to where it's used
> * (harder) could we reduce the horizontal spacing between the icon and the author. They are a bit further apart than expected for nice grouping.
> 
> If the last one is too involved, let me know and we can take it up separately/we can merge the PR without it.

hey, i think you can merge this - for the last one I will create an issue.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/landing_page.less
```diff
@@ -62,6 +62,10 @@
      height: @font-size-base;
  }
 
+ .inline-orcid-icon {
```
ppanero, line 65: couldn't it e added to the `inline-orcid` class?

PR #1004, opened 2021-07-21
resource_type: update dumping

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/760
* needs https://github.com/inveniosoftware/invenio-vocabularies/pull/77

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -8,7 +8,7 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """Routes for record-related pages provided by Invenio-App-RDM."""
-
+from elasticsearch_dsl import Q
 from flask import current_app, render_template
 from flask_login import login_required
 from invenio_access.permissions import system_identity
@@ -72,10 +72,14 @@ def get_form_pids_config():
     return pids_providers
 
 
-def _dump_resource_type_vocabulary():
+def _dump_resource_type_vocabulary(extra_filter):
     """Dump resource type vocabulary."""
     results = vocabulary_service.read_all(
-        system_identity, fields=["id", "props"], type='resourcetypes')
+        system_identity,
+        fields=["id", "props"],
+        type='resourcetypes',
+        extra_filter=extra_filter
+    )
     return [
         {
             "icon": r["props"].get("type_icon", ""),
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -113,6 +113,7 @@ export class RDMDepositForm extends Component {
         },
         identifiers: {
           ...this.config.vocabularies.identifiers,
+
```
fenekku, line 116: that's weird :)

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -86,6 +90,22 @@ def _dump_resource_type_vocabulary():
...
+        extra_filter=~Q('exists', field="props.invenio_type") |
+        Q('term', props__invenio_type="linkable")
+    )
+
+
+def _dump_depositable_resource_type_vocabulary():
+    """Dump depositable resource type vocabulary."""
+    return _dump_resource_type_vocabulary(
+        extra_filter=~Q('exists', field="props.invenio_type") |
+        Q('term', props__invenio_type="depositable")
```
fenekku, line 105: It's a little strange to me that at the end of the day we have to construct an ElasticSearch query in the UI views code to get the resource_types... (I understand some of the context... still perhaps valuable to revisit)

fenekku (approved):
sure

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
jrcastro2: Leftover from a rebase :sweat_smile: . Thanks!

invenio_app_rdm/records_ui/views/deposits.py
ppanero: I agree with this. I think it sort of breaks the resources/service paradigm. The filter should be somehow constructed at service level and this layer (view) use said function. Maybe a function is not the way and it can be an attribute or something idk. But I agree it is weird to see a elasticsearch query instantiation in the UI layer.

jrcastro2: I see your point but I cannot think of a nice a solution. 
I would say that we can crate a new folder in [`invenio-rdm-records/services`](https://github.com/inveniosoftware/invenio-rdm-records/tree/master/invenio_rdm_records/services) called vocabularies and there we can build the specific queries to retrieve both cases. WDYT?
(any suggestion is welcomed)

Issue #926, opened 2021-06-29
Vocabulary values migration


jrcastro2:
Fields that were converted to vocabularies and need the values to be migrated:
- [x] creators.role
- [x] contributors.role
- [x] additional_titles.title_type
- [x] additional_titles.lang
- [x] additional_descriptions.title_type
- [x] additional_descriptions.lang
- [ ] related_identifiers.resource_type --> existing in the v3 to v4 script
- [x] related_identifiers.relation_type
- [x] creators.affiliations
- [x] contributors.affiliations
- [x] subjects
- [x] dates.date_type

PR #1015, opened 2021-07-23
upgrade: script record upgrade from v4 to v5

ppanero:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/926

### DIFF
--- a/invenio_app_rdm/upgrade_scripts/migrate_4_0_to_6_0.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_4_0_to_6_0.py
@@ -1,0 +1,374 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Record migration script from InvenioRDM 4.0 to 6.0.
+
+Disclaimer: This script is intended to be executed *only once*, namely when
+upgrading from InvenioRDM 4.0 to 6.0!
+If this script is executed at any other time, probably the best case scenario
+is that nothing happens!
+"""
+from pathlib import Path
+
+import yaml
+from invenio_access.permissions import system_identity
+from invenio_db import db
+from invenio_rdm_records.records.api import RDMDraft, RDMRecord
+from invenio_vocabularies.proxies import current_service
+from invenio_vocabularies.records.models import VocabularyScheme, \
+    VocabularyType
+
+
+def migrate_vocabularies():
+    """Migrate old vocabularies."""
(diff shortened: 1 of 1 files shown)


ppanero:
Discussed IRL:

1- Flag check: Do you have subjects or affiliations with more than "name/subject". i.e. will they cause troubles?
2- If yes, give a recipe to fix them
3- If not (or after fix) run the script and assume:
    3.1 - id is present: is a vocabulary (note that affiliations will contain id+name in the dumped data)
    3.2 - name/subject is present: is s custom term

invenio_app_rdm/upgrade_scripts/migrate_4_0_to_6_0.py
```diff
@@ -0,0 +1,350 @@
...
+                "contributors": inv_affs_contributors
+            }
+
+    invalid_affiliations = {}
+    if invalid_affiliations_rec:
+        invalid_affiliations["record"] = invalid_affiliations_rec
+    if invalid_affiliations_draft:
+        invalid_affiliations["draft"] = invalid_affiliations_draft
+
+    print(
```
zzacharo, line 115: Let's refactor this part a bit to something like
```python
if invalid_affiliations:
        print(
            f"Your instance has {total} affiliations that need to be fixed. " +
            "Check the invalid_affiliation.yaml file for more details."
        )
        with open('invalid_affiliations.yaml', 'w') as f:
            yaml.dump(list(invalid_affiliations), f)
else:
        print(f"Your instance has no affiliations that need to be fixed.")
```

ppanero: We should also keep a `ROR` counter to tell the user if they need to index the ROR vocabulary or they can do with only free text affiliations.

ppanero: Done :)

invenio_app_rdm/upgrade_scripts/migrate_4_0_to_6_0.py
```diff
@@ -0,0 +1,364 @@
...
+            "Check the invalid_affiliation.yaml file for more details."
+        )
+        with open('invalid_affiliations.yaml', 'w') as f:
+            yaml.dump(list(invalid_affiliations), f)
+    else:
+        print(f"All your instance's affiliations are valid.")
+    if needs_ror:
+        print(
+            "You have affiliations with ROR identifiers, you need to " +
+            "add its vocabulary. Instructions to do so are available in TODO"
```
zzacharo, line 131: Note not to forget to replace the `TODO`

zzacharo (commented):
LGTM. We just need to fill in the missing links!

PR #1032, opened 2021-07-30
ui: label resource types correctly [+]

fenekku:
'correctly' means:
- with translation
- still relying on cache
- by using parent entry title if need be
  (not relying on `subtype_name` and `type_name` anymore)

also:
- refactors vocabularies dumping for deposit page
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/785

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -78,145 +78,186 @@ def get_form_pids_config():
     return pids_providers
 
 
-def _dump_resource_type_vocabulary(extra_filter):
-    """Dump resource type vocabulary."""
-    results = vocabulary_service.read_all(
-        system_identity,
-        fields=["id", "props"],
-        type='resourcetypes',
-        extra_filter=extra_filter
-    )
-    return [
-        {
-            "icon": r["props"].get("type_icon", ""),
-            "id": r["id"],
-            "subtype_name": r["props"].get("subtype_name", ""),
-            "type_name": r["props"]["type_name"],
-        } for r in results.to_dict()["hits"]["hits"]
-    ]
+class VocabulariesOptions:
+    """Holds React form vocabularies options."""
 
+    def __init__(self):
+        """Constructor."""
+        self._vocabularies = {}
 
-def _dump_linkable_resource_type_vocabulary():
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -78,145 +78,186 @@ def get_form_pids_config():
...
+            return (type_labels[type_], self._get_label(hit))
+
+    def _resource_types(self, extra_filter):
+        """Dump resource type vocabulary."""
+        type_ = 'resourcetypes'
+        all_resource_types = vocabulary_service.read_all(
+            system_identity,
+            fields=["id", "props", "title"],
+            type=type_,
+            # Sorry, we have over 100+ resource types entry at NU actually
```
fenekku, line 118: Yeah... it's a miracle I remembered `read_all` doesn't actually read all...

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -78,145 +78,186 @@ def get_form_pids_config():
...
+        type_ = hit.get("props", {}).get("type")
+
+        if id_ == type_:
+            # dataset-like case
+            return (self._get_label(hit), "")
+        elif type_ not in type_labels:
+            # safety net to generate a valid type, subtype and not break search
+            return (self._get_label(hit), "")
+        else:
+            return (type_labels[type_], self._get_label(hit))
```
jrcastro2, line 109: This seems a bit complex/specific for our resource type values, could we simplify it to this?
```suggestion
    def _get_type_subtype_label(self, hit):
        """Return (type, subtype) pair for this hit."""
        type_name = hit.get("props", {}).get("type_name")
        subtype_name = hit.get("props", {}).get("subtype_name")
        return type_name, subtype_name
```

jrcastro2 (approved):
Great refactor! :rocket: 
Just one minor comment.

invenio_app_rdm/records_ui/views/deposits.py
fenekku: Ah, those (`type_name` and `subtype_name`) are the fields we are trying not to use anymore, because they are not translations. We are trying to rely on `title` only and eventually we can phase them out.

PR #1035, opened 2021-08-02
migration script: add missing link

jrcastro2:


ppanero:
This link is incorrect. It should point to the new `customize/vocabularies/affiliations`. As is now the user will be following the prepare vocabularies section, execute the step and get a pointer (link) to the same section (Infinite loop!!!! ➿ 🔁 hehe)

PR #1049, opened 2021-08-04
global: enable facets and sort customization

lnielsen:
Depends on https://github.com/inveniosoftware/invenio-rdm-records/pull/795


invenio_app_rdm/records_ui/searchapp.py
```diff
@@ -0,0 +1,230 @@
...
+
+        # Nested facets
+        if 'childAgg' in ui:
+            ui['childAgg'].set_default('aggName', 'inner')
+            ui['childAgg'].set_default('title', title)
+
+        return ui
+
+
+class SearchAppConfig:
```
lnielsen, line 50: Taken from Invenio-Search-UI

https://github.com/inveniosoftware/invenio-search-ui/blob/master/invenio_search_ui/views.py#L59

ppanero: Much seems to be the same than `SearchAppInvenioRestConfigHelper` but in a slightly different manner. i.e. every method has one more line or one more option, many tiny changes.

Are we implying we are moving away from `invenio-search-ui`? 

If not, is it worth to create a common interface? It is just a bit weird to have two classes that do *almost exactly the same* in different modules. Maybe we can find a middle ground or migrate `invenio-shearch-ui` to the new one?

This can be shelved (i.e. open an issue) since we are in a rush for releasing.

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/search.html
```diff
@@ -14,98 +14,5 @@
...
-      "text": "{{_("Oldest")}}"
-    },
-    {
-      "sortBy": "version",
-      "text": "{{_("Version")}}"
-    }
-  ]
-}'></div>
-
+<div data-invenio-search-config='{{ search_app_config(app_id="rdm-search") | tojson }}'></div>
```
ppanero, line 17: `rdm-search`, this is *naming* and not from now... But what about `record-search`, since all are `rdm` search apps 😅

ppanero (approved):
2 shelvable ones. Config feels much more natural + doing it in one place is better than 2 (service + ui template) as before.

invenio_app_rdm/records_ui/searchapp.py
lnielsen: > Are we implying we are moving away from invenio-search-ui?

Yes. There's other problems like it's not applying proper permissions checks in search-ui and other in flexiiblities that means it's not worthwhile continue using it. Similar to that we removed usage of records-ui.

invenio_app_rdm/records_ui/searchapp.py
```diff
@@ -0,0 +1,230 @@
...
+
+from flask import current_app
+from invenio_rdm_records.searchconfig import OptionsSelector, \
+    SortOptionsSelector
+
+
+class SortConfig(SortOptionsSelector):
+    """Sort options for the search configuration."""
+
+    def map_option(self, key, option):
```
fenekku, line 21: I was suspecting an override like this may happen...

invenio_app_rdm/records_ui/searchapp.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+def search_app_context():
+    """Search app context processor."""
+    return {
+        'search_app_config': partial(
+            search_app_config, '/api/records', 'RDM_RECORD_SEARCH'),
+        'search_drafts_app_config': partial(
+            search_app_config, '/api/user/records', 'RDM_RECORD_SEARCH_DRAFTS')
+    }
```
fenekku, line 230: Could those be made to rely on the ServiceConfig? instead of having to extract the configurations again and be wary of variable names? But maybe the layers are too far apart...

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/citation.html
```diff
@@ -9,6 +9,6 @@
 <hr class="thin-line"></hr>
   <div id="recordCitation" data-record='{{ record | tojson }}'
     data-styles='{{ config.get("RDM_CITATION_STYLES") | tojson }}'
-    data-defaultstyle='{{ config.get("RDM_DEFAULT_CITATION_STYLE") | tojson }}'>
+    data-defaultstyle='{{ config.get("RDM_CITATION_STYLES_DEFAULT") | tojson }}'>
```
fenekku, line 12: :+1:

invenio_app_rdm/records_ui/searchapp.py
```diff
@@ -0,0 +1,230 @@
...
+class SearchAppConfig:
+    """Configuration generator for React-SearchKit."""
+
+    default_options = dict(
+        endpoint=None,
+        hidden_params=None,
+        app_id='rdm-search',
+        headers=None,
+        list_view=True,
+        grid_view=True,
```
fenekku, line 59: We had `grid_view=False` before. I think because the UI/frontend code for grid isn't quite there yet.

fenekku (approved):
Mostly makes sense. There is a lot of `config`s flying around now...

invenio_app_rdm/records_ui/searchapp.py
lnielsen: Yep, i've added a comment RDM-Records so that it's clear :-)

Issue #1051, opened 2021-08-04
deposit: preview explodes with errors in related identifiers


lnielsen:
## Steps to Reproduce
1. Create a new valid upload and fill in the related works.

![Screenshot 2021-08-04 at 23 58 17](https://user-images.githubusercontent.com/1698163/128260679-78fb19d4-b221-4a8e-96c8-cc26dca4218a.png)

2. Click on 'Save'
3. Remove the identifier
4. Click on "Preview"

![image](https://user-images.githubusercontent.com/1698163/128260809-30af8d33-74dd-4205-acd0-5bd925a7a13b.jpeg)

## Expected behavior

1. That the preview reports that there's a validation error on the page and that you cannot preview

or

2. That the preview simply leaves out the invalid identifier.

## Additional context

```
Traceback (most recent call last):
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/flask/app.py", line 2464, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/werkzeug/middleware/proxy_fix.py", line 169, in __call__
    return self.app(environ, start_response)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/werkzeug/middleware/dispatcher.py", line 66, in __call__
    return app(environ, start_response)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/flask/app.py", line 2450, in wsgi_app
    response = self.handle_exception(e)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/flask/app.py", line 1867, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
[...]

PR #1071, opened 2021-08-16
deposit: fix related identifiers preview bug

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1051

### DIFF
--- a/invenio_app_rdm/records_ui/views/records.py
+++ b/invenio_app_rdm/records_ui/views/records.py
@@ -11,13 +11,15 @@
 
 from os.path import splitext
 
-from flask import abort, current_app, redirect, render_template, request, \
+from flask import abort, current_app, g, redirect, render_template, request, \
     url_for
 from flask_login import current_user
 from invenio_base.utils import obj_or_import_string
 from invenio_previewer.extensions import default
 from invenio_previewer.proxies import current_previewer
+from invenio_rdm_records.proxies import current_rdm_records
 from invenio_rdm_records.resources.serializers import UIJSONSerializer
+from marshmallow import ValidationError
 
 from .decorators import pass_file_item, pass_file_metadata, pass_is_preview, \
     pass_record_files, pass_record_from_pid, pass_record_latest, \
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/records_ui/views/records.py
```diff
@@ -70,10 +70,17 @@ def open(self):
...
     files_dict = None if files is None else files.to_dict()
-
+    record_ui = UIJSONSerializer().serialize_object_to_dict(record.to_dict())
+    if is_preview:
+        # If is preview, only display values if required fields are filled in
+        record_ui['ui']['related_identifiers'] = [
+            rec for rec in record_ui['ui']['related_identifiers']
+            if rec.get('identifier') and rec.get('scheme')
+            and rec.get('relation_type')
+        ]
```
fenekku, line 80: Hmm, so although this bug was logged for `related_identifiers` it actually affects `dates`, `funding`, `references`, `alternate_identifiers`...

A small quick "solution" (which isn't really one, but might be good enough for depositors) is to move the switch-case https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/components/PreviewButton.js#L44 under `case FORM_SAVE_SUCCEEDED` only. This way clicking on preview with an invalid draft doesn't redirect to a breaking preview page. Obviously it's not ideal since one might visit the preview page via its url directly.  

A more complete alternative would be to edit the details macros to account for the possibility of missing subfields. But that's a chore and still error-prone.

Perhaps, easiest would be to also return (in addition to first quick solution) a 404 if `record_ui["errors"]` is not empty when `is_preview` i.e. the preview doesn't exist because a valid draft doesn't exist.  

Alternatives along these lines would work too.

jrcastro2: I see, I am not sure we can show the 404, since when visiting the preview page via the url directly we don't get the errors. 

As an alternative solution we could maybe add a cleaning function that cleans all the missing required values before previewing, WDYT?

(feel free to ping me in discord, probably easir to discuss IRL)

fenekku: discussed : we will try for a draft service method that just validates a draft so that we can get the `"errors"` in the `if is_preview` conditional. If there are any, we return a 404 page (seemed most appropriate). On the deposit page, we won't even redirect there if there are errors.

fenekku (approved):
:1st_place_medal:

PR #1086, opened 2021-08-31
dependencies: add invenio-logging with sentry sdk

ppanero:
Adds logging with sentry-sdk extra. It is not configured by default.

Tested with a new instance (v6.0) from scratch and no errors were shown. 

**Open questions**

- Do we want to add sentry to the docker-compose file? It would **not** be a zero-config thing, because the teams/users/project/dsn would still need to be created by the user
- Should docs be added somewhere in InvenioRDM or the Invenio ones are enough (those are the ones I followed).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,7 @@ install_requires = [
     'CairoSVG>=2.5.2,<3.0.0',
     f'invenio[base,auth,metadata,files]{invenio_version}',
     'invenio-communities>=2.5.0.dev2,<2.6.0',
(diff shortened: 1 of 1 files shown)


lnielsen:
> Do we want to add sentry to the docker-compose file? It would not be a zero-config thing, because the teams/users/project/dsn would still need to be created by the user

I would not add it at this point. 

> Should docs be added somewhere in InvenioRDM or the Invenio ones are enough (those are the ones I followed).

Later I would say.......first we need some more deployment docs

ppanero:
@lnielsen Should we merge this? The PR for the frontend part is independent.

PR #1087, opened 2021-08-31
frontend: enable sentry logging for react apps

ppanero:
- requires https://github.com/inveniosoftware/invenio-app-rdm/pull/1086

**issues**
- [security] the way we inject configuration means that the DSN would be exposed. This can be tamed at sentry project level (i.e. restrict origins).
- CSP/CORS gives some errors from the response of sentry that I haven't been able to fix.
- Still need to implement it for the search apps, but as a PoC is enough. The main issue with search apps is that the view is defined in `invenio-search-ui` and we cannot easily modify it to send the config along when rendering. Since the plan is to move out of `invenio-search-ui` this can be tackled then.

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/deposit.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/deposit.html
@@ -27,6 +27,10 @@
   {%- if permissions %}
   <input id="deposits-record-permissions" type="hidden" name="deposits-record-permissions" value='{{permissions | tojson }}'></input>
   {%- endif %}
+  {%- if sentry_config %}
+  <input id="sentry-config" type="hidden" name="sentry-config" value='{{sentry_config}}'></input>
+  {%- endif %}
+
   <div id="deposit-form"></div>
 {%- endblock page_body %}
 
--- a/invenio_app_rdm/theme/webpack.py
+++ b/invenio_app_rdm/theme/webpack.py
@@ -13,49 +13,51 @@ from invenio_assets.webpack import WebpackThemeBundle
 
 theme = WebpackThemeBundle(
     __name__,
-    'assets',
-    default='semantic-ui',
+    "assets",
+    default="semantic-ui",
     themes={
-        'semantic-ui': dict(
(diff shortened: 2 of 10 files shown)


invenio_app_rdm/theme/webpack.py
```diff
@@ -13,49 +13,51 @@
...
-                'prop-types': '^15.7.2',
-                'react-dnd': '^11.1.3',
-                'react-dnd-html5-backend': '^11.1.3',
-                'react-invenio-deposit': '^0.16.1',
-                'react-invenio-forms': '^0.8.7',
-                'react-dropzone': "^11.0.3",
-                'yup': '^0.32.9',
-                '@ckeditor/ckeditor5-build-classic': '^16.0.0',
-                '@ckeditor/ckeditor5-react': '^2.1.0',
+                "@ckeditor/ckeditor5-build-classic": "^16.0.0",
```
ppanero, line 33: sorry... had to order them alphabetically 😅

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/deposit.html
```diff
@@ -27,6 +27,10 @@
   {%- if permissions %}
   <input id="deposits-record-permissions" type="hidden" name="deposits-record-permissions" value='{{permissions | tojson }}'></input>
   {%- endif %}
+  {%- if sentry_config %}
```
ppanero, line 30: This might be common also with search at some point. Creating a common base file and adding this block there would be an option.

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/deposit.html
```diff
@@ -27,6 +27,10 @@
...
+  <input id="sentry-config" type="hidden" name="sentry-config" value='{{sentry_config}}'></input>
```
lnielsen, line 31: ``tojson``?

lnielsen:
I'm closing since we're not likely to integrate it anytime soon. @ntarocco mentioned that they tried on ILS but that it wasn't useful in the end.

PR #1090, opened 2021-09-07
installation: fix setup v0.58 issue

lnielsen:


lnielsen:
Just testing to see if the  CI will pass. If so, likely we have to add this to Invenio-Files-REST instead.

PR #1100, opened 2021-09-24
landing page: Fix subjects display

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/799

### Before

image

### After
Selection_639

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
@@ -282,11 +282,6 @@ export class RDMDepositForm extends Component {
                     }}
                   />
                   <SubjectsField
-                    initialOptions={_get(
-                      this.props.record,
-                      "ui.subjects",
-                      null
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -282,11 +282,6 @@ export class RDMDepositForm extends Component {
                     }}
                   />
                   <SubjectsField
-                    initialOptions={_get(
-                      this.props.record,
-                      "ui.subjects",
-                      null
-                    )}
```
jrcastro2, line 289: This is not really needed because (_i_) there are no subjects in the [UIObjectSchema](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/ui/schema.py#L166)
and (_ii_) we do not use the `initialOptions` [here](https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/components/SubjectsField.js#L77).

fenekku (approved):

PR #1120, opened 2021-11-16
A11y: ORCID and ROR icons on item pages

lhenze:
- alt tags for **img**
 -  aria-labels for **a**
 - title for **a**


fenekku (approved):
Thanks!

PR #1121, opened 2021-11-17
Pid fixes

zzacharo:
Depends on https://github.com/inveniosoftware/react-invenio-deposit/pull/396

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -8,7 +8,6 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """Routes for record-related pages provided by Invenio-App-RDM."""
-from copy import deepcopy
 from elasticsearch_dsl import Q
 from flask import current_app, render_template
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -71,10 +61,14 @@ def get_form_pids_config():
                                    "field blank to have one automatically "
                                    "assigned when publishing.")
             .format(scheme_label=scheme_label),
-            "unmanaged_help_text": _("Copy and paste here your {scheme_label}.")  # noqa
+            "unmanaged_help_text": _("Copy and paste here your {scheme_label} "
+                                     "or leave this field blank to have one "
```
ppanero, line 65: Quick question: didn't we agree that when:
- Chossing "external" i.e. the radio button is in "yes"
- And leaving the text blank

It would send:
```
"pids": {
    "doi": {
         "provider": "external"
         "identifier": ""    # or without this key anyway
     }
 }
 ```
 
This would not result in a new DOI, this would result in a validation error.

PR #1163, opened 2022-01-15
Typo fix in error messages: "ocurred" --> "occurred" in various files

Daniel-Mietchen:


fenekku (approved):
Thanks for spotting this!

PR #1164, opened 2022-01-15
Add/ fix German translation of an error message

Daniel-Mietchen:


mb-wali:
Hello! While we appreciate the contribution, and I have seen that the typo has been fixed by another PR - adding the translations workflow is via [Transifex](https://www.transifex.com/inveniosoftware/invenio/dashboard) rather than directly into the catalog files. Therefore I will have to close this PR.
If you would like to contribute to the german translation or any other available languages please have a look into the [Getting Started as a Translator](https://docs.transifex.com/getting-started-1/translators)

PR #1166, opened 2022-01-17
A11y - search results pages, landmarks and labels

lhenze:
- Adding "main" landmarks to records pages
- Sidebar on search results page: adding "aside" role, aria-label, and id
- Correct for/id label/input associations for facet checkboxes
- Associate facet-count with its checkbox

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/deposit.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/deposit.html
@@ -15,7 +15,6 @@
 {%- extends config.BASE_TEMPLATE %}
 
 {%- block page_body %}
-<main id="main">
   {%- if record %}
   <input id="deposits-record" type="hidden" name="deposits-record" value='{{record | tojson }}'></input>
   {%- endif %}
@@ -29,7 +28,6 @@
   <input id="deposits-record-permissions" type="hidden" name="deposits-record-permissions" value='{{permissions | tojson }}'></input>
   {%- endif %}
   <div id="deposit-form"></div>
-</main>
 {%- endblock page_body %}
 
 {%- block javascript %}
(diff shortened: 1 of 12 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/deposit.html
```diff
@@ -15,6 +15,7 @@
 {%- extends config.BASE_TEMPLATE %}
 
 {%- block page_body %}
+<main id="main">
```
fenekku, line 18: This is fine by me. I wonder though if we can place it above this line in here: https://github.com/inveniosoftware/invenio-theme/blob/master/invenio_theme/templates/semantic-ui/invenio_theme/page.html#L70, so that every Jinja page gets it right away.  You can make an issue and investigate it after.

fenekku (approved):
Thanks!

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/deposit.html
lhenze: This does seem to work and is a much better solution!
Confirming this is ok to do? I don't really have a sense of what other projects invenio_theme is used for.

lhenze: So like this: https://github.com/inveniosoftware/invenio-theme/compare/master...lhenze:a11y-page-landmarks?expand=1

fenekku: Just like that. Nice! You can submit that new one as a PR instead.

lhenze:
Ok, have done so, as well as removed the `main` landmarks added in previous PRs. 

Parent theme PR is here https://github.com/inveniosoftware/invenio-theme/pull/252

fenekku (approved):

PR #1178, opened 2022-01-24
dashboard: restructuring

zzacharo:


invenio_app_rdm/config.py
```diff
@@ -615,3 +616,16 @@ def files_rest_permission_factory(obj, action):
     "invenio_app_rdm/records/details/side_bar/export.html",
 ]
 """Template names for detail view sidebar components"""
+
+
+
+
+
+RDM_SEARCH_USER_DASHBOARD_COMMUNITIES = {
```
ntarocco, line 624: Less space above here? Maybe to be put close to the `RDM_SEARCH_DRAFTS` one?
Question: shall we change the var `RDM_SEARCH_DRAFTS` to be something like `RDM_SEARCH_USER_DASHBOARD_UPLOADS`? Warning: breaking change!

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -283,10 +283,11 @@ def new_record():
...
 #
 @login_required
-def deposit_search():
-    """List of user deposits page."""
+def dashboard(dashboard_name=None):
+    """Display user dashboard page."""
     return render_template(
-        "invenio_app_rdm/records/search_deposit.html",
+        "invenio_app_rdm/records/dashboard.html",
+        dashboard_name=dashboard_name or "uploads",
```
ntarocco, line 290: This `or "uploads"` was a temporary hack... I think that we could have a URL converter for the route `dashboard_specific` to accept only `[uploads|communities|requests]`.

invenio_app_rdm/config.py
```diff
@@ -539,7 +539,8 @@ def files_rest_permission_factory(obj, action):
     "record_file_download": "/records/<pid_value>/files/<path:filename>",
     "record_from_pid": "/<any({schemes}):pid_scheme>/<path:pid_value>",
     "record_latest": "/records/<pid_value>/latest",
-    "deposit_search": "/uploads",
+    "dashboard": "/dashboard",
+    "dashboard_specific": "/dashboard/<dashboard_name>",
```
ntarocco, line 543: Find better naming than what I came up with? :)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -0,0 +1,99 @@
...
+const rootElement = document.getElementById("invenio-user-dashboard");
+
+const TAB_PANES = [
+  // {
+  //   configDataAttribute: "invenio-search-user-uploads-config",
+  //   label: "Uploads",
+  //   pathname: "uploads",
+  // },
+  {
+    configDataAttribute: "invenio-search-user-communities-config",
```
ntarocco, line 29: Shall we merge with the `uploads` enabled instead?

ntarocco (commented):
Ok for the duplications between uploads, comm and requests React cmp.
I am wondering if we create something that will allow to create a multi-rsk-app and the only thing to duplicate is the overridden cmps, eg.:
```
MultiRSK tabs=[uploads, comm, requests] overriddenCmps=...
```
maybe not... :)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/communities.js
```diff
@@ -0,0 +1,257 @@
...
+} from "react-searchkit";
+import {
+  RDMBucketAggregationElement,
+  RDMRecordFacetsValues,
+  SearchHelpLinks,
+} from "../../search/components";
+
+function ResultsGridItemTemplate({ result, index }) {
+  return (
+    <Card fluid key={index} href={`/communities/${result.metadata.id}`}>
```
kprzerwa, line 44: is there any way to avoid hardcoding urls?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/communities.js
```diff
@@ -0,0 +1,257 @@
...
+            </span>
+          )}
+        </Item.Header>
+        {result.metadata.website && (
+          <Item.Description href={result.metadata.website} target="_blank">
+            <a
+              dangerouslySetInnerHTML={{
+                __html: result.metadata.website,
+              }}
+            ></a>
```
kprzerwa, line 80: hyperlink should be a part of a tag probably. I am not sure what will be the outcome on dangerouslySetInnerHTML set on `<a>` tag, is it needed? will it miss the url description?
```suggestion
          <Item.Description>
            <a href={result.metadata.website} target="_blank"
              dangerouslySetInnerHTML={{
                __html: result.metadata.website,
              }}
            />
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/communities.js
```diff
@@ -0,0 +1,257 @@
...
+                    <ResultsList />
+                  </Grid.Column>
+                </Grid.Row>
+              </Grid>
+            </Segment>
+          </Grid.Column>
+        </Grid.Row>
+        <Grid.Row verticalAlign="middle">
+          <Grid.Column width={4}></Grid.Column>
+          <Grid.Column width={8} textAlign="center">
```
kprzerwa, line 144: do we keep the responsive design ? if yes you might want to see into responsive props of grid columns
nit: 
```suggestion
          <Grid.Column width={8} textAlign="center" floated="right">
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/communities.js
```diff
@@ -0,0 +1,257 @@
...
+    </Grid>
+  </Container>
+);
+
+export const CommunitiesFacets = ({ aggs, currentResultsState }) => {
+  return (
+    <aside aria-label={i18next.t("filters")} id="search-filters">
+      {aggs.map((agg) => {
+        return (
+          <div className="ui accordion" key={agg.title}>
```
kprzerwa, line 238:
```suggestion
          <Accordion key={agg.title}>
```

invenio_app_rdm/records_ui/views/deposits.py
zzacharo: Ok, I haven't worked on it yet. I created https://github.com/inveniosoftware/invenio-app-rdm/issues/1184 so I or someone else can work on it independently!

invenio_app_rdm/config.py
zzacharo: It makes sense! I can change and note it down as a breaking change for our release notes! WDYT?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
```diff
@@ -0,0 +1,279 @@
...
+}
+
+export function RequestsResultsItemTemplate({ result, index }) {
+  return (
+    <Item key={index}>
+      <Item.Content>
+        <Item.Header href={`/requests/${result.id}`}>
+          {result.metadata.title}
+          {result.metadata.type && (
+            <span style={{ marginLeft: "5px" }}>
```
kprzerwa, line 148: inline style, maybe it is possible to use a class similar to `padding-r-5`

zzacharo:
@ntarocco For the multi RSK app component, I would wait to see if it adds some value and then extract it to a reusable component. Also, in such case I would probably use invenio-search-ui for such core feature, considering that the latter is used for search interfaces in Invenio/RDM.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
```diff
@@ -0,0 +1,279 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2020-2021 CERN.
```
kprzerwa, line 2:
```suggestion
// Copyright (C) 2020-2022 CERN.
```

invenio_app_rdm/config.py
zzacharo: Hmmm I checked a bit and `RDM_SEARCH_DRAFTS` is defined in [rdm-records](https://github.com/inveniosoftware/invenio-rdm-records/blob/f5f23882695b87eb92dae44a8b2bdac7b2ccac5c/invenio_rdm_records/config.py#L417) which doesn't know about dashboard. My thought now is if we should leave it like this and maybe just use it for the list of user uploads even if the naming is not ideal? Or create the `RDM_SEARCH_USER_DASHBOARD_UPLOADS ` and by default assign it to `RDM_SEARCH_DRAFTS`? For communities and requests, it might make sense to keep 2 configs e.g `RDM_SEARCH_USER_DASHBOARD_COMMUNITIES`, `RDM_SEARCH_COMMUNITIES` as we have 2 views that we search for communities and we might need to differentiate the configuration(?).

invenio_app_rdm/config.py
```diff
@@ -539,7 +539,8 @@ def files_rest_permission_factory(obj, action):
     "record_file_download": "/records/<pid_value>/files/<path:filename>",
     "record_from_pid": "/<any({schemes}):pid_scheme>/<path:pid_value>",
     "record_latest": "/records/<pid_value>/latest",
-    "deposit_search": "/uploads",
+    "dashboard": "/dashboard",
```
lnielsen, line 542: ``/me`` ?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/communities.js
zzacharo: Hmm at the moment, we don't have a way to inject configuration to the overriden components I think from somewhere central.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
zzacharo: Will rework on it in my next PR, regarding communities tab :)

invenio_app_rdm/records_ui/searchapp.py
```diff
@@ -210,6 +210,8 @@ def facets_config(config_name):
 
 def search_app_config(config_name, endpoint, overrides=None, **kwargs):
     """Search app config."""
+    import ipdb; ipdb.set_trace
```
kprzerwa, line 213: 👀

PR #1197, opened 2022-01-28
Search Facets: disclosure arrows and global focus-visible styles

lhenze:
- wrapping the `<i>` icon in a `<button>` element, because button natively gets focus and is recognized as such (you can tab to it, and screenreaders know it is interactive)
- surfacing semantic ui's @focusedFormBorderColor variable; setting global focus-visible styles (used mainly when an element is tabbed to

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
@@ -194,10 +194,13 @@ export const RDMRecordFacetsValues = ({
           </Label>
         </List.Content>
         {hasChildren ? (
-          <i
-            className={`angle ${isActive ? "down" : "right"} icon`}
+          <button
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
```diff
@@ -194,10 +194,13 @@ export const RDMRecordFacetsValues = ({
           </Label>
         </List.Content>
         {hasChildren ? (
-          <i
-            className={`angle ${isActive ? "down" : "right"} icon`}
+          <button
+            className={`iconhold`}
```
fenekku, line 198: minor: I think this should be enough
```suggestion
            className="iconhold"
```

fenekku (approved):
Looks good to me.

fenekku:
I can merge once it's marked as not a draft if you want.

PR #1199, opened 2022-01-28
dashboard: whitelist allowed routes

### LINKED ISSUE #1187: My dashboard: filter URL
Allow only `uploads`, `communities` and `requests` in the URL `/my/<dashboard_name>` using Flask routing `<any()>`. As default (`None` passed), `uploads` should be displayed.

Inject in the React app and make sure it is used to display that specific tab.

ntarocco:
* closes #1187

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -529,6 +529,8 @@ VOCABULARIES_DATASTREAM_WRITERS = {
 SEARCH_UI_SEARCH_TEMPLATE = 'invenio_app_rdm/records/search.html'
 """Search page's base template."""
 
+_DASHBOARD_ROUTES = ["uploads", "communities", "requests"]
+
 APP_RDM_ROUTES = {
     "index": "/",
     "help_search": "/help/search",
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/config.py
```diff
@@ -529,6 +529,8 @@ def files_rest_permission_factory(obj, action):
 SEARCH_UI_SEARCH_TEMPLATE = 'invenio_app_rdm/records/search.html'
 """Search page's base template."""
 
+_DASHBOARD_ROUTES = ["uploads", "communities", "requests"]
```
kprzerwa, line 532: just for me to understand the concept: do you want to keep the variable "local/protected" because it is too complex to be overridden?

kprzerwa (approved):

invenio_app_rdm/config.py
zzacharo: I guess that from Flask views point of view it is easy to add a new route but then adding a whole UI view it's more complex at the moment. That's why not exposing it as a configurable option makes probably sense.

zzacharo (approved):

PR #1205, opened 2022-02-02
oai: add datacite and oai_datacite config

mitsosf:
Requires https://github.com/inveniosoftware/invenio-rdm-records/pull/888
Replaces https://github.com/inveniosoftware/invenio-oaiserver/pull/213

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -430,6 +430,17 @@ OAISERVER_METADATA_FORMATS = {
         'schema': 'http://www.openarchives.org/OAI/2.0/oai_dc.xsd',
         'namespace': 'http://www.openarchives.org/OAI/2.0/oai_dc/',
     },
+    'datacite': {
+        'serializer': 'invenio_rdm_records.oai:datacite_etree',
+        'schema': 'http://schema.datacite.org/'
+                  'meta/nonexistant/nonexistant.xsd',
+        'namespace': 'http://datacite.org/schema/nonexistant',
+    },
+    'oai_datacite': {
+        'serializer': ('invenio_rdm_records.oai:oai_datacite_etree'),
+        'schema': 'http://schema.datacite.org/oai/oai-1.1/oai.xsd',
+        'namespace': 'http://schema.datacite.org/oai/oai-1.1/',
+    },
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/config.py
```diff
@@ -430,6 +430,17 @@ def files_rest_permission_factory(obj, action):
         'schema': 'http://www.openarchives.org/OAI/2.0/oai_dc.xsd',
         'namespace': 'http://www.openarchives.org/OAI/2.0/oai_dc/',
     },
+    'datacite': {
+        'serializer': 'invenio_rdm_records.oai:datacite_etree',
+        'schema': 'http://schema.datacite.org/'
+                  'meta/nonexistant/nonexistant.xsd',
```
ppanero, line 436: Not sure what is better, breaking the line like this or adding a `# noqa` so it passe PEP 8... In the end is for aesthetics but is harder to read the line when separated. WDYT @mitsosf ? The alternative would be:

```suggestion
        'schema': 'http://schema.datacite.org/meta/nonexistant/nonexistant.xsd',  # noqa
```

invenio_app_rdm/config.py
```diff
@@ -430,6 +430,17 @@ def files_rest_permission_factory(obj, action):
...
+        'namespace': 'http://datacite.org/schema/nonexistant',
+    },
+    'oai_datacite': {
+        'serializer': ('invenio_rdm_records.oai:oai_datacite_etree'),
```
ppanero, line 440: parenthesis not needed? just trying to avoid triggering the logic of reading a tuple when is not needed 😇

ppanero (approved):
LGTM, just two minor comments.

invenio_app_rdm/config.py
```diff
@@ -430,6 +430,17 @@ def files_rest_permission_factory(obj, action):
...
+        'schema': 'http://schema.datacite.org/oai/oai-1.1/oai.xsd',
+        'namespace': 'http://schema.datacite.org/oai/oai-1.1/',
+    },
```
ppanero, line 443: Might be worth documenting the variable like in the rest of the configs in this file.

Issue #1217, opened 2022-02-07
Fix bug when selecting a facet + open/closed, to reproduce select `open` facet and then select closed option


jrcastro2:
If a facet is selected and we change click the Open/Closed button we have some chances to get an empty query and not see the facet again to disable it.

## Steps to reproduce:
1. Go to "My dashboard" and select the "Requests" tab.
2. Click the "open" option in the status facet.
3. Click the "Closed" toggle button next to the search.
4. Empty result page is shown and you cannot deselect the open option as it's not appearing.

PR #1212, opened 2022-02-04
[Depends on react-searchkit/pull/214 and invenio-requests/pull/127]Add request tab

jrcastro2:
* closes #1181
* Needs https://github.com/inveniosoftware/invenio-requests/pull/127
* Needs https://github.com/inveniosoftware/react-searchkit/pull/214
* Needs https://github.com/inveniosoftware/invenio-theme/pull/254

image

PR #1221, opened 2022-02-08
[NEEDS https://github.com/inveniosoftware/invenio-app-rdm/pull/1212]Fix filters bug

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1217
* Needs https://github.com/inveniosoftware/invenio-app-rdm/pull/1212 

**Note for reviewers:** Only check second commit https://github.com/inveniosoftware/invenio-app-rdm/pull/1221/commits/31b8240cc7de6fe8b14a9e669db6287236c4fead

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
@@ -253,10 +253,8 @@ export const RequestToggleComponent = ({
       return;
     }
     setOpen(true);
-    const filterToRemove = userSelectionFilters.filter((obj1) => {
-      return obj1.includes("is_open");
-    });
-    const filters = filterToRemove ? filterToRemove : [];
+    // We add the selected filters by the user to remove them
+    const filters = [...userSelectionFilters];
     filters.push(["is_open", "true"]);
     updateQueryFilters(filters);
   };
@@ -266,10 +264,8 @@ export const RequestToggleComponent = ({
       return;
     }
     setOpen(false);
-    const filterToRemove = userSelectionFilters.filter((obj1) => {
-      return obj1.includes("is_open");
-    });
-    const filters = filterToRemove ? filterToRemove : [];
+    // We add the selected filters by the user to remove them
+    const filters = [...userSelectionFilters];
     filters.push(["is_open", "false"]);
     updateQueryFilters(filters);
   };


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
```diff
@@ -253,10 +253,8 @@ export const RequestToggleComponent = ({
       return;
     }
     setOpen(true);
-    const filterToRemove = userSelectionFilters.filter((obj1) => {
-      return obj1.includes("is_open");
-    });
-    const filters = filterToRemove ? filterToRemove : [];
+    // We add the selected filters by the user to remove them
+    const filters = [...userSelectionFilters];
```
jrcastro2, line 257: Is it ok to remove all the filters or should we only remove the `Status` filters?
To me it makes sense to remove all of them

NRodriguezcuellar (approved):

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
jrcastro2: To keep them all is not an option at the moment since it's the reason of the bug. The issue happens when we click a status such as "open" and then we click "Closed" filter. See below an example.
filters_bug

Based on this we can or remove all filters as it's happening in this PR or just remove the satus filters and keep the type filters.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
```diff
@@ -266,10 +264,8 @@ export const RequestToggleComponent = ({
       return;
     }
     setOpen(false);
-    const filterToRemove = userSelectionFilters.filter((obj1) => {
-      return obj1.includes("is_open");
-    });
-    const filters = filterToRemove ? filterToRemove : [];
+    // We add the selected filters by the user to remove them
+    const filters = [...userSelectionFilters];
```
zzacharo, line 268: Shouldn't that be a `let` as you mutate the variable just below?

jrcastro2: Hmm I think it can be a `const` since it does not stop array and objects from being modified it only stops the variable itself from being reassigned or being overwritten if I am not mistaken

ntarocco: `const` is ok, if you add/remove items to an array will not change the obj. Assigning the const to a new array will fail indeed.

ntarocco (approved):

Issue #1229, opened 2022-02-09
Empty result page to request search


jrcastro2:
Add the empty result pages as it is in the mockup

PR #1241, opened 2022-02-10
[DEPENDS on https://github.com/inveniosoftware/react-searchkit/pull/217] requests: add empty pages

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1229
* needs https://github.com/inveniosoftware/react-searchkit/pull/217

image

image

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
@@ -24,6 +24,7 @@ import {
   SearchBar,
   Sort,
   Toggle,
+  withState,
 } from "react-searchkit";
 import {
   Button,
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
```diff
@@ -323,6 +324,55 @@ export const RequestsFacets = ({ aggs, currentResultsState }) => {
...
+  };
+  return (
+    <>
+      <Segment placeholder textAlign="center">
+        {is_open && !queryString ? (
+          <>
+            <Header as="h2" icon>
+              <span>{i18next.t("All done!")}</span>
+            </Header>
+            <span className="grey-color">
```
kpsherva, line 348: this should be solved with subheader component of semantic ui, instead of span and custom CSS class. Custom CSS should be the last resort

jrcastro2: Thanks! Was not aware of the subheader cmp, will try to explore more semantic ui react next times before going for custom css

Issue #1238, opened 2022-02-09
Bug on reload page with Open/Closed buttons


jrcastro2:
## Describe the bug
When reloading the page the previous filter is still selected and the button value might be wrong.

PR #1242, opened 2022-02-10
requests: fix Open/Closed bug on page reload

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1238

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
@@ -14,7 +14,8 @@ import {
 import { i18next } from "@translations/invenio_app_rdm/i18next";
 import _get from "lodash/get";
 import _truncate from "lodash/truncate";
-import React, { useState } from "react";
+import React, { Component } from "react";
+import PropTypes from "prop-types";
 import {
   BucketAggregation,
   Count,
@@ -23,7 +24,6 @@ import {
   ResultsPerPage,
   SearchBar,
   Sort,
-  Toggle,
   withState,
 } from "react-searchkit";
 import {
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/search.less
```diff
@@ -44,3 +44,7 @@
 .ui.link.items > .item:hover {
   cursor: auto !important;
 }
+
+.ui.active.button.request-search-button {
+  cursor: default;
```
kpsherva, line 49: Why is this button special in the context of the whole theme? I know that you would like to make the user think the button is no longer "clickable" but I think we should think about the big picture here:
if you change it only for this button, it will be inconsistent with the rest of the semantic ui design. I would just leave the cursor as it is defined in semantic, or, if you insist, you should overwrite the whole `ui.active.button` class, but then it will change across the theme.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
```diff
@@ -246,7 +246,14 @@ export const RequestToggleComponent = ({
   updateQueryFilters,
   userSelectionFilters,
 }) => {
-  const [open, setOpen] = useState(true);
+  const [open, setOpen] = useState(undefined);
+
+  useEffect(() => {
```
ntarocco, line 251: I don't understand why this is needed. Can you please explain?

You know why I don't like effect? It is because it is hiding what it is really happening. This function will run on each `render`: do we want that? ping @kpsherva @zzacharo FIY

jrcastro2: Ah I see, well I am using functional cmp mainly because I see that all the overriden cmps are declared this way, if class components can be used I will change it as I agree it's much more readable, maybe @zzacharo knows if possible, otherwise I will give it a try

jrcastro2: For the record: After discussion decided to go for a class component that uses the withState, instad of a functional cmp overriding the Toggle cmp in RSK

zzacharo (approved):

Issue #1247, opened 2022-02-14
Form header


jrcastro2:
![image](https://user-images.githubusercontent.com/15194802/153830479-ed7b2943-4eda-40b5-b767-93196da5e4bf.png)

PR #1260, opened 2022-02-15
 Add community header

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1247
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1272
* needs https://github.com/inveniosoftware/react-invenio-deposit/pull/418

## Community selected
image

## No community selected
image

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -641,6 +641,7 @@ RDM_CITATION_STYLES_DEFAULT = 'apa'
 """Default citation style"""
 
 APP_RDM_DETAIL_SIDE_BAR_TEMPLATES = [
+    "invenio_app_rdm/records/details/side_bar/manage_menu.html",
     "invenio_app_rdm/records/details/side_bar/metrics.html",
     "invenio_app_rdm/records/details/side_bar/versions.html",
     "invenio_app_rdm/records/details/side_bar/keywords_subjects.html",
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -46,12 +46,28 @@
       </div>
     {% endif %}
 
-    {% if permissions is defined and permissions.can_edit and not is_preview %}
-    <div class="ui warning flashed top-attached manage message">
-      <div class="ui container">
-        <div id="recordManagement" data-record='{{ record | tojson }}' data-permissions='{{ permissions | tojson }}'></div>
+    {% if community %}
+    <!-- TODO: Checks below must be deleted once community is resolved -->
+    {% if community.links is defined %}
+      {%- set logoLink = community.links.logo %}
+      {%- set communityTitle = community.metadata.title %}
+    {% else %}
+      {%- set logoLink = "/static/images/square-placeholder.png" %}
+      {%- set communityTitle = community %}
+    {% endif %}
+      <div class="ui fluid container record-community-header">
+        <div class="ui fluid container">
+          <div class="ui container community-header">
+            <img
+              class="community-logo-header"
+              src="{{logoLink}}"
+              onerror="this.onerror=null;this.src='/static/images/square-placeholder.png';" />
+            <div class="community-header-info">
+              {{communityTitle}}
+            </div>
+          </div>
+        </div>
       </div>
-    </div>
     {% endif %}
 
     {% if record.is_published and not record.versions.is_latest %}
@@ -75,7 +91,7 @@
 
 
   <div class="ui container">
-    <div class="ui relaxed grid">
+    <div class="ui relaxed grid mt-5">
       <div class="two column row top-padded">
 
         {# Main content #}
--- a/invenio_app_rdm/records_ui/views/utils.py
+++ b/invenio_app_rdm/records_ui/views/utils.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio App RDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Utility functions."""
+
+from invenio_communities.communities.resolver import CommunityResolver
+from invenio_rdm_records.requests import CommunitySubmission
+
+
+def get_community_uuid(record):
+    """Retrieves community UUID.
+
+    WARNING: This should be replaced by the resolvers.
+    """
+    community_uuid = None
+    parent = record['parent']
+    if parent.get('communities'):
+        community_uuid = parent['communities']['default']
+    elif parent.get('review'):
+        review = parent['review']
+        if review['type'] == CommunitySubmission.type_id:
+            community_uuid = review['receiver'][CommunityResolver.type_id]
+    return community_uuid
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
@@ -32,6 +32,7 @@ import {
   SaveButton,
   TitlesField,
   VersionField,
+  CommunityHeader,
 } from "react-invenio-deposit";
 import { AccordionField } from "react-invenio-forms";
 import { Card, Container, Divider, Grid, Ref, Sticky } from "semantic-ui-react";
@@ -135,14 +136,19 @@ export class RDMDepositForm extends Component {
   };
 
   render() {
+    const { record, files, permissions, community } = this.props;
     return (
       <DepositFormApp
         config={this.config}
-        record={this.props.record}
-        community={this.props.community}
-        files={this.props.files}
-        permissions={this.props.permissions}
+        record={record}
+        community={community}
+        files={files}
+        permissions={permissions}
       >
+        <CommunityHeader
+          community={community}
+          imagePlaceholderLink="/static/images/square-placeholder.png"
+        />
         <FormFeedback fieldPath="message" />
         <Container style={{ marginTop: "10px" }}>
           <DepositFormTitle />
@@ -155,7 +161,7 @@ export class RDMDepositForm extends Component {
                   label={i18next.t("Files")}
                   ui={this.accordionStyle}
                 >
-                  {this.noFiles && this.props.record.is_published && (
+                  {this.noFiles && record.is_published && (
                     <p
                       style={{
                         textAlign: "center",
@@ -167,7 +173,7 @@ export class RDMDepositForm extends Component {
                     </p>
                   )}
                   <FileUploader
-                    isDraftRecord={!this.props.record.is_published}
+                    isDraftRecord={!record.is_published}
                     quota={{
                       maxFiles: 100,
                       maxStorage: 10 ** 10,
@@ -191,7 +197,7 @@ export class RDMDepositForm extends Component {
                         fieldPath={`pids.${pid.scheme}`}
                         fieldLabel={pid.field_label}
                         isEditingPublishedRecord={
-                          this.props.record.is_published === true // is_published is `null` at first upload
+                          record.is_published === true // is_published is `null` at first upload
                         }
                         managedHelpText={pid.managed_help_text}
                         pidLabel={pid.pid_label}
@@ -210,7 +216,7 @@ export class RDMDepositForm extends Component {
                   />
                   <TitlesField
                     options={this.vocabularies.metadata.titles}
-                    recordUI={this.props.record.ui}
+                    recordUI={record.ui}
                     required
                   />
                   <PublicationDateField required />
@@ -224,7 +230,7 @@ export class RDMDepositForm extends Component {
                   />
                   <DescriptionsField
                     options={this.vocabularies.metadata.descriptions}
-                    recordUI={_get(this.props.record, "ui", null)}
+                    recordUI={_get(record, "ui", null)}
                     editorConfig={{
                       removePlugins: [
                         "Image",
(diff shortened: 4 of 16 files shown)


jrcastro2:
Some discussion points:
- If a community has a no logo the link is still coming from the backend but if we try to access it, we get a 404, I would propose to remove it if no logo.
- If the community was already saved in the record, will it be resolved in the backend? (I am assuming yes)

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -129,3 +128,29 @@
 .item div.description:not(.license-item-description) {
   display: initial !important;
 }
+
+.deposits-community-header {
```
kpsherva, line 132: to which element/class are those rules applying to ? I cannot see them in the PR. If they are applying to react-invenio-deposit, maybe they should be defined there?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -129,3 +128,29 @@
...
+  margin: -20px 0px 20px 0px;
+  background-color: #ededed;
```
kpsherva, line 134: the color is hardcoded, it should be defined as a variable to allow overridability

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -129,3 +128,29 @@
...
+  height: 65px;
```
kpsherva, line 135: usage of fixed units is not the best practice, especially in the theme which relies on relative units

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -129,3 +128,29 @@
...
+
+.community-header {
+  padding-top: 10px;
+  &.empty {
+    padding-top: 23px;
+  }
+}
+
+.ui.container.community-header {
+  .ui.mini.image {
```
kpsherva, line 146: what is the problem with the default semantic ui image class?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -129,3 +128,29 @@
...
+    width: 4em;
+    height: 4em;
+    display: inline-block;
+    margin-right: 0.25em;
+  }
+}
+
+.error-message {
```
kpsherva, line 154: see the comment in the other PR

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -129,3 +128,29 @@
...
+  color: red;
```
kpsherva, line 155: hardcoded color, try to use variables defined in the theme

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -129,3 +128,29 @@
 .item div.description:not(.license-item-description) {
   display: initial !important;
 }
+
+.deposits-community-header {
+  margin: -20px 0px 20px 0px;
```
kpsherva, line 133: I am not a big fan of negative margins, because from my experience, it leads to problems with the rest of the layouts. Is there any other way to solve this ? Did you check `.top.attached` class of semantic ui containers?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
```diff
@@ -129,3 +128,29 @@
...
+  background-color: #ededed;
+  height: 65px;
+}
+
+.community-header {
+  padding-top: 10px;
```
kpsherva, line 139: what about using `.relaxed` class of container ?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -142,6 +143,10 @@ export class RDMDepositForm extends Component {
         files={this.props.files}
         permissions={this.props.permissions}
       >
+        <CommunityHeader
+          record={this.props.record}
```
kpsherva, line 147: why not using prop deconstruction ? prettier didn't complain about this?
`const {record} = this.props`;
I would also refactor the other variables too

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
jrcastro2: I agree with you but there is no css rules in react-inveni-deposit, other rules applying to cmp declared here are created in app-rdm. I guess I can try to crate them here

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
jrcastro2: The issue is that it's not a square.
### Default image
image
### Our custom css
image

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
jrcastro2: Tried to use it but seems not to work

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
jrcastro2: Using attached would attach it to the header wich is the one containing the bottom-margin and forcing all the rest to use negative margin. Instead I removed the margin from the header and also all the negative margin from other repos

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
kpsherva: then I would apply the generated padding classes from invenio theme `pt-10`

invenio_app_rdm/config.py
```diff
@@ -641,6 +641,7 @@ def files_rest_permission_factory(obj, action):
 """Default citation style"""
 
 APP_RDM_DETAIL_SIDE_BAR_TEMPLATES = [
+    "invenio_app_rdm/records/details/side_bar/manageMenu.html",
```
ntarocco, line 644: In Python we normally use the "underscore" case :)
```suggestion
    "invenio_app_rdm/records/details/side_bar/manage_menu.html",
```

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
jrcastro2: Hmm I think in this particular case it's more handy have the rule defined in this class, mainly because `community-header` class is used widely and if we want to change it we can easily change it only in this class, otherwise we would need to go to all the tags and change the `pt-XX` and we would probably forget some.

invenio_app_rdm/records_ui/views/utils.py
```diff
@@ -0,0 +1,24 @@
...
+# Invenio App RDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Utility functions."""
+
+from invenio_communities.communities.resolver import CommunityResolver
+from invenio_rdm_records.requests import CommunitySubmission
+
+
+def get_community_uuid(record):
```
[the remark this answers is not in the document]
jrcastro2: Sure, here the issue https://github.com/inveniosoftware/invenio-app-rdm/issues/1286

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
kpsherva: I think the component is ready so I would pass a prop with className to the image component created by Zach and then define this special class here, instead of overwriting the semantic ui class

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -135,14 +136,19 @@ export class RDMDepositForm extends Component {
...
-        record={this.props.record}
-        community={this.props.community}
-        files={this.props.files}
-        permissions={this.props.permissions}
+        record={record}
+        community={community}
+        files={files}
+        permissions={permissions}
       >
+        <CommunityHeader
```
Pineirin, line 148: (For myself) I should put the communities_enabled if here.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCommunity.js
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <Container fluid>
+      <Container className="community-header">
+        <Image
+          className="ui mini image"
+          size="mini"
+          src={logoLink}
+          fallbackSrc={fallbackSrc}
+        />
+        {community?.metadata?.title ? community.metadata.title : community}
```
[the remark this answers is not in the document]
jrcastro2: Temporary yes, right now we only get the UUID of the community, once it's resolved this won't be needed.

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -46,12 +46,10 @@
       </div>
     {% endif %}
 
-    {% if permissions is defined and permissions.can_edit and not is_preview %}
-    <div class="ui warning flashed top-attached manage message">
-      <div class="ui container">
-        <div id="recordManagement" data-record='{{ record | tojson }}' data-permissions='{{ permissions | tojson }}'></div>
+    {% if record.parent.communities.default %}
```
ntarocco, line 49: are you sure about this? if there is a request, the community won't appear here no?
Shouldn't this be:
```suggestion
    {% if community %}
```
?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCommunity.js
```diff
@@ -0,0 +1,31 @@
...
+// Copyright (C) 2022 CERN.
+//
+// Invenio RDM Records is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import React from "react";
+import { Container } from "semantic-ui-react";
+import { Image } from "react-invenio-forms";
+
+export const RecordCommunity = (props) => {
```
ntarocco, line 11: Why having a React component here and not going for Jinja?
Is it to re-use the `Image` cmp?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
kpsherva: what is the result of this discussion ? where do we put CSS in the end?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordManagement.js
```diff
@@ -27,28 +26,29 @@ export const RecordManagement = (props) => {
...
-        <Grid.Row>
-          <Icon name="cogs" />
-          <h2 id="record-management-header">{i18next.t('Manage')}</h2>
-        </Grid.Row>
         <Grid.Row className="record-management-row">
           <EditButton recid={recid} onError={handleError} />
+        </Grid.Row>
+        <Grid.Row className="record-management-row">
           <NewVersionButton
+            style={{ display: "block" }}
```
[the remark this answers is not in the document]
jrcastro2: I agree, I mainly used it because it's how the [`NewVersionButton`](https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/components/NewVersionButton.js#L41) was implemented.
To fix this we will need to refactor it, I will create an issue as this tasks scope already grew a lot.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/deposit.less
jrcastro2: As for my understanding we create them in `invenio-app-rdm` and once we refactor the css we will take a decision on this.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCommunity.js
jrcastro2: Discussed IRL: agreed to go for Jinja

ntarocco (approved):

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/EditButton.js
```diff
@@ -28,7 +28,13 @@ export const EditButton = (props) => {
   };
 
   return (
-    <Button color="orange" size="mini" onClick={handleClick} loading={loading}>
+    <Button
+      fluid
+      color="orange"
```
kpsherva, line 33: do we have any primary/secondary color set for edit in the theme instead of hardcoding ?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordManagement.js
```diff
@@ -27,28 +26,29 @@ export const RecordManagement = (props) => {
...
-              divClassName="share-button"
               disabled={!permissions.can_update_draft}
               recid={recid}
             />
           )}
         </Grid.Row>
         {error && (
           <Grid.Row className="record-management-row">
-            <p>{error}</p>
+            <Message negative>{error}</Message>
```
kpsherva, line 51: please add to your issue to replace this also with a generic error component, we keep using `<message negative` in several different ways, so it would be nice to harmonize it

PR #1269, opened 2022-02-15
Make ORCID icons larger and swap order

### LINKED ISSUE #1161: ORCID icon display -- not aligned with ORCID brand guidelines
Package version: <= 7 

## Describe the bug
Currently, the ORCID icon precedes the creatibutor's name.  It is displayed at 14px * 14px.
## Steps to Reproduce
1. Go to search results such as  `https://127.0.0.1:5000/search?q=&l=list&p=1&s=10&sort=newest` or a record detail page such as `https://127.0.0.1:5000/records/n24pt-mt351`

## Expected behavior

Guidelines are here: https://info.orcid.org/brand-guidelines/

Icon should follow the name and should be displayed at 16px*16px. 

## Additional context

Are there any objections to making this change?  Everywhere the icon appears, it would be after the name.  (So, before the affiliation superscript numbers.)  And slightly larger.

lhenze:
* moves Icon after name for both landing page and search results
* standardizes the color of the linked names
* standardizes html classes for landing page and search results so they can be styled together
* closes #1161

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
@@ -2,7 +2,7 @@
     Copyright (C) 2020 CERN.
     Copyright (C) 2020 Northwestern University.
     Copyright (C) 2021 Graz University of Technology.
-    Copyright (C) 2021 New York University.
+    Copyright (C) 2021-2022 New York University.
 
     Invenio RDM Records is free software; you can redistribute it and/or modify
     it under the terms of the MIT License; see LICENSE file for more details.
@@ -35,8 +35,7 @@
 {% macro show_creatibutors(creatibutors, show_affiliations=False) %}
   {% for creatibutor in creatibutors %}
   <dd class="creatibutor-wrap">
-    {{ creatibutor_icon(creatibutor) }}
-    <a class="ui tooltip-popup text-muted creatibutor-link"
+    <a class="ui tooltip-popup creatibutor-link"
       {% if show_affiliations and creatibutor.affiliations %}
         data-content="{{ creatibutor.affiliations|join('; ', attribute='1') }}"
       {% endif %}
@@ -44,15 +43,20 @@
     >
 
       <span class="creatibutor-name">
-        {{ creatibutor.person_or_org.name }}
+        {{- creatibutor.person_or_org.name -}}
       </span>
-
-      <sup class="font-tiny text-muted">
-        {% for affiliation in creatibutor.affiliations %}
+      {%- if creatibutor.affiliations -%}
+      <sup class="font-tiny">
+        {%- for affiliation in creatibutor.affiliations -%}
           {{ affiliation[0] }}{{ ", " if not loop.last }}
-        {% endfor %}
+        {%- endfor -%}
       </sup>
+      {%- endif -%}
     </a>
+    {{- creatibutor_icon(creatibutor) -}}
+    {%- if not loop.last -%}
+       <span class="separator">;</span>
+    {%- endif -%}
   </dd>
   {% endfor %}
 {%- endmacro %}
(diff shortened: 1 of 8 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
```diff
@@ -35,24 +35,28 @@
...
       </span>
-
-      <sup class="font-tiny text-muted">
-        {% for affiliation in creatibutor.affiliations %}
-          {{ affiliation[0] }}{{ ", " if not loop.last }}
-        {% endfor %}
+      {%- if creatibutor.affiliations -%}
+      <sup class="font-tiny">
+        {%- for affiliation in creatibutor.affiliations -%}
+          {{- affiliation[0] -}}{{- ", " if not loop.last -}}
```
fenekku, line 51: For Jinja interpolation `{{-` should typically not be needed. It can actually be confusing: value negated or space ignored? For `{%-` that's fine and typically nice.

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
```diff
@@ -35,24 +35,28 @@
...
+        {%- endfor -%}
       </sup>
+      {%- endif -%}
     </a>
+    {{- creatibutor_icon(creatibutor) -}}
+    {%- if not loop.last -%}
+       <span class="separator">;</span>
```
fenekku, line 58: I guess the `;` is making a comeback @jennur ;) . Jenny and I were initially trying to tone down the punctuation and make use of whitespace + the contributor icon to separate each contributor. I think I still lean that way, but we can discuss.

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
```diff
@@ -66,13 +70,13 @@
...
+<section class="ui sixteen wide column content affiliations-list" id="{{ group }}-affiliations" aria-label="{{ _('Affiliations for') }} {{ group }}">
   <dl>
   {% for affiliation in affiliations %}
-      <dd class="text-muted">
-          {{ affiliation[0] }}.
+      <dd>
+          {{- affiliation[0] }}.
 
-          {%- if affiliation[2] %}
+          {% if affiliation[2] %}
```
fenekku, line 79: minor: would flip these `-` if possible as per above (won't repeat comment for other instances).

fenekku (approved):
The end result looks overall good. Punctuation-wise it seems busy for me though. 

(the screenshot has some strange contributors which I assume was just some wip artefact)

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
lhenze: Oh, I see, I hadn't realized. One factor is that they won't all have ORCID or group icons. Yes let's discuss, and also discuss the best way to make sure it's straightforward to theme.

jennur: I too suggest to rather add some extra space to separate the creatibutors. Even if we are to keep the `;` I think some extra white space is needed between the creatibutors.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/landing_page/creatibutors.less
```diff
@@ -2,12 +2,61 @@
...
+
+  .creatibutor-wrap {
+    color: @mutedTextColor;
+    display: flex;
+    flex-wrap: wrap;
+    &:nth-child(n):not(:last-child) {
+      margin-right: 0.5rem;
+    }
+    .identifier-link, .group.icon {
+      margin: 0 1px 0 5px;
```
jennur, line 36: We should use rem/em values to ensure the spacing stays consistent when scaling the screen.

jennur (commented):
Looks good to me! Just left a few comments about some minor details. 

Thanks for fixing the color contrast issue with muted text!

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
lhenze: Sounds good; it's now bumped from .5rem to 1rem

Issue #1284, opened 2022-02-18
Replace placeholder with square-placeholder


jrcastro2:
Delete placeholder from invenio-communities and replace it with square-placeholder from invenio-theme

PR #1285, opened 2022-02-18
placeholder: replace with square-placeholder

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1284


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/communities.js
```diff
@@ -55,7 +55,7 @@ export function CommunitiesResultsItemTemplate({ result, index }) {
     <Item key={index}>
       <Image
```
kpsherva, line 56: another place for possible reusable component :)

Pineirin (approved):
LGTM! Again, we can use this as a temporal solution.

Issue #1318, opened 2022-03-01
Landing page: "Get a link" button doesn't copy


lnielsen:
Package version (if known): <version>

## Steps to Reproduce

1. On a record you have published, click "Share"
2. Click "Get a link"
3. Notice that the link was not copied into the clipboard

This happens for me on Safari.

PR #1323, opened 2022-03-01
landing page: fix bug where get-link button doesnt copy

jennur:
Related to #1318 

- Copy on get-link now works for Chrome, but Safari still has the same issues likely because of async issues due to the creation of the link...


jennur:
Created [#1324](https://github.com/inveniosoftware/invenio-app-rdm/issues/1324) to fix the issue in Safari

PR #1356, opened 2022-03-07
semantic-ui: add packaged theming

kpsherva:


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/frontpage.less
```diff
@@ -72,26 +73,189 @@ only screen and (max-width: 993px) {
...
+  font-size: 1.2em;
+  font-weight: 300;
+}
+
+.section-content-light-bg {
+  background-color: rgba(173, 206, 224, 0.52);
+}
+
+.section-content-white-bg {
+  background-color: #fff;
```
zzacharo, line 148: I guess you can use `@white` here, no?

kpsherva: sure but I prefer to leave it aside and refactor colors separately, otherwise it will be super difficult to keep track of all the changes. I plan to introduce it gradually.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/frontpage.less
```diff
@@ -72,26 +73,189 @@ only screen and (max-width: 993px) {
...
 .goto-button a:active {
-    color: white !important;
-    background-color: #01599c;
+    color: white;
+}
+
+.section-content-dark-bg,
+.section-content-white-bg,
+.section-content-light-bg {
+  padding-top: calc(30px);
```
jennur, line 91: We should use em/rem values instead of px to make sure the spacing is scaling well with the browser's font-size setting. (For this to work as expected everywhere, we also need to replace the `font-size: 14px` from the  html- and body-element with a rem value.)

kpsherva: yes I plan to refactor this too, but in the next iteration, for now I am removing redundant files

PR #1382, opened 2022-03-15
theme: add image custom rules

jrcastro2:


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/image.overrides
```diff
@@ -0,0 +1,3 @@
+/***********************************************
```
kpsherva, line 1: thanks for adding these!

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides
```diff
@@ -35,3 +35,6 @@
   cursor: auto !important;
 }
 
+.ui.relaxed.divided.items > .item.community-member-item {
+  padding: 1em 0em;
+}
```
kpsherva, line 40:
```suggestion
.ui.relaxed.divided.items {
  .item.community-member-item {
     padding: 1em 0em;
     }
}
```
this notation is more readable and flexible, no need to use `>` in less

zzacharo (approved):

PR #1384, opened 2022-03-15
searchapp: move reusable functions

jrcastro2:
* move reusable functions to
  invenio-search-ui
* requires https://github.com/inveniosoftware/invenio-search-ui/pull/131
Co-authored-by: Pablo pineirin@gmail.com

### DIFF
--- a/invenio_app_rdm/records_ui/searchapp.py
+++ b/invenio_app_rdm/records_ui/searchapp.py
@@ -7,250 +7,46 @@
 
 """Configuration helper for React-SearchKit."""
 
-from copy import deepcopy
 from functools import partial
 
 from flask import current_app
-from invenio_rdm_records.searchconfig import OptionsSelector, \
-    SortOptionsSelector
-
-
-class SortConfig(SortOptionsSelector):
-    """Sort options for the search configuration."""
-
-    def map_option(self, key, option):
-        """Generate a RSK search option."""
-        return {
-            "sortBy": key,
-            "text": option['title']
-        }
-
-
-class FacetsConfig(OptionsSelector):
-    """Facets options for the search configuration."""
-
-    def map_option(self, key, option):
-        """Generate an RSK aggregation option."""
-        title = option.get('title', option['facet']._label)
-
-        ui = deepcopy(option['ui'])
-        ui.update({
-            'aggName': key,
-            'title': title,
-        })
-
-        # Nested facets
-        if 'childAgg' in ui:
-            ui['childAgg'].setdefault('aggName', 'inner')
-            ui['childAgg'].setdefault('title', title)
-
-        return ui
-
-
-class SearchAppConfig:
-    """Configuration generator for React-SearchKit."""
-
-    default_options = dict(
-        endpoint=None,
-        hidden_params=None,
-        app_id='rdm-search',
-        headers=None,
-        list_view=True,
-        grid_view=False,
-        pagination_options=(10, 20, 50),
-        default_size=10,
-        default_page=1,
-        facets=None,
-        sort=None,
-        initial_filters=[],
-    )
-
-    def __init__(self, configuration_options):
-        """Initialize the search configuration.
-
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/records_ui/searchapp.py
```diff
@@ -7,250 +7,46 @@
...
     return {
-        'search_app_config': partial(
+        'search_app_rdm_config': partial(
             search_app_config,
             'RDM_SEARCH',
             current_app.config['RDM_FACETS'],
-            '/api/records'),
-        'search_app_user_uploads_config': partial(
+            current_app.config['RDM_SORT_OPTIONS'],
+            '/api/records',
```
kpsherva, line 24: I am not really a fan of hardcoded URLs, can we get it from any config?

invenio_app_rdm/records_ui/searchapp.py
```diff
@@ -233,24 +20,33 @@ def search_app_context():
             search_app_config,
             'RDM_SEARCH',
             current_app.config['RDM_FACETS'],
-            '/api/records'),
+            current_app.config['RDM_SORT_OPTIONS'],
+            '/api/records',
+            {"Accept": "application/vnd.inveniordm.v1+json"}
```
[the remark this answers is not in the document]
kpsherva: sorry, silly question. Why in some searches we have
```
{"Accept": "application/vnd.inveniordm.v1+json"}
```
and in the others

```
{"Accept": "application/json"}
```

?

invenio_app_rdm/records_ui/searchapp.py
```diff
@@ -7,250 +7,46 @@
...
-        'search_app_user_requests_config': partial(
+            current_app.config['RDM_SORT_OPTIONS'],
+            '/api/user/communities',
+            {"Accept": "application/json"}
+        ),
+        'search_app_rdm_user_requests_config': partial(
             search_app_config,
             'RDM_SEARCH_USER_REQUESTS',
             current_app.config['REQUESTS_FACETS'],
+            current_app.config['RDM_SORT_OPTIONS'],
```
kpsherva, line 47: I understand there is no sort options for the user requests and we must use the global ones, is that correct?

invenio_app_rdm/records_ui/searchapp.py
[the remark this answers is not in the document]
ntarocco: @kpsherva https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/config.py#L40

setup.py
```diff
@@ -68,6 +68,7 @@
     f'invenio[base,auth,metadata,files]{invenio_version}',
     'invenio-logging[sentry-sdk]>=1.3.0,<1.4.0',
     'invenio-rdm-records>=0.35.1,<0.36.0',
+    'invenio-search-ui>=1.2.0,<1.3.0',
```
ntarocco, line 71: it is already part of `invenio` above, so it should not be needed. Can you please check?

jrcastro2: You are right not needed anymore as it is part of `invenio`

invenio_app_rdm/records_ui/searchapp.py
jrcastro2: Not yet (not sure if there will be), `RDM_SORT_OPTIONS` are the default ones for rdm.

invenio_app_rdm/records_ui/searchapp.py
jrcastro2: I agree, it doesn't look like we can get all these URLs from an existent config (unless I am mistaken), might be worth to create a new one, any suggestion for the name? maybe `RDM_SEARCH_CONFIG_URLS`

ntarocco: As discussed, next PR ;)

zzacharo (approved):

PR #1385, opened 2022-03-15
theme: convert navbar to semantic ui

kpsherva:


jennur (approved):

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/theme.js
```diff
@@ -0,0 +1,14 @@
+
+/* Expand and collapse navbar  */
+const toggle = document.querySelector(".menu-icon");
+const menu = document.querySelector("#invenio-nav");
```
ntarocco, line 4:
```suggestion
const menu = document.getElementById("invenio-nav");
```

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -73,3 +69,37 @@ body {
...
   padding: 1rem 0 !important;
 }
+
+// Export page
+pre.export.result {
+    display:block;
+    white-space: pre-wrap;
+    padding:9.5px;
+    margin:0 0 10px;
+    font-size:13px;
```
ntarocco, line 79: do we use `px` for fonts? I thought that also @jennur was advocating for `em`/`rem`?
Well, apparently, there are [many opinions](https://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css) on this :)

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.variables
```diff
@@ -2,20 +2,43 @@
...
+      Site Colors
+--------------------*/
 
-@linkColor: #005b9e;
-@linkUnderline: none;
-@linkHoverColor: darken(saturate(@linkColor, 20), 15, relative);
-@linkHoverUnderline: underline;
+@green: #048622;
+@signupColor: @green;
+@searchButtonColor: #fb8273;
```
ntarocco, line 19: Shall we define a color instead so that with a name we can immediately understand what color is that? Same above and below?
```suggestion
@searchButtonColor: @<color-name>;
```

invenio_app_rdm/theme/webpack.py
```diff
@@ -44,7 +45,7 @@
                 'react-dnd-html5-backend': '^11.1.0',
                 'react-dropzone': "^11.0.0",
                 "react-i18next": "^11.11.0",
-                'react-invenio-deposit': '^0.18.0',
+                'react-invenio-deposit': '^0.18.1',
```
ntarocco, line 48: no need to increase? It will still get the latest. I removed all the non-zero-versions some time ago.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
jennur: I'm voting for em/rem, yes. Then it'll always scale with the browser settings. Also, semantic ui is already using em/rem.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -73,3 +69,37 @@ body {
...
+//  color: #0377cd;
+//}
+//
+//.record-elem .record-description {
+//  color: #000000;
+//}
+//
+//.panel-free-title {
+//  color: #0377cd;
+//}
```
Pineirin, line 105: Why adding commented css? Doesn't also seem to have a related class.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
kpsherva: this is only copy paste of the code, I can change it in the next iteration, i agree we should use relative units

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
kpsherva: because it is still work in progress and I will remove it later if I find no related element

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/javascript.html
```diff
@@ -0,0 +1,11 @@
+{# -*- coding: utf-8 -*-
+
+  This file is part of Invenio.
+  Copyright (C) 2015-2018 CERN.
```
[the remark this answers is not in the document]
kpsherva: I will fix in the next pr

PR #1386, opened 2022-03-15
requests: inject default query param

NRodriguezcuellar:
- added query params variable to inject into react
closes https://github.com/inveniosoftware/invenio-requests/issues/129

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -105,7 +105,6 @@ BABEL_DEFAULT_LOCALE = 'en'
 BABEL_DEFAULT_TIMEZONE = 'Europe/Zurich'
 """Default time zone."""
 
-
 # Invenio-I18N
 # ============
 # See https://invenio-i18n.readthedocs.io/en/latest/configuration.html
@@ -169,7 +168,6 @@ def files_rest_permission_factory(obj, action):
 FILES_REST_PERMISSION_FACTORY = files_rest_permission_factory
 """Set default files permission factory."""
 
-
 # Invenio-IIIF
 # =================
 # See https://invenio-iiif.readthedocs.io/en/latest/configuration.html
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/requests_ui/views/requests.py
```diff
@@ -22,6 +22,7 @@
 def requests_detail(request=None, pid_value=None):
     """Community detail page."""
     request_dict = request.to_dict()
+    default_query_config = dict(size=15)
```
jrcastro2, line 25: If I am not mistaken, I think the idea of having it here is to allow users to override it easily, to do so I think we should get the size from the config instead of hardcoding it here
```suggestion
    default_query_config = current_app.config[REQUESTS_SEARCH_RESULT_SIZE]
```
And add the `REQUESTS_SEARCH_RESULT_SIZE` so that it can easily be overriden by any instance. 

Feel free to change the name

NRodriguezcuellar: ah fair! thanks :)

invenio_app_rdm/config.py
```diff
@@ -670,3 +670,5 @@ def files_rest_permission_factory(obj, action):
 RDM_REQUESTS_ROUTES = {
     'user-dashboard-community-submission': '/me/requests/<pid_value>',
 }
+
+RDM_REQUESTS_CONVERSATION_PAGE_SIZE = 15
```
kpsherva, line 674: sorry, naming police.
this module does not know anything about conversation tab, in the backend it is called "timeline" everywhere so lets keep it consistent. Additionally, shouldn't this variable be defined in invenio-requests?
```suggestion
RDM_REQUESTS_TIMELINE_PAGE_SIZE = 15
```

invenio_app_rdm/requests_ui/views/requests.py
```diff
@@ -23,6 +23,10 @@ def requests_detail(request=None, pid_value=None):
     """Community detail page."""
```
kpsherva, line 23: just noticed, could you change the docstring to ```Request detail page.```?

invenio_app_rdm/config.py
ntarocco: Move to `invenio-requests`? Why here?
```suggestion
REQUESTS_TIMELINE_PAGE_SIZE = 15
```
Please also add a docstring with explanation

ntarocco (approved):

ntarocco (approved):

zzacharo (approved):

Issue #1337, opened 2022-03-01
User Dashboard tab: replace current implementation 


kpsherva:
Replace Tab component with Menu, due to lack of flexibility of Tab,
see: https://github.com/inveniosoftware/invenio-app-rdm/pull/1259/files#diff-b51234080d58650a7ea3f1c7fc9513a330868fdce4328f318410bca2aa8fb86bR133

PR #1397, opened 2022-03-18
Dashboard: use semantic ui menu instead of tab panes

Pineirin:
Remove communities enabled flag in dashboard.
Replace class use for classname.
closes: inveniosoftware/invenio-app-rdm#1337

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
@@ -11,63 +11,54 @@ import _map from "lodash/map";
 import { SearchApp } from "@js/invenio_search_ui/components";
 import { i18next } from "@translations/invenio_app_rdm/i18next";
 import _camelCase from "lodash/camelCase";
+import _lowerCase from "lodash/lowerCase";
 import React, { Component } from "react";
 import ReactDOM from "react-dom";
 import { overrideStore } from "react-overridable";
-import { Container, Tab, Menu } from "semantic-ui-react";
+import { Container, Segment, Menu } from "semantic-ui-react";
 import { defaultComponents as CommunitiesDefaultComponents } from "./components/communities";
 import { defaultComponents as UploadsDefaultComponents } from "./components/uploads";
 import { defaultComponents as RequestsDefaultComponents } from "./components/requests";
 
 const rootElement = document.getElementById("invenio-user-dashboard");
-const getConfigFromDataAttribute = (element, attr) => {
+const getConfigFromDataAttribute = (_, attr) => {
   const dataValue = rootElement.dataset[attr];
   return JSON.parse(dataValue);
 };
 
-const TAB_PANES = [
-  {
-    configDataAttribute: "invenio-search-user-uploads-config",
-    label: i18next.t("Uploads"),
-    pathname: "uploads",
-  },
-  {
-    configDataAttribute: "invenio-search-user-communities-config",
-    label: i18next.t("Communities"),
-    pathname: "communities",
-  },
-  {
-    configDataAttribute: "invenio-search-user-requests-config",
-    label: i18next.t("Requests"),
-    pathname: "requests",
-  },
-];
-
 const replaceURLPathname = (newPathname) =>
   window.history.replaceState({}, "", `/me/${newPathname}`);
 
-class DashboardTabs extends Component {
+const MENU_ITEMS = {
+  uploads: {
+    configDataAttribute: "invenio-search-user-uploads-config",
+    label: i18next.t("Uploads"),
+  },
+  communities: {
+    configDataAttribute: "invenio-search-user-communities-config",
+    label: i18next.t("Communities"),
+  },
+  requests: {
+    configDataAttribute: "invenio-search-user-requests-config",
+    label: i18next.t("Requests"),
+  },
+};
+
+class DashboardMenu extends Component {
   constructor(props) {
     super(props);
-    const activeTabName = getConfigFromDataAttribute(
+    const activeMenuName = getConfigFromDataAttribute(
       rootElement,
-      _camelCase("active-tab-name")
-    );
-    const communitiesEnabled = getConfigFromDataAttribute(
-      rootElement,
-      _camelCase("communities-enabled")
+      _camelCase("active-menu-element")
     );
 
-    this.ACTIVE_TAB_PANES = communitiesEnabled ? TAB_PANES : [TAB_PANES[0]];
-
-    const routes = this.ACTIVE_TAB_PANES.map((pane) => pane.pathname);
+    const activeMenuElement = activeMenuName;
     this.state = {
-      defaultActiveTab: routes.indexOf(activeTabName),
+      currentActiveMenuElement: activeMenuElement,
     };
-
     // replace URL with the first pathname when not defined
-    if (window.location.pathname.endsWith("/" + activeTabName) === false) {
-      replaceURLPathname(activeTabName);
+    if (window.location.pathname.endsWith(`/${activeMenuName}`) === false) {
+      replaceURLPathname(activeMenuName);
     }
 
     for (const [componentId, component] of Object.entries({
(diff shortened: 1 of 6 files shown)


kpsherva (commented):
To discuss by the team: I think implementing this menu tab as jinja will reduce the amount of code significantly, since we don't keep the selected filters between the tabs anyway

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -78,78 +76,68 @@ class DashboardTabs extends Component {
...
-          <Menu.Item key={index} className="selected-menu-tab">
-            {pane.label}
-          </Menu.Item>
-        ),
-        render: () => (
+        menuLabel: pane.label,
+        content: (
           <Container>
-            <Tab.Pane>
+            <Segment attached="bottom">
```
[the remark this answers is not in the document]
Pineirin: Are you sure? I'm not sure about this, but from my understanding, unless I use the variable in the render, it won't get initialized. I have tested this hypothesis using logs in the constructors and it seems to be true...

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -78,78 +76,68 @@ class DashboardTabs extends Component {
...
-          defaultActiveIndex={this.state.defaultActiveTab}
-          panes={this.panes}
-          menu={{ secondary: true, pointing: true }}
-          onTabChange={this.onTabChange}
-          renderActiveOnly={true}
-        />
+        <div id="dashboard-tab">
+          <Container fluid id="dashboard-tab-menu-container">
+            <Container>
+              {" "}
```
ntarocco, line 120: Why this space?

ntarocco:
> To discuss by the team: I think implementing this menu tab as jinja will reduce the amount of code significantly, since we don't keep the selected filters between the tabs anyway

This needs to be discussed. I prefer the nice experience of having one React app and you can quickly switch between tabs. Having them as different routes, it will require a reload of the page.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -78,78 +71,63 @@ class DashboardTabs extends Component {
...
       <Container fluid>
-        <Element
-          id="dashboard-tab"
-          defaultActiveIndex={this.state.defaultActiveTab}
-          panes={this.panes}
-          menu={{ secondary: true, pointing: true }}
-          onTabChange={this.onTabChange}
-          renderActiveOnly={true}
-        />
+        <div id="dashboard-menu">
```
kpsherva, line 99: please move the id to the container component above

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -25,7 +26,7 @@ const getConfigFromDataAttribute = (element, attr) => {
   return JSON.parse(dataValue);
 };
 
-const TAB_PANES = [
+const TAB_PANELS = [
```
kpsherva, line 29:
```suggestion
const menuElements = [
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -78,78 +71,63 @@ class DashboardTabs extends Component {
...
-          panes={this.panes}
-          menu={{ secondary: true, pointing: true }}
-          onTabChange={this.onTabChange}
-          renderActiveOnly={true}
-        />
+        <div id="dashboard-menu">
+          <Container fluid id="dashboard-menu-container">
+            <Container>
+              <Menu pointing secondary>
+                {this.panels.map(({ menuLabel }, index) => {
```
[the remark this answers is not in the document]
Pineirin: I have designed an alternative solution that reduces the loops to one, but it may be a bit complex if readability is serious concern.

```
    const { currentActiveMenuElement } = this.state;
    const content = this.panels.reduce(
      (previousValue, { menuLabel, appId, config, key }) => {
        previousValue.menuItems.push(
          <Menu.Item
            name={menuLabel}
            key={key}
            active={key === currentActiveMenuElement}
            onClick={this.handleItemClick}
          />
        );
        key === currentActiveMenuElement &&
          (previousValue.view = (
            <Container key={key}>
              <Segment attached="bottom">
                <SearchApp appName={appId} key={appId} config={config} />
              </Segment>
            </Container>
          ));
        return previousValue;
      },
      { menuItems: [], view: null }
    );

    return (
      <Container id="dashboard-menu" fluid>
        <Container fluid id="dashboard-menu-container">
          <Container>
            <Menu pointing secondary>
              {content.menuItems}
            </Menu>
          </Container>
        </Container>
        {content.view}
      </Container>
    );
  }
  ```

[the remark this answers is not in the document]
Comment (jrcastro2) [reply invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js]: IMHO: Ideally the code generating the <Menu.Item> (the loop) should be moved to when the component is mounted (as it is the same always) and the render method should only take care of checking the active method.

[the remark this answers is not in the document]
Comment (Pineirin) [reply invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js]: I'm not 100% sure of this:
To control if Menu.Item element is or isn't active, you must place the active property in the element itself. If we define the Menu.Item in the constructor, the condition will always remain the same (the component is not re-created, is only re-rendered, so the active condition must be placed in the render). Currently, when you press any menu item, the state changes, so the component is re-rendered, with a new active Menu.Item.

File: invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js [hunk 6]
```diff
@@ -78,78 +77,65 @@ class DashboardTabs extends Component {
       overrideStore.add(componentId, component);
     }
 
-    this.panes = this.ACTIVE_TAB_PANES.map((pane, index) => {
+    this.panels = this.menuElements.map((pane) => {
```
ntarocco, line 80:
```suggestion
    this.menuItems = this.menuElements.map((pane) => {
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -25,49 +26,47 @@ const getConfigFromDataAttribute = (element, attr) => {
...
+    const activeMenuName = getConfigFromDataAttribute(
       rootElement,
-      _camelCase("communities-enabled")
+      _camelCase("active-menu-element")
     );
 
-    this.ACTIVE_TAB_PANES = communitiesEnabled ? TAB_PANES : [TAB_PANES[0]];
-
-    const routes = this.ACTIVE_TAB_PANES.map((pane) => pane.pathname);
+    this.menuElements = [
```
ntarocco, line 40: You could have kept it outside the component, so that it was a static obj.
Why having a list and not an obj? You would avoid the `find`:
```
MENU_ITEMS = {
    "uploads": {...},
    ...
}
``` 
and then:
```
handleItemClick = (e, { name }) => {
    const activeMenuElement = MENU_ITEMS[_lowerCase(name)].key
    this.setState({
      currentActiveMenuElement: activeMenuElement,
    });
    replaceURLPathname(_lowerCase(name));
  };
```

ntarocco, line 40:
```suggestion
    this.menuItems = [
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -78,78 +72,68 @@ class DashboardTabs extends Component {
...
-          menu={{ secondary: true, pointing: true }}
-          onTabChange={this.onTabChange}
-          renderActiveOnly={true}
-        />
-      </Container>
+    const { currentActiveMenuElement } = this.state;
+    const items = this.menuItems;
+
+    const activeContent = items[currentActiveMenuElement];
+    const menus = Object.keys(items).map(
```
ntarocco, line 102: See [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries#iterating_through_an_object):
```suggestion
    const menus = Object.keys(items).forEach(([key, value]) => {
        const { menuLabel } = value;
        return (<Menu.Item
            name={menuLabel}
            key={key}
            active={key == currentActiveMenuElement}
            onClick={this.handleItemClick}
          />)}
          );
```

kpsherva (approved):

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -11,63 +11,57 @@ import _map from "lodash/map";
...
-  {
+const replaceURLPathname = (newPathname) =>
+  window.history.replaceState({}, "", `/me/${newPathname}`);
+
+const MENU_ITEMS = {
+  uploads: {
     configDataAttribute: "invenio-search-user-uploads-config",
     label: i18next.t("Uploads"),
-    pathname: "uploads",
+    key: "uploads",
```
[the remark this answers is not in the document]
kpsherva: we could use the pathname as the key, or keep both

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -78,78 +72,66 @@ class DashboardTabs extends Component {
...
-    if (menu.tabular === true && menuPosition === "right") {
-      menu.tabular = "right";
-    }
+      );
+    });
 
     return (
-      <Container fluid id="dashboard-tab-menu-container">
+      <Container id="dashboard-menu" fluid>
+        <Container id="dashboard-menu-container" fluid>
```
kpsherva, line 116:
```suggestion
      <Container id="dashboard-menu-container" fluid>
        <Container id="dashboard-menu" fluid>
```

PR #1399, opened 2022-03-18
members: add css rules

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-communities/issues/458
* closes https://github.com/inveniosoftware/invenio-communities/issues/456

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/grid.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/grid.overrides
@@ -85,6 +85,7 @@
     padding-top: 5px;
     padding-bottom: 5px;
   }
+
 }
 
 #rdm-footer-element {
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
@@ -436,3 +436,11 @@ dl.details-list {
 .shadowless {
   box-shadow: none !important;
 }
+
+.flex-container {
+  display: flex !important;
+}
+
+.contents-container {
+  display: contents !important;
+}
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides
@@ -68,15 +68,14 @@
   }
 }
 
-.ui.relaxed.divided.items {
-  .item.community-member-item {
-    padding: 1em 0em;
-  }
-}
-
-
 /*Description*/
 .item div.description:not(.license-item-description) {
   display: initial !important;
 }
 
+.ui.items {
+  .item.members-dopdown-option {
+    white-space: initial;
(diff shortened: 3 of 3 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/grid.overrides
```diff
@@ -85,6 +85,10 @@
     padding-top: 5px;
     padding-bottom: 5px;
   }
+
+  .column.searchbar-with-filters{
```
kpsherva, line 89:
```suggestion
  .flex-container {
```
I am trying to remove all the naming which is connected with our elements of the interface, and replace them with the semantic way of naming, which is based on the actual "behaviour", which should reduce the amount of custom CSS - `flex-container` becomes reusable not only for search bar filters. it should be also placed in site.overrides

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/input.overrides
```diff
@@ -1,3 +1,7 @@
 /***********************************************
          Invenio App RDM Input Overrides
 ***********************************************/
+
+.ui.input.searchbar-with-filters{
```
kpsherva, line 5: similar comment as above

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -433,3 +433,11 @@ dl.details-list {
   margin-left: auto;
 }
 
+.flex-container {
+  display: flex !important;
+}
+
+
+.contents-container {
```
kpsherva, line 441: what is this class needed for ?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides
```diff
@@ -68,15 +68,14 @@
...
-
 /*Description*/
 .item div.description:not(.license-item-description) {
   display: initial !important;
 }
 
+.ui.items {
+  .item.members-dopdown-option {
+    white-space: initial;
+    min-width: 350px;
```
kpsherva, line 79: please use relative units

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides
```diff
@@ -68,15 +68,14 @@
...
 .item div.description:not(.license-item-description) {
   display: initial !important;
 }
 
+.ui.items {
+  .item.members-dopdown-option {
+    white-space: initial;
+    min-width: 25em;
+  }
+}
```
Pineirin, line 81: Is it ok to not leave a new line at the end of the css or we should?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
kpsherva: I meant more what kind of a problem does it solve, sorry for not being specific. We noticed also yesterday that the elements need to be aligned vertically in the middle (filters are a bit misaligned with the search bar

PR #1403, opened 2022-03-23
header login: add a11y to profile menu

### LINKED ISSUE #1281: Header: The account menu is not announced by a screen reader, nor accessible by keyboard
The items in the dropdown are not announced by the screen reader. The menu is also not fully accessible by keyboard.

Screenshot 2022-02-17 at 16 09 20

jennur:
Closes #1281

- Changed the global account-menu to be dropdown-only, not link + dropdown as before.
- Added aria-attributes for a11y
- Added keyboard accessibility
- Tested with VoiceOver & axe DevTools

## Screenshots
Screenshot 2022-03-23 at 09 42 31
Screenshot 2022-03-23 at 09 42 35

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
@@ -57,22 +57,32 @@
     }
 
     &.right.right-navbar {
-      padding-right: 0;
-      display: flex;
+        display: flex;
+        padding-right: 0;
 
-      .item {
+      > .item {
         padding-left: 15px;
         margin-right: 15px;
       }
     }
   }
 
-  .user-profile {
-    a.ui.button {
-      width: 13em;
+  .rdm-plus-menu {
+    padding-right: 15px;
+  }
+
+  #user-profile-dropdown {
+    width: 13em;
+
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
```diff
@@ -57,13 +57,18 @@
...
+    #user-profile-dropdown-btn {
+      width: 13rem;
       white-space: nowrap;
       overflow: hidden;
       text-overflow: ellipsis;
     }
+
+    #user-profile-menu.ui.menu {
+      margin-left: .25rem;
+      width: 13rem;
```
kpsherva, line 70: why is the width needed? wouldn't it be better to move it to the parent #user-profile-dropdown instead of applying twice the same rule for the children ?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
```diff
@@ -46,7 +46,7 @@
     }
     &.right.right-navbar {
        padding-right: 0;
-      .item {
+      > .item {
```
kpsherva, line 49: 👍

lhenze:
Hi, please rebase?

PR #1415, opened 2022-03-30
theme: add auto adjusted column grid

kpsherva:


NRodriguezcuellar (commented):
Maybe an example of this style applied :smile: ?

ntarocco (approved):

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -430,7 +430,9 @@ dl.details-list {
   box-shadow: none !important;
 }
 
-
-.contents-container {
-  display: contents !important;
+.auto-column-grid {
```
ntarocco, line 433: I would also appreciate understanding (comment? other?) when this should be used.

PR #1455, opened 2022-04-21
request views: reroute request views to general view

NRodriguezcuellar:

closes https://github.com/inveniosoftware/invenio-requests/issues/186
closes https://github.com/inveniosoftware/invenio-requests/issues/184

### DIFF
--- a/invenio_app_rdm/requests_ui/views/requests.py
+++ b/invenio_app_rdm/requests_ui/views/requests.py
@@ -10,50 +10,47 @@
 
 from flask import current_app, g, render_template
 from flask_login import login_required
+from invenio_communities.members.services.request import CommunityInvitation
+from invenio_communities.proxies import current_communities
+from invenio_communities.views.decorators import pass_community
 from invenio_rdm_records.proxies import current_rdm_records_service
+from invenio_rdm_records.requests import CommunitySubmission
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/requests_ui/views/requests.py
```diff
@@ -78,11 +75,55 @@ def requests_detail(request=None, pid_value=None):
...
-    except TemplateNotFound:
-        return render_template(
-            "invenio_requests/details/index.html",
-            # TODO: implement and use request UI serializer
-            request=request_dict,
-            record=topic_record,
-            default_query_config=default_query_config
-        )
+
+def invitation_details(community_pid_value=None,
```
zzacharo, line 79: Shouldn't this require the user to login?

NRodriguezcuellar: it still has the `@login_required decorator` :smile:

zzacharo: I see it in the `requests_detail` view but not on this one, no? If so, maybe you can move it to the `requests_view` one

NRodriguezcuellar: ah my bad, you are right.

zzacharo (approved):

PR #1467, opened 2022-04-22
Communities menu: use flask menu to generate communities menu

Pineirin:
closes: inveniosoftware/invenio-communities#408

### DIFF
--- a/invenio_app_rdm/communities_ui/views/ui.py
+++ b/invenio_app_rdm/communities_ui/views/ui.py
@@ -11,6 +11,7 @@
 from flask import Blueprint, current_app, render_template
 from flask_babelex import lazy_gettext as _
 from flask_login import current_user
+from flask_menu import current_menu
 from invenio_pidstore.errors import PIDDeletedError, PIDDoesNotExistError
 from invenio_records_resources.services.errors import PermissionDeniedError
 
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/communities_ui/views/ui.py
```diff
@@ -56,6 +57,27 @@ def create_ui_blueprint(app):
...
+    @blueprint.before_app_first_request
+    def register_menus():
+        """Register community menu items."""
+        communities = current_menu.submenu('communities')
+        communities.submenu('search').register(
+            'invenio_app_rdm_communities.communities_detail',
+            text=_('Search'),
+            order=1,
+            expected_args=["pid_value"],
+            **communities_kwargs(
```
zzacharo, line 75: I think for simplicity you can inject the kwargs directly here without using the `communities_kwargs` function. If you prefer you can do 
```
**dict(class="search icon", permissions=True)
```

kpsherva (approved):
for housekeeping: we left the icon as a parameter for now until better solution found

the discussion was on the validity of keeping frontend attributes (icon name) in the backend settings (flask-menu). Flask menu implementation enforces us for now to place the icon in the menu items registration, until better solution is found

PR #1471, opened 2022-04-25
deposit form: layout improvements

jennur:
- Adjusted according to updates in PR https://github.com/inveniosoftware/react-invenio-deposit/pull/472

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/form.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/form.overrides
@@ -28,16 +28,12 @@
       margin-bottom: 1rem;
     }
 
-    .additional-descriptions {
-      margin-top: -1rem;
-
-      .ui.grid {
-        margin: 1rem 0 0 0;
-      }
-    }
-
     .invenio-group-field.fields {
       margin: 1rem 0 2rem 0;
+
+      .ui.icon.button.close-btn {
+        margin-top: 1.75rem;
+      }
     }
 
     .additional-descriptions .ui.grid .row,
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/form.overrides
```diff
@@ -60,14 +56,32 @@
...
           position: absolute;
           top: 1rem;
           right: 1.3rem;
+          margin-top: 0;
+        }
+      }
+    }
+
+    .additional-descriptions {
+      margin-top: -1rem;
```
kpsherva, line 69: is it possible to do without negative margins ? they create a lot of issues, especially when used with SUI

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/form.overrides
```diff
@@ -60,14 +56,32 @@
...
+
+      .ui.icon.button.close-btn {
+        margin-top: 0;
+      }
+
+      .ui.grid {
+        margin: 1rem 0 0 0;
```
kpsherva, line 76: seems to be related to the negative margin ? why is it needed ?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/form.overrides
```diff
@@ -60,14 +56,32 @@
...
+
+        .row {
+          padding-bottom: 2rem;
```
kpsherva, line 79: there is a utility class you could use instead `rel-pb-2` unless it is meant to be applied to every row inside of the additional-descriptions

jennur:
Please also review https://github.com/inveniosoftware/react-invenio-forms/pull/128 which is related to this PR.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/form.overrides
jennur: Agreed to keep. Discussed in person.

Issue #1469, opened 2022-04-22
Preview page: Back to edit button misaligned


jennur:
The back to edit button should be aligned with the content below.

<img width="304" alt="Screenshot 2022-04-22 at 16 42 57" src="https://user-images.githubusercontent.com/21052053/164737757-b59e43eb-3a1e-4d05-b3e0-55ea8234478f.png">

Issue #1468, opened 2022-04-22
Preview page: preview & community header styling issues


jennur:
The preview header and the community header should not have spacing between them.

<img width="771" alt="Screenshot 2022-04-22 at 16 27 52" src="https://user-images.githubusercontent.com/21052053/164736634-1de4774b-3e5c-4bc0-be77-d8a28354f7d2.png">

PR #1476, opened 2022-04-25
Swap preview and community header

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1468
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1469

Swapped the community header and preview header, and aligned back-button.

## Screenshot
Screenshot 2022-04-26 at 08 56 50

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -20,43 +20,17 @@
 
 {%- block page_body %}
   <section id="banners" class="banners" aria-label="{{ _('Information banner') }}">
-    {% if is_preview %}
-      <div class="ui info flashed top attached manage message">
-        <div class="ui container">
-          <div class="ui relaxed grid">
-            <div class="column">
-              <div class="row">
-                <div class="header">
-                  <i class="eye icon"></i>
-                  <strong>{{_ ("Preview")}}</strong>
-                </div>
-                <p>
-                  {% if not is_draft %}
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/container.overrides
```diff
@@ -13,6 +13,10 @@
     margin-right: 0 !important;
   }
 
+  & ~ .ui.info.flashed.message.manage {
+    margin-top: -@defaultMargin;
```
kpsherva, line 17: could we avoid the negative margin?

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -72,6 +46,34 @@
       </div>
     {% endif %}
 
+    {% if is_preview %}
+    <div class="ui info flashed top attached manage message">
+      <div class="ui container">
+        <div class="ui relaxed grid">
+          <div class="column">
+            <div class="row">
```
kpsherva, line 54: the grid is not needed here, 
to have the icon in the message it should be replaced with icon class, see https://semantic-ui.com/collections/message.html (Icon message section)

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/container.overrides
jennur: Added comment

kpsherva (approved):

Issue #1486, opened 2022-04-26
accessibility: '.hidden' class causing conflicts 


NRodriguezcuellar:
Currently there is a class in site.overrides called `.hidden` and it is conflicting with the `.hidden` class used by semantic-ui on input fields. It is causing inputs (confirmed for radio buttons, other inputs to be tested) to not be focus-able and important information to not be seen by accessibility software. 

link to class https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides#L17

example of a solution could be (this one is specific for radio buttons)

```less
.hidden {
  &:not(input[type="radio"]) {
    display: none;
  }
}

```

PR #1489, opened 2022-04-27
remove hidden-class from global overrides

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1486

- Removed the hidden-class from `site.overrides` and applied it only to the `h3` header in the sidebar.
- The hidden-class was added in this commit https://github.com/inveniosoftware/invenio-app-rdm/commit/d71ca9d52f0cb12d37a16bb46f03b6314238f96a , just 5 days ago, so unlikely it's used anywhere else.

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/header.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/header.overrides
@@ -33,3 +33,9 @@
     font-size: 1rem;
   }
 }
+
+aside.sidebar {
+  h3.hidden {
+    display: none;
+  }
+}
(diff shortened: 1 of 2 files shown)


NRodriguezcuellar (approved):

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/header.overrides
```diff
@@ -33,3 +33,9 @@
     font-size: 1rem;
   }
 }
+
+aside.sidebar {
+  h3.hidden {
+    display: none;
```
kpsherva, line 39:
```suggestion
aside.sidebar {
   .hidden {
    display: none;
```
not a blocker: would be enough, especially if you want to hide other elements than h3, but it works for this specific case

kpsherva (approved):

PR #1490, opened 2022-04-27
Communities relations

alejandromumo:
closes https://github.com/inveniosoftware/invenio-communities/issues/500


ppanero (commented):
@alejandromumo question: the `types` are not needed anymore because now it is read from db/es, right?

ppanero:
waiting till communities has been released.

ppanero (approved):

Issue #1196, opened 2022-01-27
Add name to Hamburger menu on all pages


lhenze:
From NYU A11y review -

At various responsive break points, the header navigation links move into a hamburger menu. This menu does not have an accessible name.

PR #1494, opened 2022-04-28
Burger menu: add accessibility

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1196

## Changes made (a11y)

- It was necessary to separate the burger-button and the close-menu-button to ensure good accessibility
- Added aria-attributes
- Changed the burger-button from checkbox to button as the label-solution with `role="button"` did not work properly with the screenreader.
- Added click-handlers (also triggered by keyboard because of `<button>`)

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
@@ -108,7 +108,7 @@
 @media all and (max-width: @largestTabletScreen) {
   #invenio-nav.ui.menu {
 
-    &.active .ui.menu {
+    &.active #invenio-menu.ui.menu {
       flex-direction: column;
       position: fixed;
       right: 0;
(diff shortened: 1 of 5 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
```diff
@@ -220,75 +224,62 @@
...
-}
-
-.rdm-menu-btn:checked ~ .rdm-menu-icon .navicon:before {
-  transform: rotate(-45deg);
-}
-
-.rdm-menu-btn:checked ~ .rdm-menu-icon .navicon:after {
-  transform: rotate(45deg);
-}
+  right: -1rem;
```
kpsherva, line 261: 👀 I guess the negative values were not avoidable ?

jennur: It's needed to remain the clickable area of the button and aligning with the content below. In fact this value was off, it should be `-1.5rem`.

Screenshot 2022-04-29 at 11 40 52
Screenshot 2022-04-29 at 11 45 56

lhenze (changes requested):
This looks really good.  I tested with keyboard only, voice control, and voice over.  The only glitch I notice is that "rdm-burger-menu-icon"  is still focusable underneath the expanded panel.  If you were to set the button to "display:none" when the panel is opened,  or otherwise remove it from the DOM, then the tab order would again follow the visual flow of the page.

lhenze (approved):

kpsherva (approved):

Issue #1495, opened 2022-04-28
My dashboard requests tab not responsive


jennur:
Should use the `<GridResponsiveSidebarColumn>` component from `react-invenio-forms`:
https://github.com/inveniosoftware/react-invenio-forms/blob/master/src/lib/elements/GridResponsiveSidebarColumn.js

<img width="543" alt="Screenshot 2022-04-28 at 10 30 20" src="https://user-images.githubusercontent.com/21052053/165711459-08e9955c-cdc5-4874-8caf-3099756484fd.png">

PR #1496, opened 2022-04-28
my dashboard requests: adjust layout for mobile and tablet

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1495

## Screenshots
Screenshot 2022-04-28 at 13 01 44
Screenshot 2022-04-28 at 13 01 30
Screenshot 2022-04-28 at 13 01 16

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
@@ -37,6 +37,7 @@ import {
   Label,
   Segment,
 } from "semantic-ui-react";
+import { GridResponsiveSidebarColumn } from "react-invenio-forms";
 import {
   RDMBucketAggregationElement,
   RDMRecordFacetsValues,
@@ -62,7 +63,7 @@ export const RequestsResults = ({
                   verticalAlign="middle"
                   className="small pt-5 pb-5 highlight-background"
                 >
-                  <Grid.Column width={4}>
+                  <Grid.Column mobile={16} tablet={4} computer={4} className="mt-5 mb-5">
                     <Count
                       label={() => (
                         <>
@@ -72,9 +73,8 @@ export const RequestsResults = ({
                     />
                   </Grid.Column>
                   <Grid.Column
-                    width={12}
-                    textAlign="right"
-                    className="padding-r-5"
+                    mobile={16} tablet={12} computer={12}
+                    className="text-align-right-tablet text-align-right-computer"
                   >
                     {sortOptions && (
                       <Sort
@@ -88,6 +88,7 @@ export const RequestsResults = ({
                     )}
                   </Grid.Column>
                 </Grid.Row>
+
                 <Grid.Row>
                   <Grid.Column>
                     <ResultsList />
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -422,6 +422,29 @@ dl.details-list {
   text-align: center;
 }
 
+.text-align-right {
```
ntarocco, line 425: does it make sense to add this rule to the common `invenio-theme`, maybe?

ntarocco (commented):
It is looking nice! Good work! :)

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
kpsherva: if not already there ? I am not sure, but I think I saw a class handling this

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -422,6 +422,29 @@ dl.details-list {
...
+  text-align: right;
+
+  &-mobile {
+    @media all and (max-width: @largestMobileScreen) {
+      text-align: right;
```
kpsherva, line 430: why is it needed for each device? if you make
```
.text-align-right{
  text-align: right !important;
}
```
(as we use important in the utility classes, like this one, it should do the job for all devices

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
```diff
@@ -144,35 +145,34 @@ export function RequestsResultsItemTemplate({ result, index }) {
...
               user: result.created_by.user,
             })} */}
-            {`opened ${differenceInDays} by you`}
-          </span>
+            {i18next.t('Opened ') + differenceInDays + i18next.t(' by you')}
+          </div>
+
           {result.receiver.community && (
-            <>
+            <div className="inline-computer">
```
kpsherva, line 168:
```suggestion
            <div className="flex">
```
is this class doing the job?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
jennur: It's needed if the alignment is different on different devices.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
jennur: No, but the flex class could be extended to cover this case if that's a better solution.

kpsherva: what do you mean by extended? Could you give more details ?
in case if it is too complicated we can leave it like this for now. 
I think the PR needs a rebase

jennur: The flex class by itself is not enough to display the content correctly. It will have to be combined with sibling classes that make sure the content is breaking to a column for mobile (e.g. `.column-mobile`) and aligns correctly on desktop (e.g. `align-items-center` on the parent or `.align-self-center` to the children  - both of which will need adjustment for column view on mobile). Adding `.flex` to the parent does the same as having the children displaying as inline-blocks on all devices. Adding `.flex` to the children is the same as having them display as blocks.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/requests.js
```diff
@@ -144,35 +145,34 @@ export function RequestsResultsItemTemplate({ result, index }) {
...
               <Icon className="default-margin" name="users" />
               <span className="ml-5">
                 {/* TODO: Replace by resolved receiver */}
                 {/* {result.receiver.community} */}
```
kpsherva, line 172: can be replaced with affiliation?

kpsherva (approved):

Issue #1478, opened 2022-04-26
my dashboard: show upload request status


kpsherva:
Package version (if known): <version>

## Expected behavior

Search result item should display its request status (for drafts only) when awaiting review for community

## Screenshots (if applicable)

![Screenshot 2022-04-26 at 10 17 20](https://user-images.githubusercontent.com/38131488/165254865-8384ee4c-9e31-4cfa-b911-d6fbbb91e91f.png)

PR #1497, opened 2022-04-28
my dashboard: show record status in uploads list

NRodriguezcuellar:
closes: https://github.com/inveniosoftware/invenio-app-rdm/issues/1478

Screenshot from 2022-04-29 13-23-40

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/uploads.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/uploads.js
@@ -34,53 +34,13 @@ import {
 import { axiosWithconfig, SearchItemCreators } from "../../utils";
 import { DashboardResultView, DashboardSearchLayoutHOC } from "./base";
 
-const DeleteDraftButton = (props) => {
-  const [modalOpen, setModalOpen] = useState(false);
-
-  const handleOpen = () => setModalOpen(true);
-
-  const handleClose = () => setModalOpen(false);
-
-  const handleDelete = async () => {
-    const resp = await axiosWithconfig.delete(
-      `/api/records/${props.record.id}/draft`,
-      {},
-      {
-        headers: { "Content-Type": "application/json" },
-      }
-    );
-    handleClose();
-    window.location.reload();
-  };
-
-  return (
-    <>
-      <Button
-        compact
-        size="small"
-        floated="right"
-        color="red"
-        onClick={handleOpen}
-      >
-        <Icon name="trash alternate outline" />
-        {i18next.t("Delete")}
-      </Button>
-
-      <Modal open={modalOpen} onClose={handleClose} size="tiny">
-        <Modal.Content>
-          <h3>{i18next.t("Are you sure you want to delete this draft?")}</h3>
-        </Modal.Content>
-        <Modal.Actions>
-          <Button onClick={handleClose} floated="left">
-            {i18next.t("Cancel")}
-          </Button>
-          <Button color="red" onClick={handleDelete}>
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/uploads.js
```diff
@@ -34,6 +34,15 @@ import {
 import { axiosWithconfig, SearchItemCreators } from "../../utils";
 import { DashboardResultView, DashboardSearchLayoutHOC } from "./base";
 
+const statuses = {
+  in_review: { color: "yellow", title: i18next.t("In review") },
```
ntarocco, line 38: given that such colors are not used in a `grid`, are they matching the upload form?

kpsherva (commented):
the label placement is quite different than in the mockup, and creates misalignment with the text below
in the mockup the status label was quite highlighted by separating its own area in the item
Screenshot 2022-04-28 at 18 02 49

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/uploads.js
Pineirin: The colors roughly match the colors used in the upload form: https://github.com/inveniosoftware/react-invenio-deposit/blob/93d6a0d2ae057f4031b50e614aef33d4acb2a1ac/src/lib/components/DepositStatus/DepositStatusBox.js

kpsherva:
we have now two labels with the same colour (red on `Declined` and `Metadata only`). Would it be an option to change `metadata-only` colour? Red usually indicates negative correlation in the interface, like "Decline" "delete" "remove" while metadata only is more an information. @ntarocco @lnielsen @jennur

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/components/uploads.js
kpsherva: on another topic: those colors are not easily overridable or extendable in the current implementation, they are hardcoded

jennur:
I agree the color of `metadata-only` could be a different color. Another suggestion is to maybe move the status label to the other side of the container to separate it from the metadata-labels.
Screenshot 2022-05-02 at 10 16 35

If we also remove the delete-button:
Screenshot 2022-05-02 at 10 26 30

kpsherva:
@jennur I agree to remove the delete button since it is potentially dangerous action (does it even have a modal to confirm?)

I think however the status of the community submission status should stay on the left side as initially indicated in the mockups, since it is quite important information

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.variables
```diff
@@ -24,10 +24,10 @@
 @signupColor: @green;
 @searchButtonColor: @secondaryColor;
 
-@accessRightOpen: @green;
-@accessRightRestricted: @red;
-@accessRightEmbargoed: @yellow;
-@accessRightMetadataOnly: @red;
+@accessRightOpen: @blue;
```
kpsherva, line 27: please revert this, for the reasons discussed on the teams chat

lhenze:
Hi, I suggest the text treatment of the labels (content type, metadata only, data) should also factor in whether they are actionable or not ...  there is a ticket somewhere for allowing some of them to be clickable and act like facets.  It makes sense to me that "Book" and "Metadata Only" act that way but that date doesn't.

PR #1501, opened 2022-05-02
requests UI: adapt details page header to "referrer"

### LINKED ISSUE #1474: request details: inconsistent views and routes
Package version (if known): <version>

## Describe the bug
Request header should change depending on the initial view from which it was reached from. As an example: It is not currently the case for the community submission, when user arrives to the request details page, the record topic header is visible instead of the "My dashboard" header

## Expected behavior

Screenshot 2022-04-25 at 11 09 53

should replace the header in

Screenshot 2022-04-25 at 11 09 40

## Additional context

1. when the record topic should be displayed ? it seems that there is no use case to display it, since it can be only reached from the community details page (will be replaced with community header) or the user dashboard (will be replaced with "my dashboard" header)
Answer:  the topic header is outdated, it should be replaced either by the community header or my dashboard header, depending on the "parent" view

kpsherva:
* change header depending on if the request was reached
  from communities request search or the user dashboard

* closes #1474

Screenshot 2022-05-03 at 15 00 51
Screenshot 2022-05-03 at 15 00 17
Screenshot 2022-05-03 at 15 00 29


invenio_app_rdm/records_ui/views/records.py
```diff
@@ -78,7 +78,7 @@ def dashboard(dashboard_name=None):
         current_user
     )['avatar']
     return render_template(
-        "invenio_app_rdm/records/dashboard.html",
+        "invenio_app_rdm/users/dashboard.html",
```
ntarocco, line 81: Is this change correct?

invenio_app_rdm/requests_ui/views/requests.py
```diff
@@ -40,58 +45,75 @@ def requests_detail(request=None, default_query_config=None):
...
+        is_draft=is_draft,
+        permissions=permissions,
+        files=[],
+        default_query_config=default_query_config,
+        custom_base_template=custom_base_template,
+        community=community,
+        user_avatar_url=avatar_url,
+        display_menu_only=display_menu_only,
+        types={
+            "organization": _("Organization"),
```
ntarocco, line 91: This was changed by @alejandromumo... I believe that we should not hardcode it anymore.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -87,11 +88,12 @@ class DashboardMenu extends Component {
     this.setState({
       currentActiveMenuElement: activeMenuElement,
     });
-    replaceURLPathname(_lowerCase(name));
+    window.location.replace(`/me/${_lowerCase(name)}`);
```
ntarocco, line 91: It is not clear what is happening here...

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/index.js
```diff
@@ -120,20 +122,36 @@ class DashboardMenu extends Component {
...
-            <div className="rel-pt-2">
-              <SearchApp
-                appName={activeContent.appId}
-                key={activeContent.appId}
-                config={activeContent.config}
-              />
-            </div>
-          </Segment>
-        </Container>
+        {!menuOnly && (
```
ntarocco, line 125: The `menuOnly` is to keep the 3 tabs (Records, Communities, Requests) and show the request details inside?
You mentioned that another solution would be to have the menu in the Jinja template and this would be avoided... feasible in a short time?

invenio_app_rdm/requests_ui/views/requests.py
alejandromumo: `types` is implemented as a vocabulary related to communities in https://github.com/inveniosoftware/invenio-communities/pull/531 , which is still unreleased. 

After it releases, the community has a `type` relation and there's no need to hard-code types anymore. Therefore passing `types` is not needed.  If you want to render a community's type, simply access `community.type` in the template.

kpsherva: I just moved this code to split it, so it is a bit side scope from this PR, and I was not aware of progress with types. Since the communities types is it still not released, is it OK with everyone if it is addressed in a separate PR/issue ?

alejandromumo: I was checking the code of `invenio_requests/community-submission/index.html` and I don't find `types` value  being used there. I tried to check on that template's extends (`invenio_requests/details/index.html`) but I don't find the template. Am I missing something?

If that's the case and `types` is unused, maybe you don't need to inject `types` values in the context.  @kpsherva

PR #1538, opened 2022-05-10
test

NRodriguezcuellar:
closes https://github.com/inveniosoftware/invenio-requests/issues/187


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -439,6 +443,10 @@ dl.details-list {
   display: flex;
 }
 
+.flex-align-center {
```
kpsherva, line 446: there is the same class below

kpsherva, line 446:
```suggestion
.align-items-center {
```
to make it consistent with other naming

Issue #1568, opened 2022-05-13
records details: long titles are overflowing 


NRodriguezcuellar:
When setting a long record title, it overflows on the details page
![Screenshot from 2022-05-13 16-07-23](https://user-images.githubusercontent.com/55200060/168305035-8b062ae9-e32a-4c0e-8ff0-10ff0e9ac0af.png)

![Screenshot from 2022-05-13 16-45-08](https://user-images.githubusercontent.com/55200060/168308585-11af0bb0-ea2a-4fac-8d63-651241e4832d.png)

PR #1573, opened 2022-05-16
record details: fix title overflowing if too long

NRodriguezcuellar:
Now wraps to show the whole title instead of overflowing out of the layout

closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1568

before:
Screenshot from 2022-05-16 09-53-37
Screenshot from 2022-05-17 14-19-28

after:
Screenshot from 2022-05-16 09-52-01
Screenshot from 2022-05-17 14-20-32

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
@@ -68,6 +68,10 @@ button:focus-visible, a:focus-visible {
   box-shadow: 0 2px 5px rgba(0, 0, 0, 0.2);
 }
 
+.wrap-overflowing-text {
+  word-wrap: break-word;
+}
+
 // Log-in and sign-up
 .cover-page {
   background-image: @navbarBackgroundImage;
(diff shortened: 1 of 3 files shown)


zzacharo (approved):

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -68,6 +68,10 @@ button:focus-visible, a:focus-visible {
   box-shadow: 0 2px 5px rgba(0, 0, 0, 0.2);
 }
 
+.wrap-overflowing-text {
+  word-wrap: break-word;
```
kpsherva, line 72: how will word wrapping look like with multiple lines? will it break any layout?

kpsherva: after this change it can. What will happen if you add even more lines? how does it look like in the request details view ? in the searches ?

NRodriguezcuellar: i will test it :+1:

NRodriguezcuellar: So `adding lines` does not really change anything it just a long string. This only affects the record details page.

Pineirin (approved):

jrcastro2:
In the screenshot you shared it looks like there is an empty space in the citation, IMHO the text should fill that space, WDYT?

image

NRodriguezcuellar:
> 

I would say that with text it is usually better to not make it too wide!

kpsherva (approved):

PR #1575, opened 2022-05-16
creatibutors: add accessibility to affiliations accordion

jennur:
Tested with VoiceOver on Mac.

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
@@ -69,8 +69,14 @@
 
 {% macro affiliations_accordion(group, affiliations) %}
 <div class="ui sixteen wide tablet three wide computer column title right aligned bottom aligned">
-    <button class="ui affiliations-button button mini up mr-0" aria-controls="{{ group }}-affiliations">{{ _("Show affiliations") }}</button>
-    <button class="ui affiliations-button button mini down mr-0" aria-controls="{{ group }}-affiliations">{{ _("Hide affiliations") }}</button>
+    <button class="ui affiliations-button button mini mr-0"
+            aria-controls="{{ group }}-affiliations"
+            aria-expanded="false"
+            data-open-text="{{ _('Show affiliations') }}"
+            data-close-text="{{ _('Hide affiliations') }}"
+    >
+      {{ _("Show affiliations") }}
+    </button>
 </div>
 
 <section class="ui sixteen wide column content" id="{{ group }}-affiliations" aria-label="{{ _('Affiliations for') }} {{ group }}">
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/theme.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/theme.js
@@ -37,11 +37,21 @@ $tabElement.on("keydown", function (event) {
   }
 });
 
-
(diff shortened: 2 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/theme.js
```diff
@@ -37,11 +37,19 @@ $tabElement.on("keydown", function (event) {
   }
 });
 
-
+const affiliationsButton = $('.title .affiliations-button');
 $(".ui.accordion.affiliations").accordion({
```
kpsherva, line 41: https://github.com/inveniosoftware/invenio-requests/pull/221 I think this global .ui.accordion class implementation covers also this element, no?

jennur: No, this accordion title has a button as a child, which is the trigger.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/theme.js
```diff
@@ -37,11 +37,19 @@ $tabElement.on("keydown", function (event) {
...
   selector: {
     trigger: ".title .affiliations-button",
   },
+  onOpening: () => {
+    affiliationsButton.attr('aria-expanded', true);
+    affiliationsButton.html("Hide affiliations");
+  },
+  onClosing: () => {
+    affiliationsButton.attr('aria-expanded', false);
+    affiliationsButton.html("Show affiliations");
```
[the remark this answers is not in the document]
jennur: Because it needs to change when opening/closing. If not added here, the text will remain "Hide affiliations". The translation is indeed an issue here, maybe the button text can be added as a data-attribute instead?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/theme.js
kpsherva: so maybe it would make sense to unify them ? to have the same implementation? is there any reason why we need one being triggered by title and the other by button?

jennur: We don't want it to be triggered by the entire title if there is a button, because when it's spanning the full width of the container, that entire field will be clickable, not only the button.

jennur: The selector is specified in the init function, and we can only have one init function per accordion, otherwise they will override each other.

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
```diff
@@ -69,8 +69,14 @@
 
 {% macro affiliations_accordion(group, affiliations) %}
 <div class="ui sixteen wide tablet three wide computer column title right aligned bottom aligned">
-    <button class="ui affiliations-button button mini up mr-0" aria-controls="{{ group }}-affiliations">{{ _("Show affiliations") }}</button>
-    <button class="ui affiliations-button button mini down mr-0" aria-controls="{{ group }}-affiliations">{{ _("Hide affiliations") }}</button>
+    <button class="ui affiliations-button button mini mr-0"
```
kpsherva, line 72:
```suggestion
    <button class="ui trigger button mini mr-0"
```

kpsherva:
agreed to move the changes to invenio-theme

Issue #661, opened 2021-03-02
files: implement garbage collection in files-rest of files instances that have no references.


zzacharo:

lnielsen:
Tracked in https://github.com/inveniosoftware/product-rdm/issues/40

PR #1598, opened 2022-05-17
Members: add responsive layout to tablet resolution [DEPENDS ON https://github.com/inveniosoftware/invenio-communities/pull/671]

Pineirin:
Fix mobile UI issues

closes: [inveniosoftware/invenio-app-rdm#661](https://github.com/inveniosoftware/invenio-communities/issues/661)
closes: https://github.com/inveniosoftware/invenio-communities/issues/646
depends on: https://github.com/inveniosoftware/invenio-communities/pull/671

Screenshot from 2022-05-18 15-49-10
Screenshot from 2022-05-18 15-49-21
Screenshot from 2022-05-18 15-49-34

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
@@ -67,13 +67,25 @@
 }
 
 .ui.dropdown.fluid-mobile {
-  @media all and (max-width: @largestMobileScreen) {
+  @media all and (max-width: @largestTabletScreen) {
     width: 100%;
     min-width: 0em;
     display: block;
 
     & > .dropdown.icon {
       float: right;
+    }
+  }
+}
+
+.mobile-table-cell .ui.dropdown{
+  @media all and (max-width: @largestMobileScreen) {
+    min-width: 7em !important;
+
+    & > .menu {
+      min-width: 14em !important;
+      border-top-width: 1px !important;
+      border-radius: @defaultBorderRadius 0em @defaultBorderRadius @defaultBorderRadius;
     }
   }
 }
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/table.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/table.overrides
@@ -43,4 +43,16 @@
       }
     }
   }
+
+  tbody tr td {
+    &.mobile-table-cell {
+      @media all and (max-width: @largestMobileScreen) {
+        display: inline-grid !important;
+        grid-template-columns: min-content auto;
+        grid-column-gap: 0.5em;
+        align-items: center;
+        width: 100% !important;
+      }
+    }
+  }
 }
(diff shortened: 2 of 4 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
```diff
@@ -77,3 +77,17 @@
     }
   }
 }
+
+.mobile-table-cell .ui.dropdown{
+  @media all and (max-width: @largestMobileScreen) {
+    min-width: 0em !important;
```
kpsherva, line 83: please change to `7em` otherwise it is a bit strange

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
```diff
@@ -77,3 +77,17 @@
...
+
+    & > .menu {
+      min-width: 14em !important;
+      left: auto !important;
+      right: 0px !important;
+      border-top-width: 1px !important;
+      border-radius: 0.28571429rem 0rem 0.28571429rem 0.28571429rem;
```
kpsherva, line 90: please dont hardcode values. There is a variable specifying border-radius. How to find the variable: https://inveniordm.docs.cern.ch/customize/styling-theme/

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
```diff
@@ -77,3 +77,17 @@
     }
   }
 }
+
+.mobile-table-cell .ui.dropdown{
```
kpsherva, line 81: this change should be in dropdown.overrides

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
```diff
@@ -77,3 +77,17 @@
...
+  @media all and (max-width: @largestMobileScreen) {
+    min-width: 0em !important;
+
+    & > .menu {
+      min-width: 14em !important;
+      left: auto !important;
+      right: 0px !important;
+      border-top-width: 1px !important;
```
kpsherva, line 89: why are these needed? could you avoid using `important` ?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
Pineirin: It is on it.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
Pineirin: The rule I'm overriding was marked as important, so I don't think I can. This rule is needed because in semantic ui react, as the dropdown should have the same width than the dropped text, not top margin is required.

Pineirin: The problem with this approach is that there is also a blue line on the bottom of the dropdown when you expand (see screenshot 3). I'm not a fan of this, as it is not consistent with the other views, but after a while trying things I could only find one alternative solution (using before) and it feels quite hackish.

kpsherva: where was it marked as important ?

kpsherva (approved):

zzacharo (approved):

Issue #1584, opened 2022-05-16
Landing page has too much whitespace to the community branding


lnielsen:
Package version (if known): <version>

## Describe the bug

Reducing white space of the marked area.

![Screenshot 2022-05-16 at 16 08 02](https://user-images.githubusercontent.com/1698163/168611769-8b5b61f8-a9d5-4c0c-8737-38cf77dd0fa9.png)

I would expect an identical whitespace as when no community branding is displayed:

![Screenshot 2022-05-16 at 16 10 07](https://user-images.githubusercontent.com/1698163/168612035-8d389a0d-94e5-407e-bcf0-f182392e9bda.png)

PR #1607, opened 2022-05-17
record details: reduce whitespace community header

NRodriguezcuellar:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1584

before:
Screenshot from 2022-05-17 17-01-11

after:
Screenshot from 2022-05-17 17-02-43


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/container.overrides
```diff
@@ -6,7 +6,6 @@
   background-color: @highlightHeaderColor;
   padding-top: @celledPadding;
   padding-bottom: @celledPadding;
-  margin-bottom: @defaultMargin;
```
kpsherva, line 9: could you include screenshots of all the other affected places which are using this class>

NRodriguezcuellar: found a way without breaking anything :smile:

Issue #1595, opened 2022-05-17
List of invitations table should use whitespace


lnielsen:
![Screenshot 2022-05-17 at 11 39 33](https://user-images.githubusercontent.com/1698163/168781297-b4bb2b34-e437-4c3f-8471-34c15a113186.png)

In addition the status  should be capitalise, and use the correct icon.

![Screenshot 2022-05-17 at 12 03 23](https://user-images.githubusercontent.com/1698163/168786301-cdbb295c-a1c6-4985-9051-e211608156c7.png)

PR #1640, opened 2022-05-19
Invitations: fix tablet view [DEPENDENCIES]

Pineirin:
Fix mobile view.
Add status icon.
Add extra space for user full name.
closes: #1595
depends on: https://github.com/inveniosoftware/invenio-communities/pull/681
depends on: https://github.com/inveniosoftware/invenio-requests/pull/238

Screenshot from 2022-05-23 14-59-42
Screenshot from 2022-05-23 14-59-50
Screenshot from 2022-05-23 14-59-58

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/button.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/button.overrides
@@ -38,7 +38,7 @@
 }
 
 .auto-column-grid {
-  .ui.button {
+  .ui.button.dropdown {
     &:last-child {
       margin-right: 0;
     }
@@ -55,10 +55,12 @@
   padding: 0;
 }
 
-.ui.button.fluid-mobile {
-  @media all and (max-width: @largestMobileScreen) {
-    width: 100%;
-    min-width: 0em;
-    display: block;
+.ui.button{
+  &.fluid-responsive {
+    @media all and (max-width: @largestTabletScreen) {
+      width: 100%;
+      min-width: 7em;
+      display: block;
+    }
   }
 }
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
@@ -66,14 +66,15 @@
   min-width: 10em;
 }
 
-.ui.dropdown.fluid-mobile {
-  @media all and (max-width: @largestTabletScreen) {
-    width: 100%;
-    min-width: 0em;
-    display: block;
+.ui.dropdown {
+  &.fluid-responsive {
+    @media all and (max-width: @largestTabletScreen) {
+      width: 100%;
+      display: block;
 
-    & > .dropdown.icon {
-      float: right;
+      & > .dropdown.icon {
+        float: right;
+      }
     }
   }
 }
--- a/invenio_app_rdm/users_ui/views/dashboard.py
+++ b/invenio_app_rdm/users_ui/views/dashboard.py
@@ -47,7 +47,6 @@ def communities():
     url = current_user_resources.users_service.links_item_tpl.expand(
         current_user
     )['avatar']
-    print("HERE")
     return render_template(
         "invenio_app_rdm/users/communities.html",
         searchbar_config=dict(searchUrl=get_search_url()),


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/button.overrides
```diff
@@ -56,7 +56,7 @@
 }
 
 .ui.button.fluid-mobile {
-  @media all and (max-width: @largestMobileScreen) {
+  @media all and (max-width: @largestTabletScreen) {
```
jennur, line 59: Can we rather add another class for tablet `.fluid-tablet` with `@media screen and (min-width: @tabletBreakpoint) and (max-width: @largestTabletScreen)` , so that it's not affecting other cases where the button should not be fluid on tablet? Then apply both `.fluid-mobile` and `.fluid-tablet` to the button that should be fluid on both screen sizes?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/table.overrides
```diff
@@ -22,6 +22,11 @@
       padding-top: 0;
     }
 
+    &:not(.unstackable) thead tr:not(.stackable-content) {
```
kpsherva, line 25: what is the usage of the `.stackable-content` class ?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
```diff
@@ -66,14 +66,28 @@
...
-    min-width: 0em;
-    display: block;
-
-    & > .dropdown.icon {
-      float: right;
+.ui.dropdown {
+  &.fluid-tablet{
+    @media screen and (min-width: @tabletBreakpoint) and (max-width: @largestTabletScreen) {
+      width: 100%;
+      min-width: 0em;
```
kpsherva, line 73: I am not convinced for the min-width:0, I think we should have there a better value adjusted per device, otherwise all the dropdowns will have random minimal lengths based on their content

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/table.overrides
Pineirin: It was to fix the bulk actions in the tablet, but I think that a merge avoided the reason it was included in the first place.

invenio_app_rdm/requests_ui/views/requests.py
```diff
@@ -126,12 +126,12 @@ def user_dashboard_request_view(request, **kwargs):
 @pass_community
 def community_dashboard_request_view(request, community, **kwargs):
     """Community dashboard requests details view."""
-    request_type = request["type"]
-
     avatar = current_user_resources.users_service.links_item_tpl.expand(
         current_user
     )["avatar"]
 
+    request_type = request["type"]
```
kpsherva, line 133: why is this moved below?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/button.overrides
```diff
@@ -55,10 +55,19 @@
...
+    @media screen and (min-width: @tabletBreakpoint) and (max-width: @largestTabletScreen) {
+      width: 100%;
+      min-width: 0em;
+      display: block;
+    }
+  }
+  &.fluid-mobile {
+    @media all and (max-width: @largestMobileScreen) {
+      width: 100%;
+      min-width: 0em;
```
kpsherva, line 69: I think I commented on that, but my comment was removed? 
I would apply min-width at least of 4em, 0 will cause the dropdowns to be very varied between them in length

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/button.overrides
```diff
@@ -55,10 +55,19 @@
...
-.ui.button.fluid-mobile {
-  @media all and (max-width: @largestMobileScreen) {
-    width: 100%;
-    min-width: 0em;
-    display: block;
+.ui.button{
+  &.fluid-tablet {
+    @media screen and (min-width: @tabletBreakpoint) and (max-width: @largestTabletScreen) {
+      width: 100%;
+      min-width: 0em;
```
kpsherva, line 62: I think I commented on that, but my comment was removed? 
I would apply min-width at least of 4em, 0 will cause the dropdowns to be very varied between them in length

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/button.overrides
```diff
@@ -55,10 +55,19 @@
...
+      display: block;
+    }
+  }
+  &.fluid-mobile {
+    @media all and (max-width: @largestMobileScreen) {
+      width: 100%;
+      min-width: 0em;
+      display: block;
```
kpsherva, line 70: it looks like fluid-tablet class is not needed ? since it is code duplication

I would replace it with 
```
&.fluid-responsive {
  @media screen and max-width: @largestTabletScreen {
 ```
 and there is no need to duplicate custom classes

File: invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides [hunk 8]
```diff
@@ -66,14 +66,26 @@
...
-  @media all and (max-width: @largestTabletScreen) {
-    width: 100%;
-    min-width: 0em;
-    display: block;
-
-    & > .dropdown.icon {
-      float: right;
+.ui.dropdown {
+  &.fluid-tablet{
+    @media screen and (min-width: @tabletBreakpoint) and (max-width: @largestTabletScreen) {
```
kpsherva, line 71: same duplication comment as above

invenio_app_rdm/requests_ui/views/requests.py
Pineirin: Merge issues.

kpsherva (approved):

Issue #1581, opened 2022-05-16
Landing page: Move ROR icon to after the organisation name.


lnielsen:
Package version (if known): <version>

## Describe the bug

Move ROR icon to after organisation name similar to the requirements for the ORCID icon

![Screenshot 2022-05-16 at 15 50 26](https://user-images.githubusercontent.com/1698163/168608134-c58f4be7-3c36-4e6d-8f25-4c2948cf295a.png)

lnielsen:
Seems to already be in the works in #1519

PR #1519, opened 2022-05-06
CSS: standardize identity icon display

lhenze:
### Description

Small CSS changes to standardize displays of creatibutors' identity icons between views --  landing page view,  public search results view,  and dashboard view.  This affects the spacing between the name and the icon, the spacing between the name-icon groups, and the size of the icons.

### Checklist

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [ ] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).

PR #1641, opened 2022-05-19
CSS: standardize identity icon display

ppanero:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1581

@kpsherva @jennur I've applied the changes you suggested in https://github.com/inveniosoftware/invenio-app-rdm/pull/1519 and rebased on master.

Screenshot 2022-05-20 at 10 23 28
Screenshot 2022-05-20 at 10 23 17

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/icon.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/icon.overrides
@@ -3,7 +3,8 @@
 ***********************************************/
 
 .inline-id-icon {
-  height: @fontSizeBase;
+  height: 1rem;
+  min-height: 16px;
   width: auto;
   vertical-align: text-bottom;
 }
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads.js
@@ -143,8 +143,10 @@ export const RDMRecordResultsListItem = ({ result, index }) => {
         <Item.Header as="h2">
           <a href={viewLink}>{title}</a>
         </Item.Header>
-        <Item.Meta className="creatibutors">
-          <SearchItemCreators creators={creators} />
+        <Item.Meta>
+          <div className="creatibutors">
+            <SearchItemCreators creators={creators} />
+          </div>
         </Item.Meta>
         <Item.Description>
           {_truncate(description_stripped, {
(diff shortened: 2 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/icon.overrides
```diff
@@ -3,7 +3,8 @@
 ***********************************************/
 
 .inline-id-icon {
```
ppanero, line 5: Note: I did not make the name refactoring as requested in the original PR by @kpsherva because this broke more thing accross modules, e.g.

Screenshot 2022-05-20 at 10 20 00

I think that name refactor can be done as a separate task if needed

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/icon.overrides
```diff
@@ -3,7 +3,8 @@
...
-  height: @fontSizeBase;
+  height: 1rem;
```
zzacharo, line 6: @jennur isn't it better to use a variable here?

jennur: I can't really say, as I don't know the reason for this change, but I'm guessing there is a reason (@lhenze )? In any case I think the value of the variable is the same, so not sure if it matters.

jennur (approved):
Looks good to me!

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/icon.overrides
jennur: Edit: Actually, the `@fontSizeBase` is 14px, so imo having it be 1rem is better to follow the settings of the browser.

ppanero:
@zzacharo after talking to @jennur IRL, we decided to go ahead with the 1rem. Then, she will discuss with @lhenze and decide if its useful to export it to a variable :)

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/icon.overrides
lhenze: The ORCID requirement is "Scale the iD icon to match the line height of the surrounding text but go no smaller than 16×16." so maybe 1em would actually be correct?  (Rem seems to render fine though currently, I don't think the icon appears in any large header text.)

lhenze: The ORCID requirement is "Scale the iD icon to match the line height of the surrounding text but go no smaller than 16×16." so maybe 1em would actually be correct?  (Rem seems to render fine though currently, I don't think the icon appears in any large header text.)

PR #1672, opened 2022-05-30
upgrade: add migration script from v8 to v9

alejandromumo:


invenio_app_rdm/upgrade_scripts/migrate_8_0_to_9_0.py
```diff
@@ -0,0 +1,96 @@
...
+                award = funding.get('award', {})
+                funder = funding.get('funder', {})
+                if award.get('identifier') and award.get('scheme'):
+                    award_identifier = award.pop('identifier')
+                    award_scheme = award.pop('scheme')
+                    funding['award']['identifiers'] = [{'identifier': award_identifier, 'scheme': award_scheme}]
+                if funder.get('identifier') and funder.get('scheme'):
+                    funder_identifier = funder.pop('identifier')
+                    funder.pop('scheme')
+                    funding['funder']['id'] = funder_identifier
```
alejandromumo, line 41: Is this necessary for custom funders? I added it as a preventive measure.

alejandromumo: marking as resolved: since a funder is migrated as a "custom" funder, the "id" property is not needed. Additionally, it could create issues when trying to resolve the funder's id after editing an existing record.

invenio_app_rdm/upgrade_scripts/migrate_8_0_to_9_0.py
```diff
@@ -0,0 +1,98 @@
...
+                        'scheme': award_scheme
+                    }]
+                if funder.get('identifier') and funder.get('scheme'):
+                    funder.pop('identifier')
+                    funder.pop('scheme')
+                if award.get('title', ''):
+                    award_title = award.pop('title')
+                    funding['award']['title'] = {'en': award_title}
+                if funder.get('title', ''):
+                    funder.pop('title')
```
slint, line 48: I think `title` was actually `name` (for "custom" funders). I'm not sure though if this is handled in the deposit form properly...

PR #1674, opened 2022-05-30
ci(publish): pinned babel version

mb-wali:
with the release of `2.10.*` the compile_catalog does not actually compile.
for a quick fix, we restrict the version until `2.9.1`, until the issue is resolved.

issue created: https://github.com/inveniosoftware/invenio-app-rdm/issues/1675

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install setuptools wheel babel
+          pip install setuptools wheel "babel<=2.9.1"
 
       - name: Build package
         run: |


.github/workflows/pypi-publish.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install setuptools wheel babel
+          pip install setuptools wheel "babel<=2.9.1"
```
kpsherva, line 29:
```suggestion
          pip install setuptools wheel "babel==2.9.1"
```
I would go ahead with this since it is a quick fix, also if possible maybe its worth adding a comment with the issue link?

mb-wali:
Closing since it was only logging level change - the `.mo` files are inside the zip packaged.

Issue #1632, opened 2022-05-19
logos: ensure that community logos are not getting cut by default


zzacharo:
We should ensure that any logo is properly displayed and not getting cut in all screen sizes.

## Screenshots

![169040280-6d55858e-26a6-48e6-a7e5-736ff11a87e6](https://user-images.githubusercontent.com/22594783/169260211-2e3ddd31-cb8e-493f-ab5f-b7c5d34a93f2.png)

PR #1677, opened 2022-05-31
css: resize logo to fit content

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1632

image


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/card.overrides
```diff
@@ -70,7 +70,12 @@
     display: flex;
     align-items: center;
     background: transparent;
-    overflow: hidden;
+
+    img {
+      object-fit: scale-down;
```
kpsherva, line 75: was this PR updated? I don't see the newly added placeholder class rules here

jrcastro2: Oups ... I did forget ... Now it's updated. Thanks a lot.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/card.overrides
```diff
@@ -68,9 +68,19 @@
...
+      max-height: 100%;
+      max-width: 100%;
+    }
+
+    &.fallbackImage {
```
kpsherva, line 79:
```suggestion
    &.fallback_image, &.placeholder {
```

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/card.overrides
kpsherva: could you also remove or increase the custom height rules above in line 69 ? it is breaking the default card w/h ratio. In addition, please remove the blank display of description if there is no text, otherwise it is rendered with unused white spaces (this part not in CSS, but react code)
Screenshot 2022-05-31 at 14 31 50

Screenshot 2022-05-31 at 14 30 18

kpsherva: please modify the height to have a square card :)

jrcastro2: Ok, will do :+1: 
FYI: By doing so, the [PR in react-invenio-forms](https://github.com/inveniosoftware/react-invenio-forms/pull/137) is not needed anymore

kpsherva: I think it is good enough :) thank you for changing this

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/card.variables
```diff
@@ -0,0 +1,2 @@
+@largeDesktopScreenCardHeight: 305px;
```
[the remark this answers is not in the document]
jrcastro2: Yes but they are really custom values because the width is being calculated dynamically for the cards. Maybe it would be worth to have a custom css for square cards instead of using SUI `.five.cards`. I also tried to look for a SUI variable to use, instead of creating new ones but couldn't find ....

kpsherva (approved):

PR #1689, opened 2022-06-01
dashboard requests: correct searchbar behaviour

NRodriguezcuellar:
now correctly holds the filter state while switching between open/closed and also when doing an empty search

closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1690

Screenshot from 2022-06-01 15-13-54
Screenshot from 2022-06-01 15-13-48
Screenshot from 2022-06-01 15-12-39

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
@@ -145,12 +145,17 @@ export const RDMRecordSearchBarElement = withState(
     currentQueryState,
   }) => {
     const placeholder = passedPlaceholder || i18next.t("Search");
+
+    const onSearch = () => {
+      updateQueryState({ ...currentQueryState, queryString });
+    };
+
     const onBtnSearchClick = () => {
-      updateQueryState({ ...currentQueryState, filters: [], queryString });
+      onSearch();
     };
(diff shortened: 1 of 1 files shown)


kpsherva (approved):

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
```diff
@@ -145,12 +145,17 @@ export const RDMRecordSearchBarElement = withState(
...
   }) => {
     const placeholder = passedPlaceholder || i18next.t("Search");
+
+    const onSearch = () => {
+      updateQueryState({ ...currentQueryState, queryString });
+    };
+
     const onBtnSearchClick = () => {
-      updateQueryState({ ...currentQueryState, filters: [], queryString });
+      onSearch();
```
jrcastro2, line 154: I am not sure why we were resetting the filters on search but this change will affect several searchbars (like the upload and communities searchbars) as it's `RDMRecordSearchBarElement`, just want to double check that you tested all of them :)

NRodriguezcuellar: i think a misunderstanding from the earlier days of development, also don't think there were any filters before. Thanks for the reminder, all tested :)

Issue #1702, opened 2022-06-07
theme: decrease negative space between facet elements


kpsherva:
Currently there is a lot of negative space between the elements of the facets. It needs slight adjustment to reduce the space. 

![Screenshot 2022-06-07 at 09 44 00](https://user-images.githubusercontent.com/38131488/172330570-55918957-1302-453a-b6b8-aaad2d6c877d.png)

In addition, the first facet is not aligned with the search header, correct the alignment:
![Screenshot 2022-06-07 at 10 13 40](https://user-images.githubusercontent.com/38131488/172330995-a7345975-267c-406f-9573-3947ed8b4e1f.png)

PR #1705, opened 2022-06-07
facets: reduce negative space and align facets with results header

NRodriguezcuellar:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1702

Screenshot from 2022-06-10 10-48-52
Screenshot from 2022-06-08 09-51-44
Screenshot from 2022-06-08 09-51-33

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/card.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/card.overrides
@@ -1,14 +1,15 @@
 .ui.card.borderless.facet {
   box-shadow: none;
 
-  .content {
+  & > .content {
     border-top: none;
+    padding: 0;
+    margin-bottom: @defaultMargin;
 
     h2.header:not(.ui) {
       display: flex;
       align-items: center;
       justify-content: space-between;
-      margin-top: 1rem;
       font-size: 1.1em;
     }
   }


jennur:
@kpsherva I agree. The spacing in this screenshot looks better I think. 
image
Maybe the column-widths are off in the other pages? The sidebar is meant to follow the layout like here https://github.com/inveniosoftware/invenio-search-ui/blob/master/invenio_search_ui/assets/semantic-ui/js/invenio_search_ui/components/SearchApp.js#L135 with 4-column with for the sidebar and 12 on the main content (on desktop).

NRodriguezcuellar:
> @kpsherva I agree. The spacing in this screenshot looks better I think. image ~Maybe the column-widths are off in the other pages? The sidebar is meant to follow the layout like here https://github.com/inveniosoftware/invenio-search-ui/blob/master/invenio_search_ui/assets/semantic-ui/js/invenio_search_ui/components/SearchApp.js#L135 with 4-column with for the sidebar and 12 on the main content (on desktop).~ Edit: It looks like it is a big-screen issue?

it is because the cards have a fixed widths, on big screens that means there is more space left

jennur:
Regarding the latest updated screenshot, I don't think the spacing should be affected in the sidebar for tablet/mobile. It looks a bit cramped now in the slide-in-version of the sidebar. It's only the widest screen that has too much spacing, no?

NRodriguezcuellar:
> Regarding the latest updated screenshot, I don't think the spacing should be affected in the sidebar for tablet/mobile. It looks a bit cramped now in the slide-in-version of the sidebar. It's only the widest screen that has too much spacing, no?

i updated the picture, better like this?

kpsherva:
> i updated the picture, better like this?
definitely better, but I think you did it with CSS while you could change the column width in the grid of the page? Or there is a PR missing?

jennur (approved):

jennur:
I just noticed, after approving, that on the widest screen, the facets are not aligned with the header anymore. Maybe it should be solved by increasing the width of the right column on the biggest screen like Karolina suggested? There is a setting for the widest screen in SUI `widescreen` https://react.semantic-ui.com/collections/grid/ Although the content of the right column will get very wide in that case. Idk what we prefer, more space between the columns and more narrow content,  or less space and wider content.

kpsherva (approved):

Issue #1518, opened 2022-05-06
i18n: `<vocabulary-item>.en` is hardcoded breaking i18n


zzacharo:
We should run this through all repositories as the vocabulary values should always be localized when used in react or Jinja templates.

PR #1706, opened 2022-06-07
[NEEDS #invenio-communities/pull/704]searchapp: update response handler

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1518

### DIFF
--- a/invenio_app_rdm/communities_ui/views/communities.py
+++ b/invenio_app_rdm/communities_ui/views/communities.py
@@ -12,8 +12,8 @@ from flask import render_template
 from invenio_communities.views.decorators import pass_community
 
 
-@pass_community
-def communities_detail(pid_value, community):
+@pass_community(serialize=True)
+def communities_detail(pid_value, community, community_ui):
     """Community detail page."""
     permissions = community.has_permissions_to(
         ["update", "read", "search_requests", "search_invites"]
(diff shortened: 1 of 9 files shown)


invenio_app_rdm/communities_ui/views/communities.py
```diff
@@ -12,8 +12,8 @@
 from invenio_communities.views.decorators import pass_community
 
 
-@pass_community
-def communities_detail(pid_value, community):
+@pass_community(serialize=True)
+def communities_detail(pid_value, community, community_ui):
```
zzacharo, line 16: I don't think we need to keep both `community` and `community_ui` here. The first one is enough :)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -330,7 +328,7 @@ export class RDMDepositForm extends Component {
                     if (funding.funder) {
                       const funderName =
                         funding.funder?.name ??
-                        funding.funder?.title?.en ??
+                        funding.funder?.title ??
```
zzacharo, line 331:
```suggestion
                        funding.funder?.title_l10n
```
?

invenio_app_rdm/communities_ui/views/communities.py
jrcastro2: `Community` is needed to check for the permissions and `community_ui` is needed as it comes with the `ui` key with the necessary translations, why do you think is not needed?

zzacharo (approved):

Issue #1715, opened 2022-06-08
navbar: close icon does not leave any padding on the right side


kpsherva:
![Screenshot 2022-06-08 at 17 24 15](https://user-images.githubusercontent.com/38131488/172660720-8adcca7c-9e19-46d3-935b-16bf2777b3a0.png)

Padding of the hamburger menu and the logo does not align with the site padding
![Screenshot 2022-06-08 at 17 48 34](https://user-images.githubusercontent.com/38131488/172661230-a7ade330-f7ee-45ef-9faa-5cdcdf5a57a0.png)

PR #1729, opened 2022-06-13
mobile menu: add space to close button

NRodriguezcuellar:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1715

Screenshot from 2022-06-14 10-23-00

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
@@ -173,12 +173,7 @@
   width: 100%;
 
   .ui.input {
-    width: 90% !important;
-
-    & > input {
-      width: 90% !important;
-    }
-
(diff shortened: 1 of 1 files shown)


kpsherva:
there is a misalignment with the searchbar on the right side as well, could you fix it ?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
```diff
@@ -250,7 +250,7 @@
       #rdm-close-burger-menu-icon {
         position: absolute;
         top: 2rem;
-        right: 0;
+        right: 2rem;
```
kpsherva, line 253: if you are fixing also misalignment with the searchbar, it would be easier to set common padding on the parent element

jrcastro2 (approved):
LGTM!
However I noticed a few things in the image you posted that would be nice to fix, feel free to create a new task if you prefer
image

PR #1739, opened 2022-06-14
1668 namespace request

fenekku:
- requests: don't override global request in Jinja templates
- extra_requires: remove elasticsearch6


invenio_app_rdm/requests_ui/templates/semantic-ui/invenio_requests/community-invitation/user_dashboard.html
```diff
@@ -6,14 +6,14 @@
...
-{% set title = request.title %}
+{% set title = invenio_request.title %}
 {% extends "invenio_requests/details/index.html" %}
 {% from "invenio_requests/macros/request_header.html" import invitation_request_header %}
 
 
 {% block request_header %}
   {% set back_button_url = url_for("invenio_app_rdm_users.requests") %}
-  {{ invitation_request_header(back_button_url, _("Back to requests"), request=request, accepted=invitation_accepted) }}
+  {{ invitation_request_header(back_button_url, _("Back to requests"), request=invenio_request, accepted=invitation_accepted) }}
```
kpsherva, line 16:
```suggestion
  {{ invitation_request_header(back_button_url, _("Back to requests"), invenio_request=invenio_request, accepted=invitation_accepted) }}
```
shouldn't we also change it in the macro? will it have the same namespacing issues as in the template itself?

fenekku: It doesn't matter in the macro currently, because that macro doesn't use the global `request`. So I don't think it's necessary for now.

kpsherva (approved):

Issue #1714, opened 2022-06-08
members: action dropdowns changing size


kpsherva:
- [x] On the action the dropdown changes size to leave space for the green icon. The size  of the dropdown should not change depending on the action. Test the behaviour also with error returned

![Screenshot 2022-06-08 at 17 20 01](https://user-images.githubusercontent.com/38131488/172659327-120b7219-1e7f-4b73-bc52-1efc54726d00.png)
![Screenshot 2022-06-08 at 17 20 13](https://user-images.githubusercontent.com/38131488/172659330-3b6e196e-4fa0-4363-904a-5cf3c754de54.png)

- [x] dropdown icon is placed misaligned with the dropdown (too big right padding?)

alejandromumo:
Result: 

![Screenshot at Jun 15 11-07-39](https://user-images.githubusercontent.com/21204744/173789844-16c96a04-beb1-408a-b235-53f351664d70.png)

The error popup does not disappear, to be fixed in a new issue

kpsherva:
is it also implemented on the role dropdown and invitations tab?

PR #1741, opened 2022-06-14
members: reserves space for success/error icon

alejandromumo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1714

The issue was that the dropdown container width was adjusting when the icon rendered. E.g. when the icon was rendered, the dropdown container shrinked.

This fix reserves space for the success/error icon by creating a `div`. Does not reduce the dropdown size a lot and it's enough to make the whole cell static when changing the selection.

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
@@ -80,12 +80,10 @@
 }
 
 @minWidthSmallerScreens: calc(100% + 5rem);
-
 .members-dropdown-container {
   .ui.dropdown {
     border: @selectionBorder;
-    padding: @selectionPadding;
-    padding-right: @defaultPadding;
+    padding: @selectionVerticalPadding @selectionHorizontalPadding;
     border-radius: @selectionBorderRadius;
 
     .menu {
@@ -97,7 +95,17 @@
       font-weight: normal;
 
     }
+
+    .dropdown.icon {
+      padding-top: 0.25em;
+    }
   }
+
+.action-status-container {
+  width: 1em;
+  padding: @selectionVerticalPadding @selectionHorizontalPadding;
+  border-radius: @selectionBorderRadius;
+}
 
   @media screen and (max-width: @largestTabletScreen) {
     .ui.dropdown {
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/icon.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/icon.overrides
@@ -20,3 +20,8 @@
     color: @errorTextColor;
   }
 }
+
+.members-icon-container {
+  width: 1em;
+  height: 1em;
+}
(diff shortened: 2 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
```diff
@@ -97,8 +95,18 @@
...
 
     }
+
+    .dropdown.icon {
+      padding-top: 0.25em;
+    }
   }
 
+.members-icon-container {
+  width: 1em;
```
Pineirin, line 105: I'm not sure what is the purpose of the width here.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.variables
```diff
@@ -7,3 +7,5 @@
 // Tablets/phones were left as-is because they have less screen estate
 @selectionComputerMaxMenuHeight : 575px;
 @selectionWidescreenMaxMenuHeight : 575px;
+
+@membersContainerPadding: @selectionVerticalPadding @selectionHorizontalPadding @selectionVerticalPadding @selectionHorizontalPadding;
```
alejandromumo, line 11: On second thought, this might be unnecessary as it is too specific for members table and not that re-usable. What do you think?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
alejandromumo: This is reserving space by creating a small container for the success / error icon. The dropdown is `fluid`, meaning that it will use all the space available. When an icon appears, that same space becomes smaller and the dropdown is re-scaled while the icon appears thus creating this issue.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
```diff
@@ -97,8 +95,18 @@
       font-weight: normal;
 
     }
+
+    .dropdown.icon {
+      padding-top: 0.25em;
```
kpsherva, line 100: this should not be necessary if you use `flex` class on the parent element (if possible)

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.variables
kpsherva: I think maybe not needed as a new variable, you could assign straight to the class:
```less
.members-action-status-container { 
padding: @selectionVerticalPadding @selectionHorizontalPadding;
```
PS. no need to assign the variable for all 4 directions, if you have
```
padding: 1em 2em;
```
it assigns the paddings per "axis" and shortens the notation

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.overrides
alejandromumo: I could not manage to do it with `flex`, the icon inside the SUI Dropdown would not align to the end of the flexbox.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.variables
alejandromumo: thanks for the hints! 👍

kpsherva (approved):

Issue #1736, opened 2022-06-14
Creatibutors: Spacing to icon


jennur:
![creators-icon](https://user-images.githubusercontent.com/21052053/173547530-40054d43-885a-41a7-b2d0-69f993708bf6.png)

PR #1750, opened 2022-06-15
creatibutors: add spacing between icon and separator

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1736

- Added separator class to creatibutors-component in `SearchItemCreators` (to keep same markup in both places).
- Had to specify `.ui.items > .item .meta .creatibutors` for the styling to be preferred over default styling in the search-results case.
- Added spacing to separator class

## Screenshots
Screenshot 2022-06-15 at 14 54 12
Screenshot 2022-06-15 at 14 54 00
Screenshot 2022-06-15 at 14 53 49

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
@@ -571,3 +571,10 @@ dl.details-list {
 .left-floated {
   float: left;
 }
+
+.separated:not(:last-child)::after {
+  content: @listSeparator;
+  display: inline-block;
+  font-size: 1em;
+  margin-left: .1rem;
+}
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html
@@ -39,7 +39,7 @@
 
 {% macro show_creatibutors(creatibutors, show_affiliations=False) %}
   {% for creatibutor in creatibutors %}
-  <dd class="creatibutor-wrap">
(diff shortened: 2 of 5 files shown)


zzacharo (approved):

kpsherva:
I am thinking if it is not an bit of an overkill to wrap the class around only the separators? a lot of typing for one semicolon... :)
alternative idea: what if we solve the matter once for all by implementing pseudoelement for horizontal list items, something like 
```less
.list
.item
&.separated:not(:last-child){
  &::after{
  content: ";"
  mergin-left: 0.1em;
  }
  }
  ```
  
  react usage
```javascript
  <List.Item className="separated" ... >
  ```
  
  html
  
  ````
  <div class="ui list" >
  <div class="item separated">
```

jennur:
## Updates
Removed the separator from the markup, and added it to the CSS within the `.separated`-class, to be used on list items.

Screenshot 2022-06-20 at 10 59 47
Screenshot 2022-06-20 at 10 59 35
Screenshot 2022-06-20 at 10 59 24
Screenshot 2022-06-20 at 10 59 19

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -571,3 +571,10 @@ dl.details-list {
 .left-floated {
   float: left;
 }
+
+.separated:not(:last-child)::after {
```
kpsherva, line 575: 🥳

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -571,3 +571,10 @@ dl.details-list {
...
+  content: ";";
```
kpsherva, line 576: could be as a variable in site.variables
`@listSeparator: ";"`
for easy overriding

PR #1753, opened 2022-06-16
Communities dashboard: re-design the view of the dashboard communities

Pineirin:
closes: inveniosoftware/invenio-communities#701


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities.js
```diff
@@ -52,63 +51,88 @@ function ResultsGridItemTemplate({ result, index }) {
...
-        className="community-logo"
+        className="community-logo computer tablet only"
       />
       <Item.Content>
-        <Item.Extra className="user-communities labels-actions">
-          {/* For reduced spacing between labels. */}
-          <span>
-            <Label size="tiny" color="grey">
+        <Item.Extra className="user-communities labels-actions mobile-centered">
+          {community_type && (
```
kpsherva, line 68: lets discuss the implementation, I think it could be simplified

NRodriguezcuellar:
On tablet and widescreen the community logo looks a bit awkward when there is a community description (it sits aligned with the labels, which should have their own line IMO). In any case i think it would better if it was centered (can be as simple as adding 2 classes to the image container `mt-auto mb-auto`)

NRodriguezcuellar:
> On tablet and widescreen the community logo looks a bit awkward when there is a community description (it sits aligned with the labels, which should have their own line IMO). In any case i think it would better if it was centered (can be as simple as adding 2 classes to the image container `mt-auto mb-auto`)

combined with maybe reducing the margins on the labels seems nice e.g:
Screenshot from 2022-06-22 16-03-54

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/ComputerTabletCommunitiesItem.js
```diff
@@ -0,0 +1,77 @@
...
+  index,
+  community_type,
+  visibility,
+}) => {
+  return (
+    <Item key={index} className="computer tablet only flex">
+      <Image
+        wrapped
+        src={result.links.logo}
+        fallbackSrc="/static/images/square-placeholder.png"
```
Pineirin, line 24: same as below.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/ComputerTabletCommunitiesItem.js
```diff
@@ -0,0 +1,77 @@
...
+      />
+      <Item.Content>
+        <Item.Extra className="user-communities">
+          {community_type && (
+            <Label size="tiny" color="blue">
+              <Icon name="tag" />
+              {community_type}
+            </Label>
+          )}
+          {visibility === "public" ? (
```
Pineirin, line 36: Use only one label and make it capital.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -24,6 +24,9 @@ body {
       &.inline {
         display: inline-block !important;
       }
+      &.flex {
```
jrcastro2, line 27: Instead of adding it everywhere would it make sense to just add important to `flex` class [here](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides#L486) ping @kpsherva

jrcastro2 (commented):
LGTM!

In the screenshot you shared of the tablet view, I would try to fix the alignment between the button and the searchbar, I think there is enough space to keep the button with the same width as in the computer
picture1

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
Pineirin: > LGTM!
> 
> In the screenshot you shared of the tablet view, I would try to fix the alignment between the button and the searchbar, I think there is enough space to keep the button with the same width as in the computer picture1

I can/will patch it a bit for the english language. But, using fixed widths in grid, as we are doing for every view, I cannot guarantee it will work with all locales.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities.js
```diff
@@ -50,67 +48,35 @@ function ResultsGridItemTemplate({ result, index }) {
...
   return (
-    <Item key={index}>
-      <Image
-        wrapped
-        src={result.links.logo}
-        fallbackSrc="/static/images/square-placeholder.png"
-        size="tiny"
-        className="community-logo"
+    <>
+      <ComputerTabletCommunitiesItem
```
alejandromumo, line 63: Why are two components needed? Do Mobile and Computer layout change that much?
Did you consider using composition or another  pattern that allowed some reusability?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/ComputerTabletCommunitiesItem.js
```diff
@@ -0,0 +1,77 @@
...
+          </Button>
+        </Item.Extra>
+        <Item.Header as="h2">
+          <a href={`/communities/${result.id}`}>{result.metadata.title}</a>
+        </Item.Header>
+        <Item.Meta>
+          <div
+            className="truncate-lines-2"
+            dangerouslySetInnerHTML={{
+              __html: result.metadata.description,
```
alejandromumo, line 59: did you consider any other way of displaying the description? Can the metadata description be used for XSS? React [https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml](docs) alert for this

Pineirin: Maybe we should have a general discussion about this topic and evaluate alternatives, as the dangerouslySetInnerHTML tag is used in a fair amount of places in InvenioRDM.

kpsherva: I would say this is out of the scope for this PR

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities.js
```diff
@@ -50,67 +48,35 @@ function ResultsGridItemTemplate({ result, index }) {
...
+  const community_type = result.ui?.type?.title_l10n;
+  const visibility = result.access.visibility;
+  let visibilityColor = "red";
+  let visibilityText = i18next.t("Restricted");
+  let visibilityIcon = "ban";
+  if (visibility === "public") {
+    visibilityColor = "green";
+    visibilityText = i18next.t("Public");
+    visibilityIcon = undefined;
+  }
```
kpsherva, line 60: I think this part needs a little simplification
```suggestion
const isPublic = visibility === "public";
const visibilityColor = isPublic ? "green": "red" ;
const visibilityText = isPublic ?  i18next.t("Public") : i18next.t("Restricted");
```
additionally I think you don't need the middleman component `CommunitiesResultsItemTemplate` you can directly use ComputerTabletCommunitiesItem and MobileCommunitiesItem inside the parent directly, and compute the values above in each component, making it more flexible and less complex

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/ComputerTabletCommunitiesItem.js
```diff
@@ -0,0 +1,77 @@
...
+import { Image } from "react-invenio-forms";
+import { Button, Icon, Item, Label } from "semantic-ui-react";
+
+export const ComputerTabletCommunitiesItem = ({
+  result,
+  index,
+  community_type,
+  visibilityColor,
+  visibilityText,
+  visibilityIcon
```
kpsherva, line 19: all of these props should be removed, we need only to accept result to simplify the interface of this component

kpsherva (approved):

Issue #1754, opened 2022-06-16
Make custom .only-classes more flexible


jennur:
In `site.overrides`: 

- [x] Adjust the `.only` classes to be default `display: block`
- [x] Add `.inline` to each breakpoint with `display: inline-block`
- [x] Make sure to update all cases where `.only` is used (if needed), **NOT** in combo with `.ui.grid`. 
- [x] Search both `invenio-app-rdm` and `invenio-communities` for custom `.only`, and check the UI thoroughly, on all screen sizes, to make sure everything looks as expected.

PR #1755, opened 2022-06-16
site overrides: update only classes to include both block and inline-…

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1754


kpsherva (approved):

Pineirin (approved):

Pineirin:
> are there any other repositories we should update the classes accordingly or only this one ?

There was another use of it on invenio-communities, https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/templates/semantic-ui/invenio_communities/details/header.html. But maybe @jennur tested that no changes were needed.

Issue #1716, opened 2022-06-08
Communities: Restrict logo upload to a specific size


kpsherva:
- [ ] Implement a restriction (validation) for the community logo size (backend)
- [ ] add information about the required image size in the upload box on the frontend

PR #1767, opened 2022-06-29
settings: added logo size quota.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1716

- Added maximum size allowed for community logos.
  - By default it is set to 1 MB, can be modified per instance. 
  - The magic number 1MB was decided based on a search about image standards for web apps. For logos the recommended size was 500 KB, thus an extra 500KB were given. 

Note: this change should be documented for the instance configuration.


ntarocco (commented):
See my comments in the `invenio-communities` PR, config to be moved there?

Issue #1771, opened 2022-06-30
record list item: improve mobile view


jennur:
The mobile view for the record list items needs some improvements and alignment with the layout for request-list items.

## Current
<img width="367" alt="Screenshot 2022-06-30 at 16 20 07" src="https://user-images.githubusercontent.com/21052053/176701265-cd4cd1d4-ace1-4d66-a7a7-1384409cbd37.png">

On mobile/tablet:
- Icon should be next to header

On mobile:
- Buttons should be moved to bottom

## Mockups

<img width="433" alt="Screenshot 2022-06-30 at 14 18 25" src="https://user-images.githubusercontent.com/21052053/176700862-a567e3de-9aaf-4052-af73-3ed83195e341.png">
<img width="772" alt="Screenshot 2022-06-30 at 14 16 35" src="https://user-images.githubusercontent.com/21052053/176700873-5c5a0fd1-6fc5-424d-ab4f-bf41a35018bf.png">

PR #1778, opened 2022-07-04
Dashboard uploads: refactor uploads view

Pineirin:
Remove empty content.
Create a custom mobile interface.
Fix request_items naming.
closes: #1771

Computer:
Screenshot from 2022-07-05 15-37-44

Tablet:
Screenshot from 2022-07-05 15-37-53

Mobile:
Screenshot from 2022-07-05 15-38-04

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/ComputerTabletUploadsItem.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/ComputerTabletUploadsItem.js
@@ -1,0 +1,124 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2022 CERN.
+//
+// Invenio App RDM is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import { i18next } from "@translations/invenio_app_rdm/i18next";
+import _truncate from "lodash/truncate";
+import React from "react";
+import { Button, Icon, Item, Label } from "semantic-ui-react";
+import { SearchItemCreators } from "../../utils";
+import PropTypes from "prop-types";
+
+export const ComputerTabletUploadsItem = ({
+  result,
+  editRecord,
+  statuses,
+  access,
+  uiMetadata,
+}) => {
+  const { accessStatusId, accessStatus, accessStatusIcon } = access;
+  const {
+    descriptionStripped,
+    title,
+    creators,
+    subjects,
+    publicationDate,
+    resourceType,
+    createdDate,
+    version,
+    isPublished,
+    viewLink,
+  } = uiMetadata;
+
+  const icon = isPublished ? (
+    <Icon name="check" className="positive" />
+  ) : (
+    <Icon name="upload" className="negative" />
+  );
+
+  return (
+    <Item
+      key={result.id}
+      className="deposits-list-item mb-20 computer tablet only flex"
+    >
+      <div className="status-icon mr-10">
+        <Item.Content verticalAlign="top">
+          <Item.Extra>{icon}</Item.Extra>
+        </Item.Content>
+      </div>
+      <Item.Content>
+        <Item.Extra className="labels-actions">
+          {result.status in statuses && result.status !== "published" && (
+            <Label size="tiny" className={statuses[result.status].color}>
+              {statuses[result.status].title}
+            </Label>
+          )}
+          <Label size="tiny" className="primary">
+            {publicationDate} ({version})
+          </Label>
+          <Label size="tiny" className="neutral">
+            {resourceType}
+          </Label>
+          <Label size="tiny" className={`access-status ${accessStatusId}`}>
+            <i className={`icon ${accessStatusIcon}`} />
+            {accessStatus}
+          </Label>
+          <Button compact size="small" floated="right" onClick={() => editRecord()}>
+            <Icon name="edit" />
+            {i18next.t("Edit")}
+          </Button>
+          {isPublished && (
+            <Button compact size="small" floated="right" href={viewLink}>
+              <Icon name="eye" />
+              {i18next.t("View")}
+            </Button>
+          )}
+        </Item.Extra>
+        <Item.Header as="h2">
+          <a href={viewLink} className="truncate-lines-2">
+            {title}
+          </a>
+        </Item.Header>
+        <Item.Meta>
+          <div className="creatibutors">
+            <SearchItemCreators creators={creators} />
+          </div>
+        </Item.Meta>
+        <Item.Description>
+          {_truncate(descriptionStripped, {
+            length: 350,
+          })}
+        </Item.Description>
+        <Item.Extra>
+          {subjects.map((subject) => (
+            <Label key={subject.title_l10n} size="tiny">
+              {subject.title_l10n}
+            </Label>
+          ))}
+          <div>
+            {}
+            <small>
+              {createdDate ? (
+                <>
+                  {i18next.t("Uploaded on")} <span>{createdDate}</span>
+                </>
+              ) : (
+                i18next.t("No creation date found.")
+              )}
+            </small>
+          </div>
+        </Item.Extra>
+      </Item.Content>
+    </Item>
+  );
+};
+
+ComputerTabletUploadsItem.propTypes = {
+  result: PropTypes.object.isRequired,
+  editRecord: PropTypes.func.isRequired,
+  statuses: PropTypes.object.isRequired,
+  access: PropTypes.object.isRequired,
+  uiMetadata: PropTypes.object.isRequired,
+};
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/MobileUploadsItem.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/MobileUploadsItem.js
@@ -1,0 +1,133 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2022 CERN.
+//
+// Invenio App RDM is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import { i18next } from "@translations/invenio_app_rdm/i18next";
+import _truncate from "lodash/truncate";
+import React from "react";
+import { Button, Icon, Item, Label } from "semantic-ui-react";
+import { SearchItemCreators } from "../../utils";
+import PropTypes from "prop-types";
+
+export const MobileUploadsItem = ({
+  result,
+  editRecord,
+  statuses,
+  access,
+  uiMetadata,
+}) => {
+  const { accessStatusId, accessStatus, accessStatusIcon } = access;
+  const {
+    descriptionStripped,
+    title,
+    creators,
+    subjects,
+    publicationDate,
+    resourceType,
+    createdDate,
+    version,
+    isPublished,
+    viewLink,
+  } = uiMetadata;
+
+  const icon = isPublished ? (
+    <Icon name="check" className="positive" />
+  ) : (
+    <Icon name="upload" className="negative" />
+  );
+
+  return (
+    <Item key={result.id} className="deposits-list-item mb-20 mobile only flex">
+      <Item.Content className="centered">
+        <Item.Extra className="labels-actions">
+          {result.status in statuses && result.status !== "published" && (
+            <Label size="tiny" className={statuses[result.status].color}>
+              {statuses[result.status].title}
+            </Label>
+          )}
+          <Label size="tiny" className="primary">
+            {publicationDate} ({version})
+          </Label>
+          <Label size="tiny" className="neutral">
+            {resourceType}
+          </Label>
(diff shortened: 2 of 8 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/ComputerTabletUploadsItem.js
```diff
@@ -0,0 +1,120 @@
...
+      </div>
+      <Item.Content style={{ cursor: "default" }}>
+        <Item.Extra>
+          {result.status in statuses && result.status !== "published" && (
+            <Label size="tiny" color={statuses[result.status].color}>
+              {statuses[result.status].title}
+            </Label>
+          )}
+          <Label size="tiny" color="blue">
+            {publicationDate && publicationDate} ({version && version})
```
alejandromumo, line 61: If `publicationDate === undefined` && `version === undefined`, would there be an empty `Label` component?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/ComputerTabletUploadsItem.js
```diff
@@ -0,0 +1,120 @@
...
+          </div>
+        </Item.Meta>
+        {description_stripped && (
+          <Item.Description>
+            {_truncate(description_stripped, {
+              length: 350,
+            })}
+          </Item.Description>
+        )}
+        <Item.Extra>
```
alejandromumo, line 103: Same suggestion as below, `subjects` might be an empty array and `createdDate` might be `undefined`. Leaving the code with an empty `Item.Extra` component.

```suggestion
        { 
          (subjects || createdDate) && 
            <Item.Extra>
                ....
           </Item.Extra>
        }
```

`createdDate` is validated twice in this suggestion though.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/MobileUploadsItem.js
```diff
@@ -0,0 +1,122 @@
...
+import { i18next } from "@translations/invenio_app_rdm/i18next";
+import _get from "lodash/get";
+import _truncate from "lodash/truncate";
+import React from "react";
+import { Button, Icon, Item, Label } from "semantic-ui-react";
+import { SearchItemCreators } from "../../utils";
+
+export const MobileUploadsItem = ({ result, index, editRecord, statuses }) => {
+  const access_status_id = _get(result, "ui.access_status.id", "open");
+  const access_status = _get(result, "ui.access_status.title_l10n", "Open");
```
alejandromumo, line 16: Same as before regarding internasionalisation

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/utils.js
```diff
@@ -43,7 +43,9 @@ const apiConfig = {
 
 export const axiosWithconfig = axios.create(apiConfig);
 
-export function SearchItemCreators({ creators }) {
+export function SearchItemCreators({ creators, className }) {
+  const spanClass = className ? className : "creatibutor-wrap separated";
```
alejandromumo, line 47:
```suggestion
  const spanClass = className ?? "creatibutor-wrap separated";
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/utils.js
```diff
@@ -56,12 +58,12 @@ export function SearchItemCreators({ creators }) {
         icon = "/static/images/orcid.svg";
         break;
       case "ror":
-        link = "https://ror.org/"+identifier;
+        link = "https://ror.org/" + identifier;
```
alejandromumo, line 61: As a personal preference I use string templating thanks to previous eslint rules. See [this](https://eslint.org/docs/latest/rules/prefer-template) 

```suggestion
        link = `https://ror.org/${identifier}`;
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/utils.js
```diff
@@ -56,12 +58,12 @@ export function SearchItemCreators({ creators }) {
...
         linkTitle = i18next.t("ROR profile");
         icon = "/static/images/ror-icon.svg";
         break;
       case "gnd":
-        link = "https://d-nb.info/gnd/"+identifier;
+        link = "https://d-nb.info/gnd/" + identifier;
```
alejandromumo, line 66: Same suggestion for string templating

alejandromumo:
LGTM, left smaller comments / suggestions.

Good work! Views are looking good 👍

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/ComputerTabletUploadsItem.js
```diff
@@ -0,0 +1,143 @@
...
+        </Item.Description>
+        <Item.Extra>
+          {subjects.map((subject, index) => (
+            <Label key={index} size="tiny">
+              {subject.title_l10n}
+            </Label>
+          ))}
+          <div>
+            <small>
+              {i18next.t("Uploaded on")} <span>{createdDate}</span>
```
alejandromumo, line 136: In case `createdDate` is not defined, you might  end up with a label like:

`Uploaded on No creation date found.`

Is it intended?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/ComputerTabletUploadsItem.js
```diff
@@ -0,0 +1,143 @@
...
+    result,
+    "ui.publication_date_l10n_long",
+    i18next.t("No publication date found.")
+  );
+  const resource_type = _get(
+    result,
+    "ui.resource_type.title_l10n",
+    i18next.t("No resource type")
+  );
+  const version = result.ui?.version;
```
alejandromumo, line 58: `version` might be `undefined` and is later printed to the user as a `label`. 

Maybe set `version` to `''`?
```suggestion
  const version = result.ui?.version ?? '';
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/utils.js
```diff
@@ -108,7 +106,7 @@ export function SearchItemCreators({ creators }) {
     return link;
   }
   return creators.map((creator, index) => (
-    <span className="creatibutor-wrap separated" key={index}>
+    <span className={spanClass} key={index}>
```
jennur, line 109:
```suggestion
    <span className={`creatibutor-wrap separated ${className}`} key={index}>
```
Can we keep "creatibutor-wrap separated" always by default, then we don't need to add these classes to the mobile view.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/MobileUploadsItem.js
```diff
@@ -0,0 +1,145 @@
...
+import { i18next } from "@translations/invenio_app_rdm/i18next";
+import _get from "lodash/get";
+import _truncate from "lodash/truncate";
+import React from "react";
+import { Button, Icon, Item, Label } from "semantic-ui-react";
+import { SearchItemCreators } from "../../utils";
+
+export const MobileUploadsItem = ({ result, index, editRecord, statuses }) => {
+  const access_status_id = _get(
+    result,
```
jennur, line 16: Could maybe all these shared variables be collected in the parent and passed down to the computer/mobile component, so that we don't need to define them twice?

Pineirin: I agree with you, but I got some backlash in another similar PR about having components with a huge number of parameters passed. I would say we need to establish clearly what should and should not get duplicated @kpsherva.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/utils.js
Pineirin: I'm not sure if it would be best to allow the component to have all it's className overridden, as it would make it more reusable. I've made the change but I would like if others in the team give their opinion.

NRodriguezcuellar: if they should always have this class and will otherwise break the component ui then it makes sense to always pass it, otherwise i would prefer to make it completely overridable

jennur: As far as I know, the classes belong to the component and should not be overridden. It is the name with connected identifier icon which we want to keep consistent in styling across the entire site.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/UploadsResultProps.js
```diff
@@ -0,0 +1,39 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2022 CERN.
+//
+// Invenio App RDM is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import { i18next } from "@translations/invenio_app_rdm/i18next";
+import _get from "lodash/get";
+
+export const uploadsResultProps = (result) => {
```
kpsherva, line 10: it's quite unorthodox in our code base to separate a function to provide the props
additionally, as a developer I wouldn't like to jump to another file to find out how my props are calculated.
most of these statements can be simplified, let me know if you need hints on how - please place them in the places they are used

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/MobileUploadsItem.js
```diff
@@ -0,0 +1,133 @@
...
+
+MobileUploadsItem.propTypes = {
+  result: PropTypes.object.isRequired,
+  index: PropTypes.string,
+  editRecord: PropTypes.func.isRequired,
+  statuses: PropTypes.object.isRequired,
+};
+
+MobileUploadsItem.defaultProps = {
+  index: null,
```
Pineirin, line 132: Remove nulls from defaultProps (use undefined instead).

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/ComputerTabletUploadsItem.js
```diff
@@ -0,0 +1,121 @@
...
+  );
+
+  return (
+    <Item key={index} className="deposits-list-item mb-20 computer tablet only flex">
+      <div className="status-icon mr-10">
+        <Item.Content verticalAlign="top">
+          <Item.Extra>{icon}</Item.Extra>
+        </Item.Content>
+      </div>
+      <Item.Content style={{ cursor: "default" }}>
```
kpsherva, line 45: could you remove the inline style?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/ComputerTabletUploadsItem.js
```diff
@@ -0,0 +1,124 @@
...
+          })}
+        </Item.Description>
+        <Item.Extra>
+          {subjects.map((subject) => (
+            <Label key={subject.title_l10n} size="tiny">
+              {subject.title_l10n}
+            </Label>
+          ))}
+          <div>
+            {}
```
NRodriguezcuellar, line 101: does this do anything?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/MobileUploadsItem.js
```diff
@@ -0,0 +1,133 @@
...
+        </Item.Extra>
+        <Item.Extra>
+          <Button
+            compact
+            size="small"
+            floated="right"
+            onClick={() => editRecord()}
+            className="fluid-responsive"
+          >
+            <Icon name="edit" />
```
NRodriguezcuellar, line 106: maybe there is a difference in how it looks but otherwise you can also pass the `icon` prop to the button 😄 . Same as below

NRodriguezcuellar (approved):

kpsherva (approved):

PR #1781, opened 2022-07-05
poc: implement custom fields in the deposit form

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1782

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -251,6 +251,21 @@ class VocabulariesOptions:
         return self._vocabularies
 
 
+def load_custom_fields(conf_ui, conf_backend):
+    """Load custom fields configuration."""
+    from invenio_vocabularies.custom_fields import VocabularyCF
+
+    for section_cfg in conf_ui:
+        vocabulary_fields = []
+        fields = section_cfg["fields"]
+        for field in fields:
+            field_type = conf_backend.get(field["field"])
+            if isinstance(field_type, VocabularyCF):
+                field["options"] = field_type.options()
+                vocabulary_fields.append(field["field"])
+    return {"ui": conf_ui, "vocabularies": vocabulary_fields}
+
+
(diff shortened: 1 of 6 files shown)


invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -251,6 +251,21 @@ def dump(self):
...
+def load_custom_fields(conf_ui, conf_backend):
+    """Load custom fields configuration."""
+    from invenio_vocabularies.custom_fields import VocabularyCF
+
+    for section_cfg in conf_ui:
+        vocabulary_fields = []
+        fields = section_cfg["fields"]
+        for field in fields:
+            field_type = conf_backend.get(field["field"])
+            if isinstance(field_type, VocabularyCF):
```
ppanero, line 263: we now have a `realtion_cls` attr in the field obj. so we can check if None or not. is implicit but avoid cast check

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/custom_fields/CustomAutocompleteDropdown.js
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+
+import { FieldLabel, RemoteSelectField } from "react-invenio-forms";
+import _get from 'lodash/get';
+import _isArray from 'lodash/isArray';
+import { Field } from 'formik';
+
+export default class CustomAutocompleteDropdown extends Component {
```
ppanero, line 9: wondering if somehow this can be reusable for othervocabularies (not just custom fields ones). e.g. subjects, names

invenio_app_rdm/records_ui/views/deposits.py
ppanero: https://github.com/inveniosoftware/invenio-records-resources/issues/358

Issue #1770, opened 2022-06-30
Community names with large words break the layout on mobile


alejandromumo:
When a community has a title with large words, the whole page layout breaks in mobile. See:

![Screenshot at Jun 30 10-40-19](https://user-images.githubusercontent.com/21204744/176633295-c49f6f3b-ad83-482b-8912-c81595913154.png)

kpsherva:
after testing, found more issues:

![Image](https://user-images.githubusercontent.com/38131488/195841047-1e3652c9-53e7-4f26-84c1-3d8963a511e7.png)
![Image](https://user-images.githubusercontent.com/38131488/195841046-b8bf3e57-de1a-47fc-ae42-3d67c47d6571.png)

PR #1794, opened 2022-07-08
communities: added class to access-status div to set its width to fit…

alejandromumo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1770

- Set property `word-break: break-word` to headers. For headers with long words they were not breaking thus creating issues in the UI.
- Added a new class for `access-status` divs that enclose `span` and `icon` elements. See https://github.com/inveniosoftware/invenio-communities/pull/757  for comments about this.

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/ComputerTabletCommunitiesItem.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/ComputerTabletCommunitiesItem.js
@@ -18,7 +18,7 @@ export const ComputerTabletCommunitiesItem = ({ result, index }) => {
   const visibilityText = isPublic ? i18next.t("Public") : i18next.t("Restricted");
   const visibilityIcon = isPublic ? undefined : "ban";
   return (
-    <Item key={index} className="computer tablet only flex">
+    <Item key={index} className="computer tablet only flex community-item">
       <Image
         wrapped
         src={result.links.logo}
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/MobileCommunitiesItem.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/MobileCommunitiesItem.js
@@ -18,7 +18,7 @@ export const MobileCommunitiesItem = ({ result, index }) => {
   const visibilityText = isPublic ? i18next.t("Public") : i18next.t("Restricted");
   const visibilityIcon = isPublic ? undefined : "ban";
   return (
-    <Item key={index} className="mobile only">
+    <Item key={index} className="mobile only community-item">
       <Item.Content className="centered">
         <Item.Extra className="user-communities">
           {communityType && (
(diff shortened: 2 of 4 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/header.overrides
```diff
@@ -39,3 +39,7 @@ aside.sidebar {
     display: none;
   }
 }
+
+.ui.header  {
```
kpsherva, line 43: this is modifyng the theme globally. Are you sure this is what you want to do?

jennur: Are there any downsides of having it apply to all? It will only make sure that no headings are overflowing its container, no?

Pineirin: Maybe it could create an oversized container that causes layout problems. A better solution would be to truncate IMO, but as a default case I think that word-break: break-word seems better.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/ComputerTabletCommunitiesItem.js
```diff
@@ -18,7 +18,7 @@ export const ComputerTabletCommunitiesItem = ({ result, index }) => {
   const visibilityText = isPublic ? i18next.t("Public") : i18next.t("Restricted");
   const visibilityIcon = isPublic ? undefined : "ban";
   return (
-    <Item key={index} className="computer tablet only flex">
```
alejandromumo, line 21: This does not seem to be used by `app-rdm`. There is an implementation there and that's the one being used.
However, I fixed it in both places.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/ComputerTabletCommunitiesItem.js
```diff
@@ -18,7 +18,7 @@ export const ComputerTabletCommunitiesItem = ({ result, index }) => {
...
+    <Item key={index} className="computer tablet only flex community-item">
```
alejandromumo, line 21: This does not seem to be used by `app-rdm`. There is an implementation there and that's the one being used.
However, I fixed it in both places.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/communities_items/MobileCommunitiesItem.js
```diff
@@ -18,7 +18,7 @@ export const MobileCommunitiesItem = ({ result, index }) => {
   const visibilityText = isPublic ? i18next.t("Public") : i18next.t("Restricted");
   const visibilityIcon = isPublic ? undefined : "ban";
   return (
-    <Item key={index} className="mobile only">
+    <Item key={index} className="mobile only community-item">
```
alejandromumo, line 21: This does not seem to be used by `app-rdm`. There is an implementation there and that's the one being used.
However, I fixed it in both places.

kpsherva (approved):

PR #1797, opened 2022-07-08
Dashboard: fix mobile pagination [DEPENDS ON https://github.com/inveniosoftware/invenio-search-ui/pull/150]

Pineirin:
Screenshot from 2022-07-19 10-29-26
Screenshot from 2022-07-19 10-29-44

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/base.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/base.js
@@ -10,17 +10,11 @@
 import {
   SearchAppFacets,
   SearchAppResultsPane,
+  InvenioSearchPagination,
 } from "@js/invenio_search_ui/components";
 import { i18next } from "@translations/invenio_app_rdm/i18next";
 import React from "react";
-import {
-  Count,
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/base.js
```diff
@@ -78,8 +78,8 @@ export function DashboardResultView(props) {
         </Grid.Row>
         <Overridable id="DashboardResultView.resultFooter" {...props}>
           <Grid.Row verticalAlign="middle">
-            <Grid.Column width={4} />
-            <Grid.Column width={8} textAlign="center" floated="right">
+            <Grid.Column className="computer tablet only" width={4} />
```
NRodriguezcuellar, line 81: same comment as in https://github.com/inveniosoftware/invenio-communities/pull/762 here but if that is not the way then ignore this, otherwise looks good!

kpsherva: I think it is a very good point, I would say the invenio-search-ui would be the module to place it in

kpsherva (approved):

PR #1811, opened 2022-07-22
deposit form: add fieldPath prop to field components

NRodriguezcuellar:
Now explicitly adds the fieldPath prop to field components to move away from relying on defaultProps values.

related to https://github.com/inveniosoftware/react-invenio-deposit/pull/551


Pineirin (approved):
Is this the only file that gives problems with the defaultProps?

kpsherva (approved):

Issue #1204, opened 2022-02-02
Ability to provide custom text in the publication confirmation modal


max-moser:
## Status Quo

Currently, the deposit form gives warnings to the user that they can't change the set of uploaded files after publishing the record anymore:

![image](https://user-images.githubusercontent.com/6437519/152130669-b8db2276-868b-4580-8fad-c20557d24f37.png)

![image](https://user-images.githubusercontent.com/6437519/152130627-e3644ece-d260-43d3-9ad8-b6e61101e30f.png)

However, these warnings are very brief and could be missed too easily.
They should be reworded to drive the fact home that the users aren't able to change the set of files after publication anymore, along with a very short description as to why; similar to Zenodo.

Bonus points if we could customize the warning text, to reference our own sandbox system (also similar to Zenodo).

## Comparison: Zenodo

On Zenodo, the warnings related to not being able to change the files after publication are more elaborate:

![image](https://user-images.githubusercontent.com/6437519/152130710-89a70f32-caf4-4473-89d0-d980a783f1a9.png)

![image](https://user-images.githubusercontent.com/6437519/152130742-85b0e536-9ced-42ac-a94d-067e1b5205b2.png)

lnielsen:
Agree:

1. Instances should be able to provide their own text.
2. Should take into account publishing a new file and just editing of a record (where the files are already locked)

lnielsen:
From #1094 

It should be possible to provide also e.g. deposit agreement text in the publish dialog.

> We will be using the authentication page to provide some terms and conditions, but librarians at our institution also want to be able to put our deposit agreement in the publication pop-up confirmation modal.

> A simple config that is injected in the confirmation modal would do.

PR #1832, opened 2022-08-10
Allow customization of publish modal

fenekku:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1204
- depends on: https://github.com/inveniosoftware/react-invenio-deposit/pull/568 .
- Goes about it by just adding extra content to the modal. Seems like the simplest.

**Example Screenshot**
deposit_agreement_modal_3_implemented

In the community review

deposit_agreement_modal_community_submission

Serviceable.

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -636,6 +636,9 @@ APP_RDM_DEPOSIT_FORM_QUOTA = {
 APP_RDM_DISPLAY_DECIMAL_FILE_SIZES = True
 """Display the file sizes in powers of 1000 (KB, ...) or 1024 (KiB, ...)."""
 
+APP_RDM_DEPOSIT_FORM_PUBLISH_MODAL_EXTRA = ""
+"""Additional text/html to be displayed in the publish and submit for review modal."""
+
 RDM_CITATION_STYLES = [
     ("apa", _("APA")),
     ("harvard-cite-them-right", _("Harvard")),
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -302,6 +302,9 @@ def get_form_config(**kwargs):
             ]
         ),
         custom_fields=load_custom_fields(),
+        publish_modal_extra=current_app.config.get(
+            "APP_RDM_DEPOSIT_FORM_PUBLISH_MODAL_EXTRA"
+        ),
         **kwargs,
     )
 


invenio_app_rdm/config.py
```diff
@@ -631,6 +631,9 @@ def files_rest_permission_factory(obj, action):
 APP_RDM_DISPLAY_DECIMAL_FILE_SIZES = True
 """Display the file sizes in powers of 1000 (KB, ...) or 1024 (KiB, ...)."""
 
+APP_RDM_DEPOSIT_FORM_PUBLISH_MODAL_EXTRA = ""
```
kpsherva, line 634: I would propose to use react-overridable instead of config variable? I think it might be a bit overkill to store the UI panel text in the config, we will end up with an enormous config if we continue down this road

kpsherva: https://github.com/indico/react-overridable
it allows you to override any react component (the ones wrapped in the HOC the library provides), in this case it could be the modal. 
Thanks to this, we will prevent storing HTML in the config variables and increasing the config size
There are several examples of usage in the JS code, the most in invenio-communities and invenio-search-ui

fenekku: Hmmm, I don't understand how an instance can use react-overridable. Can you expand on that?

The 2 examples of react-overridable I looked at are:

1- Invenio-Requests: invenio-requests/invenio_requests/assets/semantic-ui/js/invenio_requests/requestsAppInit.js
2- Invenio-Communities: invenio-communities/invenio_communities/assets/semantic-ui/js/invenio_communities/requests/index.js

In 1-, the overriding components are imported and passed

```js
const overriddenComponents = {
  "TimelineEvent.layout.unknown": TimelineUnknownEvent,
  // ...
}

ReactDOM.render(
  <InvenioRequestsApp
    request={request}
    defaultQueryParams={defaultQueryParams}
    overriddenCmps={overriddenComponents}
    userAvatar={userAvatar}
  />,
  requestDetailsDiv
);
```

. `InvenioRequestsApp` then renders with `<OverridableContext.Provider value={overriddenCmps}>`.

The equivalent for the deposit form would be to pass the overriding components in
invenio-app-rdm/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/index.js:

```js
ReactDOM.render(
  <RDMDepositForm
    record={getInputFromDOM("deposits-record")}
    preselectedCommunity={getInputFromDOM("deposits-draft-community")}
    files={getInputFromDOM("deposits-record-files")}
    config={getInputFromDOM("deposits-config")}
    permissions={getInputFromDOM("deposits-record-permissions")}
    overriddenCmps={overriddenComponents}
  />,
  document.getElementById("deposit-form")
);
```

. `RDMDepositForm` could then be made to render with `<OverridableContext.Provider value={overriddenCmps}>`
and have the `PublishButton` (the real component we want to override) overridden.

The problem is how to fill `overriddenComponents` with components from my instance? The above works because the `render`
is under control of the module. It's not the case for instances.

In 2-, the setup is similar but of note is that it uses the

```js
parametrize(RequestDeclineButton, {
  size: "mini",
  className: "mt-10 fluid-responsive",
});
```

approach to create the overriding components first. It then does the same passing of overridden components
but via the `createSearchAppInit`. The use of the `parametrize` would be a seam for an instance to pass their
`publishExtra` there via config, but we would be back to a `config.py` value which you don't want.

So the fundamental problem is: how does an instance provide components to the deposit form for overriding?
If such a mechanism needs to be created, then my vote would be to go with the current PR first and make
a ticket to work on this mechanism later.

zzacharo: @fenekku what you commented above is what it could be done to properly solve the problem of UI customization as currently is quite scattered around in different config variables. That said, there is still some mechanism missing as you noticed to connect the pieces together. I would probably go as you suggested. Go ahead with this solution for now (there are also other config variables doing the same) and create a ticket with what you described above so we can attack this all at once i.e implement the overriding mechanism and modify the existing python config variables accordingly.

fenekku: That's good to hear. I was thinking of a registration system but didn't know how to do that with our javascript setup :sweat_smile: . I will create a ticket.

fenekku:
Got the go ahead from @zzacharo so merging.

Issue #1845, opened 2022-08-19
custom-fields: display fields in record's landing page


zzacharo:
We need to display the user's filled in custom fields in the record's landing page.

**From RFC** 

![Image](https://user-images.githubusercontent.com/22594783/185568689-d68dc828-1e6e-4af9-815f-7cae1940f99a.png)

The proposed mockup is to display the fields at the bottom of the page in the "Additional details" section. Each configured section will be displayed as a different tab.

**UX**

- [x] We should test in case of a bigger amount of configured sections how the UI design adapts.
- [x] We should test that the widgets allowing html i.e RichInput are correctly displayed and sanitized in the details page

PR #1861, opened 2022-09-06
Add custom fields landing page

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1845

# Screenshot
Screenshot 2022-09-06 at 14 46 34


invenio_app_rdm/records_ui/views/records.py
```diff
@@ -71,13 +71,27 @@ def open(self):
 #
 # Views
 #
+def load_custom_fields(conf_ui, conf_backend):
```
ppanero, line 74: shelvable: we had a similar loader somewhere else for the react part? I wonder if it could be reused/made generic.

ppanero (approved):

PR #1874, opened 2022-09-15
migration: add script to migrate to v10

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1037


ntarocco (approved):
Hard for me to review this one, I am totally out of context. If tested, it is ok with me.
Added @ppanero in case they have to add any upgrade recipe for `custom fields`

ppanero:
In terms of custom fields there is no DB change needed. We will need:
- some cli commands to be run
- creation of a folder in assets/templates

But those will go to the docs.

(Not reviewing cuz Im also OOC in OAI sets...)

PR #1879, opened 2022-09-16
Administration: add admin link in the profile menu

Pineirin:
* closes: inveniosoftware/invenio-administration#66

Admin:
Screenshot from 2022-09-16 19-47-54
Screenshot from 2022-09-16 19-48-45
Screenshot from 2022-09-19 10-49-06

Not admin:
Screenshot from 2022-09-19 10-51-59
Screenshot from 2022-09-19 10-52-12

### DIFF
--- a/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/header_login.html
+++ b/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/header_login.html
@@ -76,7 +76,19 @@
                 </a>
               {%- endfor %}
 
-              <div class="divider"></div>
+              <div class="ui divider"></div>
+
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/templates/semantic-ui/invenio_theme/header_login.html
```diff
@@ -0,0 +1,23 @@
+{# -*- coding: utf-8 -*-
```
kpsherva, line 1: this file will not be used if placed here in invenio_app_rdm becuase the application is using the one above, you should move it to invenio_administration instead (so that invenio_administration can be installed in other application and have the menu added out of the box, without overwriting the header)

kpsherva (approved):

kpsherva (approved):

PR #1910, opened 2022-10-03
dependencies: limit importlib-metadata due to broken celery imports

ppanero:


ppanero:
pinned in invenio-celery

PR #1935, opened 2022-10-12
Ignore `FieldPermissionError`s in `validate_draft` for the draft preview page

max-moser:
The reasoning behind this PR is outlined in https://github.com/inveniosoftware/invenio-drafts-resources/pull/207, which is also required for this PR to work.


slint (approved):
Agree with the changes to make this work in general, but disagree with the solution for the [original bug report](https://github.com/inveniosoftware/invenio-app-rdm/issues/1051). I think we should open an issue to find a better solution, since now this feels a bit awkward and has introduced a very weird logic... 

Can you add here a comment linking to the issue explaining why we're even calling `.validate_draft(...)`?

max-moser:
added a comment in the code and created a follow-up issue

slint (approved):

Issue #1937, opened 2022-10-13
add robots.txt


Pineirin:
We should add a robots.txt for

- [ ] CDS
- [ ] Zenodo
- [ ] WHO
- [ ] OAIS

slint:
Regarding adding specialized `robots.txt` for all the different instances, I don't think we have differences since we're using the same default InvenioRDM views (i.e. WHO doesn't have a requirement to change the search page URL).

PR #1938, opened 2022-10-13
Robots: add robots.txt

Pineirin:
closes: #1937

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -535,6 +535,7 @@ APP_RDM_USER_DASHBOARD_ROUTES = {
 
 APP_RDM_ROUTES = {
     "index": "/",
+    "robots": "/robots.txt",
     "help_search": "/help/search",
     "record_search": "/search",
     "record_detail": "/records/<pid_value>",
--- a/invenio_app_rdm/theme/views.py
+++ b/invenio_app_rdm/theme/views.py
@@ -30,6 +30,7 @@ def create_blueprint(app):
     )
 
     blueprint.add_url_rule(routes["index"], view_func=index)
+    blueprint.add_url_rule(routes["robots"], view_func=robots)
     blueprint.add_url_rule(routes["help_search"], view_func=help_search)
 
     @blueprint.before_app_first_request
@@ -61,6 +62,11 @@ def index():
     return render_template(
         current_app.config["THEME_FRONTPAGE_TEMPLATE"],
     )
+
+
+def robots():
+    """Robots.txt."""
+    return current_app.send_static_file("robots.txt")
 
 
 def help_search():
--- a/tests/ui/test_static.py
+++ b/tests/ui/test_static.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-RDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+
+def test_static(app, client, es_clear):
+    """Check if robots.txt returns a string response."""
+
+    # Save the current static folder to revert after finishing the test.
+    app_static_folder = app.static_folder
+    # Use as static folder the one of the blueprint we are testing.
+    app.static_folder = app.blueprints["invenio_app_rdm"].static_folder
+
+    response = client.get("/robots.txt")
+
+    assert response.status_code == 200
+    assert type(response.text) == str
+    assert "Disallow: /search" in response.text
+    assert "Disallow: /api" in response.text
+    assert "Disallow: /administration" in response.text
+
+    app.static_folder = app_static_folder
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -48,6 +48,7 @@ recursive-include invenio_app_rdm *.html *.js *.less *.json
 recursive-include invenio_app_rdm *.overrides
 recursive-include invenio_app_rdm *.variables
 recursive-include invenio_app_rdm *.png *.svg
+recursive-include invenio_app_rdm *.txt
 recursive-include invenio_app_rdm *.po *.pot *.mo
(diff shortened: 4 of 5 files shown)


invenio_app_rdm/config.py
```diff
@@ -746,3 +747,11 @@ def files_rest_permission_factory(obj, action):
     "tif": "tiff",
     "tiff": "tiff",
 }
+
+ROBOTS_TXT = """\
```
kpsherva, line 751: was adding a new config variable consulted with anyone ? did we consider any other implementation options?

invenio_app_rdm/config.py
```diff
@@ -746,3 +747,11 @@ def files_rest_permission_factory(obj, action):
...
+    User-agent: *
+    Disallow: /search
+    Disallow: /api
+    Disallow: /admin/
```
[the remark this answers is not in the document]
ppanero: +1

invenio_app_rdm/theme/views.py
```diff
@@ -63,6 +64,11 @@ def index():
     )
 
 
+def robots():
```
ppanero, line 67: What is the advantage of this approach vs having a static file in the cookiecutter repo? For simple configurations a config variable is okay. However, could this config become large enough to be moved to a separate file? if so, it would not be supported?

For the Describo profile, we also need to serve a static file (which is similar to this - same outcome, different content -). IIRC, it was placed in `static` and served via Nginx. This way, we also **offload the serving of the file** to Nginx and avoid passing these requests through the application. CC @alejandromumo for more details.

invenio_app_rdm/config.py
Pineirin: Other potential solutions that came to my mind are:
- Serve a robots.txt file (instead of answering the request with text).
- Define the robots .txt in the deploy configuration (I don't know properly how the deploy configuration works right now in Zenodo, or I have access).
This solution was considered the best after talking with @alexdutton.

invenio_app_rdm/theme/views.py
kpsherva: same suggestion from me, https://github.com/inveniosoftware/invenio-app-rdm/pull/1938#discussion_r995498843 I think we should go with simple static file

slint: The reason for having this proposal via config (vs. static file via `nginx`) is mainly deployment complexity. We don't have any similar file in cookiecutter (besides static assets) that get served via `nginx`. This means we have to modify cookiecutter + `invenio-cli` + helm charts (?) for getting this in.

ppanero: I do agree that maybe there are some more implications in terms of deployment. But changes should not be too big, we already have a mount for the [invenio assets](https://github.com/inveniosoftware/helm-invenio/blob/30ec941b6b8ebd60472787cd7f08495e9c0bc7de/invenio/templates/deployments/web.yaml#L118) that need to be served. I guess we can profit from this (or follow the same approach). In any case, we will need to do this work to serve the _describo_ profile?

To make a more informed decission. Today, how many requests are done to the robots.txt in Zenodo or CDS? if there ar every few I guess passing it by the application would be okay.

slint: From Zenodo ~2 weeks of access logs we have ~3.5k requests to `/robots.txt` per day (2.5 req/min). IMHO that's low in terms of load, so adding the complexity of serving via nginx seemed overkill. I understand the thought process behind it, since `robots.txt` is a bit special and lies somewhere between deployment and business logic (i.e. "what pages/content of my instance do I want Google to index?").

For Describo, we'll rely on the existing mechanism of the cookiecutter `static` folder symlinking, i.e. there's not new top-level endpoint added, just a new `/static/describo` folder.

slint (commented):
LGTM, maybe add a test if there's something equivalent already for the `help_search` view.

zzacharo (approved):

invenio_app_rdm/theme/views.py
```diff
@@ -30,6 +30,7 @@ def create_blueprint(app):
     )
 
     blueprint.add_url_rule(routes["index"], view_func=index)
+    blueprint.add_url_rule(routes["robots"], view_func=robots)
```
zzacharo, line 33: Can we register this only if we are in development mode i.e `Flask_ENV=devleopment` or `DEBUG=True`? For production/containerized versions we need to implement a rewrite rule on nginx level and use the static file.

slint: I think the nginx config will take precedence, so it won't reach the Python app ever.

zzacharo: we said maybe to disable that in production setup, just to make sure that this is not available and if users want then they configure nginx... We have a ticket already to add the nginx rule in all our deployment related instances...Do you find any caveat having the above only in development?

slint: It's more in the sense that e.g. `/static/` is not disabled in production mode, so if something was wrong in the `nginx` config we wouldn't get a warning/error.

In general, we don't have many places where we disable/enable features based on `debug`, and specifically for `/robots.txt` the impact is not that big (e.g. on Zenodo we had about 2-3req/min to the endpoint).

tests/ui/test_static.py
```diff
@@ -0,0 +1,22 @@
...
+
+    # Save the current static folder to revert after finishing the test.
+    app_static_folder = app.static_folder
+    # Use as static folder the one of the blueprint we are testing.
+    app.static_folder = app.blueprints["invenio_app_rdm"].static_folder
+
+    response = client.get("/robots.txt")
+
+    assert response.status_code == 200
+    assert type(response.text) == str
```
ntarocco, line 20: Small suggestion:
```suggestion
    assert type(response.text) == str
    assert "Disallow: /search" in response.text
    assert "Disallow: /api" in response.text
    assert "Disallow: /administration" in response.text
```

ntarocco (approved):

Issue #1934, opened 2022-10-12
static pages: integrate invenio-pages


zzacharo:
We should:
- Use invenio-pages
- Evaluate which dependencies need to be updated so we can use them in the RDM ecosystem.
- Define a mechanism to load the static pages as fixtures
- We can skip the editing of the pages at the moment.

PR #1944, opened 2022-10-19
staticPages: add static pages

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1934

image

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -44,7 +44,7 @@ recursive-include docs *.rst
 recursive-include docs *.txt
 recursive-include docs Makefile
 recursive-include invenio_app_rdm *.gitkeep
-recursive-include invenio_app_rdm *.html *.js *.less *.json
+recursive-include invenio_app_rdm *.html *.js *.less *.json *.yaml
 recursive-include invenio_app_rdm *.overrides
 recursive-include invenio_app_rdm *.variables
 recursive-include invenio_app_rdm *.png *.svg
--- a/invenio_app_rdm/cli.py
+++ b/invenio_app_rdm/cli.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-App-RDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Command-line tools for invenio app rdm."""
+import click
+from flask.cli import with_appcontext
+
+from invenio_app_rdm.fixtures import Pages
+
+
+@click.group()
+def rdm():
+    """Invenio app rdm commands."""
+
+
+@rdm.group()
+def pages():
+    """Static pages."""
+
+
+@pages.command("create")
+@click.option(
+    "-f",
+    "--force",
+    "force",
+    default=False,
+    is_flag=True,
+    show_default=True,
+    help="Creates static pages.",
+)
+@with_appcontext
+def create_static_pages(force):
+    """Create static pages."""
+    click.secho("Creating static pages...", fg="green")
+
+    Pages().run(force=force)
+
+    click.secho("Created static pages!", fg="green")
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -746,3 +746,11 @@ IIIF_FORMATS_PIL_MAP = {
     "tif": "tiff",
     "tiff": "tiff",
 }
+
+PAGES_DEFAULT_TEMPLATE = "invenio_app_rdm/default_static_page.html"
+"""Default template to render."""
+
+PAGES_TEMPLATES = [
+    ("invenio_app_rdm/default_static_page.html", "Default"),
+]
+"""List of available templates for pages."""
--- a/invenio_app_rdm/fixtures/__init__.py
+++ b/invenio_app_rdm/fixtures/__init__.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio App RDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Static pages."""
+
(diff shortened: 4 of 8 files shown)


invenio_app_rdm/static_pages/static_pages.py
```diff
@@ -0,0 +1,67 @@
...
+        return (
+            Path(Path(__file__).parent.as_posix() + os.path.join("/templates", page))
+            .read_bytes()
+            .decode("utf8")
+        )
+
+    def run(self, force=False):
+        """Run the creation of Static Pages."""
+        pages = [
+            Page(
```
zzacharo, line 42: **major**: I think as it is now the mechanism doesn't allow for an instance to define new static pages, right? I can see only that you override via template the predefined pages here.... Also, I am not sure we need to prepopulate static pages unless we convert some of the existing ones e.g help/searchguide to a static page....

invenio_app_rdm/static_pages/static_pages.py
```diff
@@ -0,0 +1,67 @@
...
+from invenio_db import db
+from invenio_pages import Page
+
+
+class StaticPages:
+    """Static pages."""
+
+    def page_data(self, page):
+        """Returns content of template."""
+        assets_path = current_webpack.project.path
```
slint, line 22: minor: I'm not sure if pages should go under assets/static. In this context, we're loading the pages similarly to fixtures, so [`app_data`](https://github.com/inveniosoftware/invenio-rdm-records/blob/fbc59a9ed17b6086930cadce077c07cb752f7481/invenio_rdm_records/fixtures/__init__.py#L40) might be the more "usual" place.

invenio_app_rdm/static_pages/static_pages.py
```diff
@@ -0,0 +1,67 @@
...
+"""Static pages."""
+import os
+from pathlib import Path
+
+from flask_webpackext import current_webpack
+from invenio_db import db
+from invenio_pages import Page
+
+
+class StaticPages:
```
slint, line 17: minor: we could maybe reuse the [`FixtureMixin`](https://github.com/inveniosoftware/invenio-rdm-records/blob/fbc59a9ed17b6086930cadce077c07cb752f7481/invenio_rdm_records/fixtures/fixture.py#L13).

Probably you'll need to override the `__init__` to pass the `force` flag, but otherwise, it could work out of the box by just defining the `create` method.

See also https://github.com/inveniosoftware/invenio-rdm-records/pull/1113, which loads communities, so that we can possibly align fixture loading in general

invenio_app_rdm/static_pages/static_pages.py
slint, line 42: minor: if we use the `FixtureMixin` described above, these entries could go in a `pages.yml` file, and the HTML file could be loaded from a `pages/` folder or similar.

invenio_app_rdm/static_pages/static_pages.py
```diff
@@ -0,0 +1,67 @@
...
+            ),
+            Page(
+                url="/contact",
+                title="Contact",
+                description="Contact",
+                content=self.page_data("contact.html"),
+                template_name="invenio_pages/default.html",
+            ),
+            Page(
+                url="/guide/search",
```
slint, line 57: minor: one feature we lose here is [the l10n](https://github.com/inveniosoftware/invenio-app-rdm/blob/3cff73b4e3726f33122176368213b662e77d7571/invenio_app_rdm/theme/views.py#L69-L72) of the help search guide, which could be achieved by dropping e.g. a `invenio_app_rdm/help/search.es.html` file to define the Spanish translation.

This might apply though in general to pages, so we might revisit in the future to see how one could provide other languages for a single page.

slint, line 57: minor: I think the search guide is probably the only "out-of-the-box" page we want to include by default. The rest of the pages include maybe too specific hardcoded information that anyways users will change for their instances.

invenio_app_rdm/static_pages/static_pages.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
```
slint, line 1: minor: in the context of the comments below, it might be better to name this `fixtures/pages.py` or similar.

slint (commented):
LGTM as a starting point. I commented some bigger suggestions/opinions on how to structure this, but it could be that I'm missing context on how we want to approach this.

(Since I'm AFK till Tuesday, take this discussion with the rest of the team).

invenio_app_rdm/static_pages/static_pages.py
```diff
@@ -0,0 +1,53 @@
...
+    def create(self, entry):
+        """Load a single page."""
+
+        db.session.add(
+            Page(
+                url=entry.pop("url"),
+                title=entry.get("title"),
+                description=entry.get("description"),
+                content=self.page_data(entry.get("template")),
+                template_name="invenio_pages/default.html",
```
zzacharo, line 50: I would probably use a new default template defined here. The reason is that the invenio-pages default is imposing some layouting... For example, I would have a:
```
{% extends 'invenio_pages/default_html' %}

{% set title = page.title %}

{% block page_body %}
{% block content %}{{ page.content|safe }}{% endblock %}
{% endblock %}
```

invenio_app_rdm/static_pages/static_pages.py
zzacharo: +1 to this!

invenio_app_rdm/static_pages/static_pages.py
jrcastro2: Yes! I already added it in the cookiecutter PR, you can check it [here](https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/235)

zzacharo: I think we can just omit the template from cookiecutter and define it here... Then, if someone needs to override it, they can do it in cookiecutter, wdyt?

zzacharo (approved):

Issue #1932, opened 2022-10-12
frontpage: add search boxes in main page


zzacharo:

jennur:
![Image](https://user-images.githubusercontent.com/21052053/195555185-00072bc9-36d1-4d47-a5e4-b4ee799a43b8.png)

slint:
After our discussions, I've written up everything in this mini-RFC: https://codimd.web.cern.ch/YlYa7084TbGbU2Vtm3j7qg

To summarize here:

- We define a default layout for the frontpage, organized in optionally overridable Jinja blocks
- We encapsulate complex components (communities carousel, top/recent records listing) into Jinja macros

PR #1946, opened 2022-10-21
front page: Create new layout and style introduction section

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1932

- Created layout blocks for easier override of the content on the front page
- Added `RecordsList` component
- Added config variable to disable intro-section
- Restyled the intro section
- Added the carousel macro from https://github.com/inveniosoftware/invenio-communities/pull/822, so this PR is needed here.

*Note:* I added some mock content in the sidebar just to show the layout, but I'm open for suggestions of what to add here. Maybe we should rather add some generic user-instructions on how to use rdm? Open for suggestions.


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/frontpage/records-list.js
```diff
@@ -0,0 +1,90 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2022 CERN.
+//
+// Invenio RDM is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { i18next } from "@translations/invenio_communities/i18next";
```
zzacharo, line 9: This should be changed to be imported from invenio-app-rdm

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/frontpage/index.js
```diff
@@ -0,0 +1,19 @@
...
+
+import React from "react";
+import ReactDOM from "react-dom";
+import RecordsList from "./records-list";
+
+const recordsListContainer = document.getElementById("records-list");
+const title = recordsListContainer.dataset.title;
+const fetchUrl = recordsListContainer.dataset.fetchUrl;
+
+recordsListContainer &&
```
zzacharo, line 15: I think the check is not needed!

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/intro_section.html
```diff
@@ -0,0 +1,93 @@
...
+      <div class="image">
+        <div class="featured-list">1</div>
+      </div>
+      <div class="content">
+        <div class="header">
+          {{ _("Configure") }}
+        </div>
+        <div class="description">
+          {% trans %}
+            Configure the datamodel, resource types, permissions and much more to make it suit your needs!
```
zzacharo, line 27: Are these changes in text needed? We are changing the translations too and it will interfere with the current translation process that is undergoing for V10.1...

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/frontpage.html
```diff
@@ -6,70 +6,61 @@
...
+        <aside class="five wide computer sixteen wide tablet column">
+
+          {% block side_column_content %}
+            <section class="ui segment rdm-sidebar">
+              <h2 class="ui small header">Lorem ipsum dolor sit amet</h2>
+              <p>
+                Consectetur adipiscing elit. Fusce posuere, est et maximus molestie, augue ante sollicitudin leo, in venenatis nibh lorem nec dolor.
+                <br/>
+                <br/>
+                Vestibulum facilisis justo et felis scelerisque consectetur. Proin congue nisi vel ex interdum interdum. In eget dignissim dolor, eget semper arcu. Morbi a velit sit amet erat rhoncus porttitor. Pellentesque leo felis, accumsan ac ligula ac, interdum efficitur sem. Nam cursus ante dolor, at sodales risus mattis id.
```
zzacharo, line 53: We can leave that empty or just the layout without content if not needed...

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/frontpage.html
```diff
@@ -6,70 +6,61 @@
...
-                target="_blank">{{ _("Code") }}</a>
-          </div>
+  {%- block intro_section %}
+    {% if show_intro_section %}
+      {% include "invenio_app_rdm/intro_section.html" %}
+    {% endif %}
+  {%- endblock intro_section %}
+
+  {%- block top_banner %}
+    {{ communities_carousel() }}
```
zzacharo, line 27: I don't see the featured communities in the screenshot...is this because communities list is empty? What happens in that case?

jennur: Yes, on install there are no communities, so the component will be empty.

invenio_app_rdm/config.py
```diff
@@ -174,6 +174,9 @@ def _(x):
 THEME_SITENAME = _("InvenioRDM")
 """Site name."""
 
+THEME_SHOW_FRONTPAGE_INTRO_SECTION = True
```
kpsherva, line 177: why do we need this config variable?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/frontpage/records-list.js
```diff
@@ -0,0 +1,90 @@
...
+
+  renderItems() {
+    const { data } = this.state;
+
+    return data.hits.length
+      ? data.hits.map((record) => {
+          return <RDMRecordResultsListItem result={record} key={record.id} />;
+        })
+      : "";
+  }
```
kpsherva, line 61:
```suggestion
    return data.hits?.map((record) => 
         <RDMRecordResultsListItem result={record} key={record.id} />;
      );
  }
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/frontpage/records-list.js
```diff
@@ -0,0 +1,90 @@
...
+
+    try {
+      const response = await this.cancellableFetch.promise;
+      this.setState({ data: response.data.hits, isLoading: false });
+    } catch (error) {
+      console.log(error);
+    }
+  };
+
+  renderItems() {
```
kpsherva, line 53: this part does not need to be moved to a separate method, can be done inside the render function

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -167,14 +167,13 @@ pre.export.result {
...
+  font-weight: 500;
+  font-size: 6em;
   color: white;
   background-color: @featuredListBackgroundColor;
   box-sizing: border-box;
-  width: 100px;
-  height: 100px;
-  border-radius: 50%;
+  width: 100%;
+  height: 150px;
```
kpsherva, line 176: do we need variables for easier customisation? Also font attributes above

invenio_app_rdm/config.py
jennur: Discussed with @lnielsen to have this config variable to easily disable the intro-section

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
jennur: I don't think anyone will override this part. It is only used in the intro-section.

kpsherva: in this case maybe it should have specific class name, otherwise I understand it is referring to featured comms or records, and we will override these clasess (discussed via chat)

kpsherva (approved):

PR #1967, opened 2022-11-15
deposit form: add configs for metadata-only and full record visibilit…

jennur:
Closes https://github.com/zenodo/zenodo-rdm/issues/46

- Adds two new config variables
    - `APP_RDM_DEPOSIT_CAN_HAVE_METADATA_ONLY`
    - `APP_RDM_DEPOSIT_FULL_RECORD_VISIBILITY`

Needs PR https://github.com/inveniosoftware/react-invenio-deposit/pull/597


invenio_app_rdm/config.py
```diff
@@ -652,6 +652,12 @@ def files_rest_permission_factory(obj, action):
 APP_RDM_DEPOSIT_FORM_PUBLISH_MODAL_EXTRA = ""
 """Additional text/html to be displayed in the publish and submit for review modal."""
 
+APP_RDM_DEPOSIT_CAN_HAVE_METADATA_ONLY = True
```
zzacharo, line 655: What about naming it in a way that explicitly depicts what this variable does i.e 
```suggestion
APP_RDM_DEPOSIT_DISABLE_METADATA_ONLY_REORDS = False
```
or 
```suggestion
APP_RDM_DEPOSIT_REQUIRE_FILES = True
```
@slint @jennur wdyt?

invenio_app_rdm/config.py
```diff
@@ -652,6 +652,12 @@ def files_rest_permission_factory(obj, action):
...
+"""Enable/disable metadata-only records"""
+
+APP_RDM_DEPOSIT_FULL_RECORD_VISIBILITY = True
+"""Enable/disable custom visibility of the full record"""
```
zzacharo, line 659:
```suggestion
"""Enable/disable record visibility section in the upload form."""
```

jennur: Thanks 😄

invenio_app_rdm/config.py
zzacharo: Ok let' go with that one then! Can you change the config? I think it should default to `False` so it shows the metadata only section!

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/deposit.html
```diff
@@ -16,19 +16,19 @@
 
 {%- block page_body %}
   {%- if record %}
-  <input id="deposits-record" type="hidden" name="deposits-record" value='{{record | tojson }}'></input>
+  <input id="deposits-record" type="hidden" name="deposits-record" value='{{record | tojson }}'/>
```
kpsherva, line 19: isn't it JSX syntax?
```suggestion
  <input id="deposits-record" type="hidden" name="deposits-record" value='{{record | tojson }}'>
```
I think input in pure HTML does not have the end `/`
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

invenio_app_rdm/config.py
```diff
@@ -652,6 +652,12 @@ def files_rest_permission_factory(obj, action):
 APP_RDM_DEPOSIT_FORM_PUBLISH_MODAL_EXTRA = ""
 """Additional text/html to be displayed in the publish and submit for review modal."""
 
+APP_RDM_DEPOSIT_REQUIRE_FILES = False
+"""Enable/disable metadata-only records"""
+
+APP_RDM_DEPOSIT_FULL_RECORD_VISIBILITY = True
```
kpsherva, line 658: what about the backend validation? it is still possible to pass the access via REST API

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/deposit.html
jennur: Apparently it doesn't matter if it's there or not, I just prefer it for the readability... although I don't have very strong opinions about it, so LMK if it should be removed.

jennur:
Closed because we decided on a different solution (https://github.com/inveniosoftware/invenio-app-rdm/pull/1970/)

PR #1969, opened 2022-11-18
CLI command for rebuilding the search indices

max-moser:
This PR introduces a new CLI command for calling `service.rebuild_index()` for every service with this method registered in the service registry (which should include all relevant services in the latest releases).

### DIFF
--- a/invenio_app_rdm/cli.py
+++ b/invenio_app_rdm/cli.py
@@ -8,6 +8,8 @@
 """Command-line tools for invenio app rdm."""
 import click
 from flask.cli import with_appcontext
+from invenio_access.permissions import system_identity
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/cli.py
```diff
@@ -40,3 +42,20 @@ def create_static_pages(force):
     Pages().run(force=force)
 
     click.secho("Created static pages!", fg="green")
+
+
+@rdm.command("rebuild-indices")
```
kpsherva, line 47:
```suggestion
@rdm.command("rebuild-all-indices")
```
to avoid name confusion with existing rebuild-index command

max-moser:
both renamed

PR #1971, opened 2022-11-18
RecordsResultsListItem: refactor and add overridable

Pineirin:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/8

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/frontpage/override.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/frontpage/override.js
@@ -1,0 +1,7 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2022 CERN.
+//
+// Invenio App RDM is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+export const overriddenComponents = {};
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
@@ -25,108 +25,16 @@ import {
   Header,
   Icon,
   Input,
-  Item,
   Label,
   Message,
   Segment,
 } from "semantic-ui-react";
-import { SearchItemCreators } from "../utils";
+import RecordsResultsListItem from "../components/RecordsResultsListItem";
 import PropTypes from "prop-types";
-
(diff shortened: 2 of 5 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/search/components.js
```diff
@@ -25,80 +25,15 @@ import {
...
-              <small>
-                {i18next.t("Uploaded on")} <span>{createdDate}</span>
-              </small>
-            </div>
-          )}
-        </Item.Extra>
-      </Item.Content>
-    </Item>
-  );
+  return <RecordsResultsListItem result={result} />;
```
kpsherva, line 36: why do we need this wrapping component?

Pineirin: My rationale behind this change was the following:

Initially we were using a functional component called RDMRecordsResultsListItem for both the results list in the frontpage and our custom RDM list of records in the search. For the frontpage, we were just using the RDMRecordsResultsListComponent for each record, while for the search we were using that component to override the ResultsList.item component.

I wanted to make some differentiation between the two use cases, as they may vary so, as part of the refactor I was doing to make RDMRecordsResultsListItem a React component (instead of a functional component), I divided it into two components. One component  is RecordsResultsListItem (https://github.com/inveniosoftware/invenio-app-rdm/pull/1971/files#diff-3c47a1ea3137a52366750ec052a4d5b1e6c685a8c5e441f851b95c7b00f24bf0) which is used by the recent uploads section, the other is RDMRecordsResultsListItem(invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/frontpage/RecordsList.js) which is used for the override of the uploads search page. This second component wraps the first one to avoid code duplciation because, in our RDM instance, they are equal.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/frontpage/override.js
```diff
@@ -0,0 +1,7 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2022 CERN.
+//
+// Invenio App RDM is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+export const overriddenComponents = {};
```
jennur, line 7: Instead of creating another wrapper for `RecordsListItem`, can we not just use the default `ListItem` from `react-searchkit` in the frontpage `RecordsList`, and then override it with the original `RDMRecordResultsListItem` in this file, and that way keep it as it was in the search components?

PR #1975, opened 2022-11-22
Migrated checksum and integrity report tasks

### LINKED ISSUE #1951: Backend: File checksum verification + integrity reports

alejandromumo:
closes #1951

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -288,6 +288,22 @@ CELERY_BEAT_SCHEDULE = {
         "task": "invenio_requests.tasks.check_expired_requests",
         "schedule": crontab(minute=3, hour=0),
     },
+    "file-checks": {
+        "task": "invenio_files_rest.tasks.schedule_checksum_verification",
+        "schedule": timedelta(hours=1),
+        "kwargs": {
+            "batch_interval": {"hours": 1},
+            "frequency": {"days": 14},
+            "max_count": 0,
+            # Query taking into account only files with URI prefixes defined by
+            # the FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES config variable
+            "files_query": "invenio_app_rdm.utils.files.checksum_verification_files_query",
+        },
+    },
+    "file-integrity-report": {
+        "task": "invenio_app_rdm.tasks.file_integrity_report",
+        "schedule": crontab(minute=0, hour=7),  # Every day at 07:00 UTC
+    },
 }
 """Scheduled tasks configuration (aka cronjobs)."""
 
@@ -772,3 +788,20 @@ PAGES_TEMPLATES = [
     ("invenio_app_rdm/default_static_page.html", "Default"),
 ]
 """List of available templates for pages."""
+
+FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES = []
+"""URI prefixes of files their checksums should be verified"""
+
+FILES_INTEGRITY_REPORT_TEMPLATE = (
+    "invenio_app_rdm/files_integrity_report/email/files_integrity_report.html"
+)
+"""Files integrity report template"""
+
+FILES_INTEGRITY_REPORT_SUBJECT = "Files integrity report"
+"""Files integrity report subject"""
+
+ADMIN_EMAIL_RECIPIENT = "info@inveniosoftware.org"
+"""Admin e-mail"""
+
+MAIL_DEFAULT_SENDER = "info@inveniosoftware.org"
+"""Default e-mail address sender."""
--- a/invenio_app_rdm/tasks.py
+++ b/invenio_app_rdm/tasks.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio App RDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+"""Celery tasks for invenio-app-rdm."""
+
+import sqlalchemy as sa
+from celery import shared_task
+from invenio_db import db
+from invenio_files_rest.models import FileInstance
+
+from .utils.files import send_integrity_report_email
+
+
+@shared_task()
+def file_integrity_report():
+    """Send a report of uhealthy/missing files to system admins."""
+    # First retry verifying files that errored during their last check
+    files = FileInstance.query.filter(FileInstance.last_check.is_(None))
+    for f in files:
+        try:
+            f.clear_last_check()
+            db.session.commit()
+            f.verify_checksum(throws=False)
+            db.session.commit()
+        except Exception:
+            pass  # Don't fail sending the report in case of some file error
+
+    unhealthy_files = (
+        FileInstance.query.filter(
+            sa.or_(
+                FileInstance.last_check.is_(None), FileInstance.last_check.is_(False)
+            )
+        )
+        .order_by(FileInstance.created.desc())
+        .all()
+    )
+
+    send_integrity_report_email(unhealthy_files)
--- a/invenio_app_rdm/utils/files.py
+++ b/invenio_app_rdm/utils/files.py
@@ -1,0 +1,120 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio App RDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Files utilities."""
+
+from __future__ import absolute_import, print_function
+
+from datetime import datetime
+
+import sqlalchemy as sa
+from flask import current_app
+from flask_mail import Message
+from invenio_files_rest.models import FileInstance
+from invenio_rdm_records.records.models import RDMDraftMetadata, RDMRecordMetadata
+from pytz import utc
+
+
+def checksum_verification_files_query():
+    """Return a FileInstance query taking into account file URI prefixes."""
+    files = FileInstance.query
+    uri_prefixes = current_app.config.get(
+        "FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES"
+    )
+    if uri_prefixes:
+        files = files.filter(
+            sa.or_(*[FileInstance.uri.startswith(p) for p in uri_prefixes])
+        )
+    return files
+
+
+def render_email_from_context(template, context):
+    """Renders a template using a provided context."""
+    template = current_app.jinja_env.get_or_select_template(template)
+    return template.render(context)
+
+
+def generate_integrity_report_entries(files):
+    """Generates the report entries."""
+    entries = []
+    for file in files:
+        entry = _generate_report_entry(file)
+        entries.append(entry)
+    return entries
+
+
+def _generate_report_entry(file):
+    """Generates a single report for a file."""
+    entry = {"file": file}
+    for o in file.objects:
+        entry["filename"] = o.key
+
+        # Find records/drafts for the files
+        draft = get_draft_from_bucket(o.bucket_id)
+        if draft and draft.json:
+            entry["draft"] = draft.json
+        else:
+            record = get_record_from_bucket(o.bucket_id)
+            if record and record.json:
+                entry["record"] = record.json
+    return entry
+
+
+def get_record_from_bucket(bucket_id):
+    """Retrieve a record from a bucket id."""
+    return RDMRecordMetadata.query.filter_by(bucket_id=bucket_id).one_or_none()
+
+
+def get_draft_from_bucket(bucket_id):
+    """Retrieve a draft from a bucket id."""
+    return RDMDraftMetadata.query.filter_by(bucket_id=bucket_id).one_or_none()
+
+
+def get_report_subject():
+    """Retrieves the integrity report e-mail's subject."""
+    subject = current_app.config.get(
+        "FILES_INTEGRITY_REPORT_SUBJECT", "Files integrity report"
+    )
+    now = datetime.now(tz=utc).strftime("%Y-%m-%d %H:%M:%S")
+    subject += f" [{now}]"
+    return subject
+
+
+def get_report_template():
+    """Retrieves the template name to be used in the files integrity report."""
+    tpl = current_app.config["FILES_INTEGRITY_REPORT_TEMPLATE"]
+    return tpl
+
+
+def render_report_email(files):
+    """Renders the report e-mail from a list of files."""
+    entries = generate_integrity_report_entries(files)
+    tpl = get_report_template()
+    context = {"entries": entries}
+    return render_email_from_context(tpl, context)
+
+
+def send_integrity_report_email(files):
+    """Sends an e-mail with a report on the provided files.
+
+    APP configs ('MAIL_DEFAULT_SENDER', 'ADMIN_EMAIL_RECIPIENT') must be set.
+    Oherwise a warning is logged.
+    """
+    try:
+        if files:
+            subject = get_report_subject()
+            body = render_report_email(files)
+            sender = current_app.config["MAIL_DEFAULT_SENDER"]
+            admin_email = current_app.config["ADMIN_EMAIL_RECIPIENT"]
+            recipients = admin_email
+            if not type(admin_email) == list:
+                recipients = [admin_email]
+            mail_ext = current_app.extensions["mail"]
+            msg = Message(subject, sender=sender, recipients=recipients, body=body)
+            mail_ext.send(msg)
+    except Exception as e:
+        current_app.logger.error("Integrity report not sent. Error: {}".format(str(e)))
--- a/tests/test_tasks.py
+++ b/tests/test_tasks.py
@@ -1,0 +1,105 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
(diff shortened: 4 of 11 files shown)


invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,145 @@
...
+    )
+    if uri_prefixes:
+        files = files.filter(
+            sa.or_(*[FileInstance.uri.startswith(p) for p in uri_prefixes])
+        )
+    return files
+
+
+URLS = {
+    "deposit_html": "{base}/uploads/{id}",
```
alejandromumo, line 36: I tried to use a config here for record/deposit routes, however did not find an adequate and noted that these paths are hardcoded elsewhere.

Let me know if there is a more consistent/robust way to retrieve paths (e.g.`/uploads` and `/records`).

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,145 @@
...
+    "record_html": "{base}/records/{id}",
+}
+
+
+def ui_link_for(tpl, **kwargs):
+    """Create an UI link using specific template."""
+    UI_URL = current_app.config.get("SITE_UI_URL")
```
alejandromumo, line 43: Is `SITE_UI_URL` the correct config to be used here? It's meant to be the site ui url so e.g. `https://zenodo-rdm.web.cern.ch/records/123456` is parsed for a faulty record and sent in the report.

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,145 @@
...
+        lines.append("Created: {}".format(f.created))
+        lines.append("Checksum: {}".format(f.checksum))
+        lines.append("Last Check: {}".format(f.last_check_at))
+        if "record" in entry:
+            lines.append(
+                "Record: {}".format(
+                    ui_link_for("record_html", id=entry["record"].get("id"))
+                )
+            )
+        if "deposit" in entry:
```
alejandromumo, line 75: Was unsure whether to use `deposit` or `upload` or `draft`

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,145 @@
...
+    """Sends an e-mail with the given report.
+
+    APP configs ('SECURITY_EMAIL_SENDER', 'ADMIN_EMAIL') must be set.
+    Oherwise a warning is logged.
+    """
+    try:
+        if report:
+            subject = "Zenodo files integrity report [{}]".format(datetime.now())
+            body = format_file_integrity_report(report)
+            sender = current_app.config.get("SECURITY_EMAIL_SENDER")
```
alejandromumo, line 131: Both configs `SECURITY_EMAIL_SENDER` and `ADMIN_EMAIL` must be set in order to receive an e-mail.

Let me know if there are more appropriate configs. Otherwise, this should be documenented.

tests/api/conftest.py
```diff
@@ -9,13 +9,6 @@
 """Pytest fixtures and plugins for the API application."""
 
 import pytest
-from invenio_app.factory import create_api
-
-
-@pytest.fixture(scope="module")
-def create_app():
```
alejandromumo, line 16: Simply moved it up one level. Needed this fixture in order to have a more complete `app` instance with extensions (e.g. `mail` extension)

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,145 @@
...
+
+def send_integrity_report_email(report):
+    """Sends an e-mail with the given report.
+
+    APP configs ('SECURITY_EMAIL_SENDER', 'ADMIN_EMAIL') must be set.
+    Oherwise a warning is logged.
+    """
+    try:
+        if report:
+            subject = "Zenodo files integrity report [{}]".format(datetime.now())
```
kpsherva, line 129:
```suggestion
            subject = "Zenodo files integrity report [{}]".format(datetime.now())
```
I think we need this a bit more customisable. It is not only Zenodo anymore. Maybe it makes sense using jinja instead which makes the email easily overridable ?

alejandromumo: I agree it should be customizable.

I will give it a try with Jinja, however I will need to adapt the test as well.

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,126 @@
...
+            body = render_report_email(files)
+            sender = current_app.config.get("SECURITY_EMAIL_SENDER")
+            admin_email = current_app.config.get("ADMIN_EMAIL")
+            recipients = [admin_email]
+            mail_ext = current_app.extensions.get("mail")
+            if sender and recipients and mail_ext and body:
+                msg = Message(subject, sender=sender, recipients=recipients, body=body)
+                mail_ext.send(msg)
+            else:
+                current_app.logger.warning(
```
kpsherva, line 119:
```suggestion
                current_app.logger.error(
```

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,126 @@
...
+                    "Integrity report not sent. Please configure an email extension, admin email and "
+                    "an email sender "
+                )
+    except Exception as e:
+        current_app.logger.warning(
```
kpsherva, line 124:
```suggestion
        current_app.logger.error(
```

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/report_email.html
```diff
@@ -0,0 +1,28 @@
...
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio App RDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+-#}
+
+{% set BASE_URL = config.SITE_UI_URL %}
+
+{%- for entry in entries -%}
```
kpsherva, line 12: could you attach a screenshot on how it renders as an email ?

alejandromumo: It can be found in the issue:

https://github.com/inveniosoftware/invenio-app-rdm/issues/1951#issuecomment-1326554003

kpsherva: I am not sure how do I read the report then. Is all good with files? Do the admins who receive the report need to take any action ? The text is a bit vague, and I think it could be confusing - having in mind we are implementing it now as a part of the framework, we should target for clear messages

alejandromumo: The report only contains unhealthy files. 

About taking any action, perhaps @slint can give his insight on that point. In my (short) experience I never took action on files that were unhealthy.

slint: I agree the report is not very user-friendly (even for admins :) ).

The usual follow-up is to inspect the files in the shell and underlying filesystem and troubleshoot from there (most of the time it's false-positives, e.g. a check that timed-out). I'm not sure how much "insight" we can include in the report though since the only "extra" information we have comes from the `FileInstance.last_check` value which can be:

- `None`: means that the file check errored during the operation
- `False`: means that the produced checksum didn't match the originally stored checksum (that's pretty bad)

alejandromumo: I updated the template to contain:

- Last check date: `last_check_at`
- Lasts check result: `last_check`

I added a small text to help the reader (updated the screenshot).

Let me know your opinion about it :)

invenio_app_rdm/config.py
```diff
@@ -767,3 +783,17 @@ def files_rest_permission_factory(obj, action):
     ("invenio_app_rdm/default_static_page.html", "Default"),
 ]
 """List of available templates for pages."""
+
+FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES = [
+    # 'root://eospublic'
+]
+"""URI prefixes of files their checksums should be verified"""
+
+FILES_INTEGRITY_REPORT_TEMPLATE = "invenio_app_rdm/report_email.html"
```
slint, line 792:
```suggestion
FILES_INTEGRITY_REPORT_TEMPLATE = "invenio_app_rdm/emails/files_integrity_report.html"
```

minor: for convention and easier tracking of all emails

invenio_app_rdm/utils/files.py
slint: minor: since we're using `Flask-Mail` under the hood, it might be a good idea to set `MAIL_DEFAULT_SENDER` (which would also get picked up by Flask-Security). I'm not sure why we're not setting this globally somewhere instead of the more specific `SECURITY_EMAIL_SENDER`. If that was the case we wouldn't need to pass `sender` at all.


For `ADMIN_EMAIL`,

slint: minor: since we're using `Flask-Mail` under the hood, it might be a good idea to set `MAIL_DEFAULT_SENDER` (which would also get picked up by Flask-Security). I'm not sure why we're not setting this globally somewhere instead of the more specific `SECURITY_EMAIL_SENDER`. If that was the case we wouldn't need to pass `sender` at all.

For `ADMIN_EMAIL`,

slint (commented):
Some naming/moves, and a question about the default email sender config.

invenio_app_rdm/utils/files.py
alejandromumo: About `MAIL_DEFAULT_SENDER` I am a bit sceptical to change it in this PR as it would possibly affect others. But it is a valid point that `SECURITY_EMAIL_SENDER`, if not set, defaults to `MAIL_DEFAULT_SENDER` because we're using `Flask-Mail` anyway.

invenio_app_rdm/config.py
```diff
@@ -767,3 +783,17 @@ def files_rest_permission_factory(obj, action):
     ("invenio_app_rdm/default_static_page.html", "Default"),
 ]
 """List of available templates for pages."""
+
+FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES = [
+    # 'root://eospublic'
```
ntarocco, line 788: I would remove this:
```suggestion

```

invenio_app_rdm/config.py
```diff
@@ -767,3 +783,17 @@ def files_rest_permission_factory(obj, action):
...
+]
+"""URI prefixes of files their checksums should be verified"""
+
+FILES_INTEGRITY_REPORT_TEMPLATE = "invenio_app_rdm/emails/files_integrity_report.html"
+"""Files integrity report template"""
+
+FILES_INTEGRITY_REPORT_SUBJECT = "Files integrity report"
+"""Files integrity report subject"""
+
+ADMIN_EMAIL_RECIPIENT = "info@inveniosoftware.org"
```
ntarocco, line 798: Please mention this (configuration of files check and admin email) to the notes for the upgrade/release: https://github.com/inveniosoftware/docs-invenio-rdm/issues/447
Can you also add in the notes anything else that was already done but forgotten?

invenio_app_rdm/config.py
```diff
@@ -767,3 +783,17 @@ def files_rest_permission_factory(obj, action):
     ("invenio_app_rdm/default_static_page.html", "Default"),
 ]
 """List of available templates for pages."""
+
+FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES = [
+    # 'root://eospublic'
+]
+"""URI prefixes of files their checksums should be verified"""
+
+FILES_INTEGRITY_REPORT_TEMPLATE = "invenio_app_rdm/emails/files_integrity_report.html"
```
ntarocco, line 792: ping @anikachurilova in case you need to something with this new e-mail added (e.g. in CDS-RDM)
Is the path `.../email/...` coherent with any other path we have for e-mails templates? It would be nice to have the same folder name for all e-mails template.

invenio_app_rdm/tasks.py
```diff
@@ -0,0 +1,41 @@
...
+from celery import shared_task
+from invenio_db import db
+from invenio_files_rest.models import FileInstance
+
+from .utils.files import send_integrity_report_email
+
+
+@shared_task()
+def file_integrity_report():
+    """Send a report of uhealthy/missing files to Zenodo admins."""
```
ntarocco, line 19:
```suggestion
    """Send a report of uhealthy/missing files to system admins."""
```

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def _generate_report_entry(file):
+    """Generates a single report for a file."""
+    entry = {"file": file}
+    for o in file.objects:
+        entry["filename"] = o.key
+
+        # Find records/deposits for the files
+        deposit = get_deposit_from_bucket(o.bucket_id)
```
ntarocco, line 56: I am maybe mistaken, but in RDM we use more the term `draft` than `deposit` (old). Shall we use the same name around?
```suggestion
        deposit = get_draft_from_bucket(o.bucket_id)
```

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def get_deposit_from_bucket(bucket_id):
+    """Retrieve a deposit from a bucket id."""
+    return RDMDraftMetadata.query.filter_by(bucket_id=bucket_id).one_or_none()
+
+
+def get_report_subject():
+    """Retrieves the integrity report e-mail's subject."""
+    subject = current_app.config.get("FILES_INTEGRITY_REPORT_SUBJECT")
```
ntarocco, line 78:
```suggestion
    subject = current_app.config.get("FILES_INTEGRITY_REPORT_SUBJECT", "Files integrity report")
```

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,125 @@
...
+    if not subject:
+        subject = "Files integrity report"
+    subject += " [{}]".format(datetime.now())
```
ntarocco, line 81: can we format the date/time to something more human readable? Otherwise, we get something like:
`'2022-12-02 14:32:46.357939'`
I would remove the microseconds and add the timezone (e.g. `UTC`) so that I am sure of the time... WDYT?

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,125 @@
...
+    return subject
+
+
+def get_report_template():
+    """Retrieves the template name to be used in the files integrity report."""
+    tpl = current_app.config.get("FILES_INTEGRITY_REPORT_TEMPLATE")
+    # Use a default template
+    if not tpl:
+        tpl = "invenio_app_rdm/emails/files_integrity_report.html"
```
ntarocco, line 90: can we avoid hardcoding the path here? or is this necessary?
If, for any reason, the default template is moved, this will be for sure forgotten.
We could otherwise add a test for `render_report_email` with the config `FILES_INTEGRITY_REPORT_TEMPLATE = None` so that we make sure it fails if moved.

invenio_app_rdm/utils/files.py
ntarocco: I would indeed use `MAIL_DEFAULT_SENDER`, but make sure that both `MAIL_DEFAULT_SENDER` and `SECURITY_EMAIL_SENDER` are documented in the docs.

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,125 @@
...
+
+    APP configs ('SECURITY_EMAIL_SENDER', 'ADMIN_EMAIL_RECIPIENT') must be set.
+    Oherwise a warning is logged.
+    """
+    try:
+        if files:
+            subject = get_report_subject()
+            body = render_report_email(files)
+            sender = current_app.config.get("SECURITY_EMAIL_SENDER")
+            admin_email = current_app.config.get("ADMIN_EMAIL_RECIPIENT")
```
ntarocco, line 113: There are too many `dict.get()` around: IMO, for what is required, we should use `dict[]` and fail if missing.
Better fail-fast whenever possible. Can you check all the occurrences?

invenio_app_rdm/utils/files.py
```diff
@@ -0,0 +1,125 @@
...
+            recipients = [admin_email]
```
ntarocco, line 114: can we check if `admin_email` is already a list, in case someone uses a list as value?

invenio_app_rdm/config.py
alejandromumo: It was changed based on this comment https://github.com/inveniosoftware/invenio-app-rdm/pull/1975#discussion_r1034977373 from @slint

alejandromumo: discussed with @anikachurilova  IRL and we will follow the `flask-security` pattern `/templates/<module_name>/email/*.html`

tests/test_tasks.py
```diff
@@ -0,0 +1,131 @@
...
+        app.config["ADMIN_EMAIL_RECIPIENT"] = None
+        file_integrity_report()
+        assert len(outbox) == 0
+
+
+def test_integrity_report_default_template(app, db, file_instance):
+    """Test invalid recipient address."""
+    assert file_instance.last_check is True
+
+    # Modify file contents to force checksum to fail
```
zzacharo, line 103: maybe we can extract this in a reusable function so we improve readability here?

zzacharo (approved):
minor comment in tests!

Issue #1931, opened 2022-10-12
OAI-PMH: Configure custom sets (e.g. for OpenAIRE)


zzacharo:
Create a feature mechanism to load oai sets from an `oai_sets.yaml` containing the below:

Based on the [OpenAIRE guidelines](https://guidelines.openaire.eu/en/latest/data/use_of_oai_pmh.html):

```yaml
- name: "OpenAIRE Data"
  spec: "openaire_data" 
  search_pattern: "resource_type.type:dataset"
```

You can find an example on how we load [users](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/users.py) in invenio-rdm-records!

PR #1979, opened 2022-11-24
add the feature mechanism to load oai sets

anikachurilova:
* add cli command for creating fixtures
* cover the fixture and cli with tests
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1931

### DIFF
--- a/invenio_app_rdm/fixtures/__init__.py
+++ b/invenio_app_rdm/fixtures/__init__.py
@@ -9,6 +9,7 @@
 
 from pathlib import Path
 
+from invenio_app_rdm.fixtures.oai_sets import OAICustomSets
 from invenio_app_rdm.fixtures.pages import StaticPages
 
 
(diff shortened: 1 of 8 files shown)


tests/fixtures/app_data/oai_sets.yaml
```diff
@@ -0,0 +1,3 @@
+- name: "Set1"
+  spec: "set1"
+  search_pattern: "resource_type.type:dataset"
```
slint, line 3:
```suggestion
  search_pattern: "metadata.resource_type.id:dataset"
```

minor: just for correctness in case someone greps around for how to filter by resource type :)

invenio_app_rdm/fixtures/__init__.py
```diff
@@ -25,3 +26,16 @@ def run(self, force=False):
             [Path("./app_data/pages"), dir_ / "pages"],
             force,
         ).load()
+
+
+class FixturesEngine:
```
slint, line 31: minor: I have the feeling that we could merge the `Pages` under `FixturesEngine`, since there's no special config or difference between the two (unless there is a reason to run these two separately). Same would apply to the CLI commands... @jrcastro2 WDYT?

slint (commented):
A decision to be made on code/CLI organization, but otherwise LGTM

kpsherva (approved):

PR #1986, opened 2022-11-29
expand: pass identity on links

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1987

### DIFF
--- a/invenio_app_rdm/users_ui/views/dashboard.py
+++ b/invenio_app_rdm/users_ui/views/dashboard.py
@@ -8,7 +8,7 @@
 
 """RDM User dashboard views."""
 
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/users_ui/views/dashboard.py
```diff
@@ -8,7 +8,7 @@
 
 """RDM User dashboard views."""
 
-from flask import render_template
+from flask import g, render_template
```
kpsherva, line 11: check the two files above this comment (for some reason I couldn't comment there):
I think this needs to be reverted, somehow the file was replaced with a symlink

PR #1988, opened 2022-12-01
deposit: parameterize templates

Samk13:

### Description

This PR parameterize deposit render templates for `create `and `edit` page. This will improve the overriding process,
see concrete issue description on [discord](https://discord.com/channels/692989811736182844/704625518552547329/1047339417322405918)

There are other templates hardcoded, please let me know if we should include them all in this PR.

### Checklist

- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -3,6 +3,7 @@
 # Copyright (C) 2019-2020 CERN.
 # Copyright (C) 2019-2020 Northwestern University.
 # Copyright (C) 2021 Graz University of Technology.
+# Copyright (C) 2022 KTH Royal Institute of Technology
 #
 # Invenio App RDM is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -538,6 +539,9 @@ VOCABULARIES_DATASTREAM_WRITERS = {
 
 SEARCH_UI_SEARCH_TEMPLATE = "invenio_app_rdm/records/search.html"
 """Search page's base template."""
+
+APP_RDM_DEPOSIT_FORM_TEMPLATE = "invenio_app_rdm/records/deposit.html"
+"""Deposit page's form template."""
 
 APP_RDM_USER_DASHBOARD_ROUTES = {
     "uploads": "/me/uploads",
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -3,6 +3,7 @@
 # Copyright (C) 2019-2021 CERN.
 # Copyright (C) 2019-2021 Northwestern University.
 # Copyright (C)      2021 TU Wien.
+# Copyright (C) 2022 KTH Royal Institute of Technology
 #
 # Invenio App RDM is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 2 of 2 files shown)


invenio_app_rdm/config.py
```diff
@@ -539,6 +539,12 @@ def files_rest_permission_factory(obj, action):
 SEARCH_UI_SEARCH_TEMPLATE = "invenio_app_rdm/records/search.html"
 """Search page's base template."""
 
+DEPOSIT_CREATE_TEMPLATE = "invenio_app_rdm/records/deposit.html"
```
zzacharo, line 542: Just to align with the rest of the [deposit config](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/config.py#L607) I would prepend the `APP_RDM` into the template name.

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -338,7 +338,7 @@ def new_record():
 def deposit_create(community=None):
     """Create a new deposit."""
     return render_template(
-        "invenio_app_rdm/records/deposit.html",
+        current_app.config["DEPOSIT_CREATE_TEMPLATE"],
```
zzacharo, line 341: Do you have a use case when you want to differentiate the create from the edit template? I am hesitating to split them even If I can think of a case...But, if there is none at the moment I would keep it the same :) Maybe @kpsherva @ntarocco can comment on that too!

kpsherva: I agree with @zzacharo, I would like to understand what would be the use case to separate them. I find it very dangerous due to complexity of the form, this change opens the gate for `current complexity^2` which has consequences for maintainability and increases the risk of possible unwanted outcomes

Samk13: One use case that I might think of off on top of my head, is if I want to show particular text that should not be displayed on the other template, yes we can create an if statement, but reducing coupling is also a valid point when we look for flexibility in my opinion.

invenio_app_rdm/config.py
kpsherva: what would be the use case for bringing this change, since the templates can be overridden by placing them in correct folder structure?

invenio_app_rdm/records_ui/views/deposits.py
zzacharo: I agree with the flexibility option, but I would enable it when needed than from the beginning as it adds on maintenance complexitiy! It's an easy change to add afterward when an actual use case would be needed!

invenio_app_rdm/config.py
Samk13: @kpsherva Please read [discord ](https://discord.com/channels/692989811736182844/704625518552547329/1047339417322405918)

zzacharo (approved):

PR #1994, opened 2022-12-06
Redirector module

alejandromumo:
closes  https://github.com/zenodo/zenodo-rdm/issues/104

### DIFF
--- a/invenio_app_rdm/redirector/resource.py
+++ b/invenio_app_rdm/redirector/resource.py
@@ -1,0 +1,105 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio App RDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+"""Invenio RDM redirector resource."""
+
+from flask import redirect, request
+from flask_resources import Resource, ResourceConfig, route
+from invenio_records_resources.services.base.config import ConfiguratorMixin, FromConfig
+
+
+class RedirectorResource(Resource):
+    """Redirector resource class.
+
+    Usage:
+
+    * Example 1 (target as an absolute url)::
+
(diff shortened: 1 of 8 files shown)


invenio_app_rdm/redirector/resource.py
```diff
@@ -0,0 +1,66 @@
...
+                )
+            )
+        return url_rules
+
+    def redirect_view_factory(self):
+        """Factory to create views which redirect requests to a URL."""
+        bp_endpoint = request.url_rule.endpoint
+
+        # Strip BP name prefix from endpoint
+        endpoint_name = bp_endpoint.replace(f"{self.config.blueprint_name}.", "")
```
alejandromumo, line 42: Endpoints are registered in the blueprint as `<blueprint_name>.<endpoint_name>`.

This reverses it to `endpoint_name` so it can be matched with the endpoint name specified in the configuration

invenio_app_rdm/redirector/resource.py
```diff
@@ -0,0 +1,66 @@
...
+        triggered_rule = self.rules[endpoint_name]
+        target = triggered_rule["target"]
+        redirect_url = target
+        code = 301
```
alejandromumo, line 46: By default it sets HTTP status code to 301 (permanently redirected). 

The redirection rule can overwrite it (see below).

invenio_app_rdm/redirector/resource.py
```diff
@@ -0,0 +1,66 @@
...
+
+        # Retrieve url from target if it's a method
+        if callable(target):
+            target_output = redirect_url()
+            if type(target_output) == tuple:
+                redirect_url, code = target_output
```
alejandromumo, line 52: if `target` returns a tuple (`url`, `code`) then both are used for the redirection

zzacharo (approved):

Issue #2002, opened 2022-12-16
Fix ADMIN_EMAIL_RECIPIENT not picked up


ppanero:
This config variable is located in app-rdm
- Should it be here? If the implementation is zenodo specific (it is used for file integrity checks), then it should be moved to zenodo-rdm/site folder.
- If not, it should be renamed to `APP_RDM_...` so it can be picked up (also overwritable). 
- Extra: should it default to `MAIL_DEFAULT_SENDER` if not present?
- Tests: currently the task is being tested using `app.config[]=`, so the task is properly tested but the config loading won't be tested.

ppanero:
@alejandromumo could you shed some light on the first point? just to know how to proceed

alejandromumo:
> Should it be here? If the implementation is zenodo specific (it is used for file integrity checks), then it should be moved to zenodo-rdm/site folder.

It is used for file integrity checks which are part of app-rdm now. 

> If not, it should be renamed to APP_RDM_... so it can be picked up (also overwritable).

Agree with the naming convention.

> Extra: should it default to MAIL_DEFAULT_SENDER if not present?

I think flask-mail does this by default. However we can / should check

> Tests: currently the task is being tested using app.config[]=, so the task is properly tested but the config loading won't be tested.

I agree that we should also test that the config is loaded. We can test wether the e-mail was sent by loaded config or the default one.

PR #2006, opened 2023-01-06
config: re-order and remove unused

ppanero:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2002

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -700,15 +700,15 @@ APP_RDM_DETAIL_SIDE_BAR_TEMPLATES = [
 ]
 """Template names for detail view sidebar components"""
 
-FILES_INTEGRITY_REPORT_TEMPLATE = (
+APP_RDM_FILES_INTEGRITY_REPORT_TEMPLATE = (
     "invenio_app_rdm/files_integrity_report/email/files_integrity_report.html"
 )
 """Files integrity report template"""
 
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/config.py
```diff
@@ -702,6 +700,29 @@ def files_rest_permission_factory(obj, action):
 ]
 """Template names for detail view sidebar components"""
 
+FILES_INTEGRITY_REPORT_TEMPLATE = (
```
ppanero, line 703: These 3 are actually being picked up. Mainly because they are in a config module entry point (not by the ext.py). 

**Major: Should I prefix them with `APP_RDM` for consistency with the above?**

kpsherva: I think yes, we should keep them consistent

kpsherva (approved):
LGTM!

invenio_app_rdm/config.py
```diff
@@ -702,6 +700,29 @@ def files_rest_permission_factory(obj, action):
 ]
 """Template names for detail view sidebar components"""
 
+APP_RDM_FILES_INTEGRITY_REPORT_TEMPLATE = (
```
zzacharo, line 703: I agree with the naming convention but just for clarification this was not needed for the config to be picked up as the invenio-app-rdm config is defined as `invenio_config.module`in setup.cfg.

zzacharo (approved):

Issue #2024, opened 2023-01-19
Landing Page: In the versions box, the DOIs could be displayed as links


max-moser:
One of the boxes on record landing pages show the available versions of the record.
If the versions have a DOI each, that will be displayed, e.g.:
![image](https://user-images.githubusercontent.com/6437519/213401088-c743622c-aa2e-4383-85e0-24d7489f45c0.png)

Currently, this DOI is just a text snippet that omits the `https://doi.org/`.
Our researchers have asked if there was a reason for that, because it makes the copy/pasting of the record's DOI more difficult.

PR #2025, opened 2023-01-19
Landing page: show version DOIs as links to doi.org

max-moser:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2024


max-moser:
Tested with a fresh `my-site` from master (app-rdm v11 dev13).
Versions box before:

image

After:
image

After, on hover (cursor not visible):
image

The link points to `https://doi.org/10.70124/6ryg6-g4010` (which won't work because it's a DataCite test mode DOI only)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js
```diff
@@ -40,11 +40,12 @@ const RecordVersionItem = ({ item, activeVersion }) => {
         )}
 
         {doi && (
-          <small
+          <a
+            href={`https://doi.org/${doi}`}
```
fenekku, line 44: very minor: for some reason I prefer the underline always being visible. I think it's because "Version" above is already a link, so one could think they both link to the same destination, but they don't necessarily/are different links, so a visual distinctness helps. But this is perhaps fuzzy, so not a blocker.

fenekku (approved):
Comment is not a blocker :+1:

kpsherva (approved):

chokribr:
Coordination: @kpsherva @fenekku two aproval gained  shall we move this PR to Approved?

kpsherva (approved):

PR #2026, opened 2023-01-19
theme: fix Safari bug for auto-column-grid

jennur:
Closes https://github.com/inveniosoftware/demo-inveniordm/issues/122

- Adds max-width to second `.auto-column-grid` child to fix excess spacing bug in Safari.
## Video recording

<video src="https://user-images.githubusercontent.com/21052053/213418936-c53473a8-5cca-470f-b68a-b997e66bc164.mov" />


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -495,6 +495,12 @@ dl.details-list {
     align-items: flex-start;
     row-gap: 1rem;
   }
+
+  > *:nth-child(2) {
```
kpsherva, line 499: does it make sense to apply in invenio-theme instead ?

jennur: It should be part of the `.auto-column-grid` class, I think we only have this class in app-rdm?

kpsherva:
we should move the whole class to invenio-theme because we are using it in invenio-administration (check the screenshot), so it should be global
Screenshot 2023-01-27 at 17 33 56

jennur:
> we should move the whole class to invenio-theme because we are using it in invenio-administration (check the screenshot), so it should be global Screenshot 2023-01-27 at 17 33 56

Ok, I see! Will move it! :)

jennur:
## Update

Removed `auto-column-grid` class, added to `invenio-theme` https://github.com/inveniosoftware/invenio-theme/pull/347

kpsherva (approved):

Issue #2018, opened 2023-01-18
override.js: estimate the effort to apply it everywhere


jrcastro2:

jrcastro2:
Places where it is currently being used:

- Invenio-app-rdm:
  - [invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/frontpage/index.js](https://github.com/inveniosoftware/invenio-app-rdm/blob/dcf4205b9c43e0af69ace752b1f043cc92f7c825/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/frontpage/index.js)
  - [invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/index.js](https://github.com/inveniosoftware/invenio-app-rdm/blob/3720dead09e644d99e21c71219ff7a1847c0feab/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/index.js)
- invenio-communities:
  - [invenio_communities/assets/semantic-ui/js/invenio_communities/community/communities-carousel/index.js ](https://github.com/inveniosoftware/invenio-communities/blob/6aa44a7b16ca063c4e26c92a30be580161e99f1d/invenio_communities/assets/semantic-ui/js/invenio_communities/community/communities-carousel/index.js)

In these places the override file it's placed inside the folder that implements the component with the overriden ones. As pointed out by @kpsherva, this forces the users to follow the same folder structure and to create multiple override.js files to customize it.
For instance, for invenio-app-rdm, in order to customize the fronpage and the deposit, the following structure would be needed.
```
assets/js/invenio_app_rdm/
└── frontpage
    └──override.js
└── deposit
   └──override.js
```

An alternative to this is to keep the override at the parent module folder and have as many variables as components that accept overrivable components.

```
assets/js/invenio_app_rdm/
└──override.js
```

The override.js can contain both variables, one for the fronpage and another for the deposit.

```javascript
export const depositOverriddenComponents = {...}
export const fronpageOverriddenComponents = {...}
``` 

Any ohter ideas are welcomed on how to improve this!

Places where it is missing and should be added:

- invenio-communities:  
[...]

PR #2033, opened 2023-01-20
overridableRegistry: add global mapping for overriden components

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2018

### DIFF
--- a/invenio_app_rdm/communities_ui/templates/semantic-ui/invenio_communities/details/index.html
+++ b/invenio_app_rdm/communities_ui/templates/semantic-ui/invenio_communities/details/index.html
@@ -19,6 +19,6 @@
 {%- block page_body %}
 {{ super() }}
 <div data-invenio-search-config='{{
-  search_app_communities_records_config(app_id="rdm-search",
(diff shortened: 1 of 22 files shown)


invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/page.html
```diff
@@ -0,0 +1,18 @@
...
+  #}
+
+{%- extends "invenio_theme/page.html" %}
+
+{%- block body %}
+  {{ super() }}
+
+  {%- block javascript %}
+    {{ super() }}
+    {{ webpack['overridable-registry.js'] }}
```
kpsherva, line 15: you could just add the sctipt to `/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/theme.js`, so you won't need to modify the template and BASE_TEMPLATE config.
Alternatively, it can be placed under `invenio-app-rdm/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/javascript.html`, the template which holds all the attached js assets

Issue #2023, opened 2023-01-18
Replace import of invenio_pages Page to use the service instead.


Pineirin:
Currently, invenio_app_rdm uses in various files the Page class from invenio_pages, directly imported.
First, this needs to change, as the Page class is getting renamed to PageModel.
Secondly, instead of using directly the class, the pages_service should be used instead.

PR #2047, opened 2023-01-27
App-rdm: use invenio-pages service

Pineirin:
closes: #2023
requires: https://github.com/inveniosoftware/invenio-pages/pull/72


tests/fixtures/test_fixtures.py
```diff
@@ -39,7 +39,7 @@ def test_load_pages(app):
         force=True,
     ).load()
 
-    pages = Page.query.all()
+    pages = current_pages_service.search(system_identity)._results
```
jrcastro2, line 42: Minor: Single leading underscores are generally used to defined "internal" attributes, meaning that we should avoid accessing them outside of the class.

ntarocco (approved):

Pineirin:
Changes from last review: change system_identity for superuser_access.

ntarocco (approved):

Pineirin:
Reverted to using system_identity, as using superuser-access required to use a problematic hack.

PR #2056, opened 2023-02-03
inject user communities memberships

zzacharo:
- requires https://github.com/inveniosoftware/invenio-communities/pull/897
- closes https://github.com/inveniosoftware/invenio-communities/issues/871

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -13,6 +13,7 @@
 from flask import current_app, g, render_template
 from flask_babelex import lazy_gettext as _
 from flask_login import login_required
+from invenio_communities.proxies import current_communities
 from invenio_i18n.ext import current_i18n
 from invenio_rdm_records.proxies import current_rdm_records
 from invenio_rdm_records.resources.serializers import UIJSONSerializer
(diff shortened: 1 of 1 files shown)


ntarocco (approved):

alejandromumo (approved):

kpsherva (approved):

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -284,6 +285,14 @@ def load_custom_fields():
     }
 
 
+def get_user_communities_memberships():
+    """Return current identity communities memberships."""
+    memberships = current_communities.service.members.config.record_cls.get_memberships(
```
ppanero, line 290: We are accessing data layer from views (resources). Wouldn't it make sense to abstract this such that the `record_cls` is not exposed outside of the service?

```
memberships =. current_communities.service.get_memberships()
```

Then the rest would happen in the communities service. I'd like to avoid breaking the encapsulation, anything that is in part of the service should be exposed over methods (i.e. config). Subservices is a thin line, so there could potentially be  `communities.service.members.get_membership()`.

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -284,6 +285,12 @@ def load_custom_fields():
     }
 
 
+def get_user_communities_memberships():
+    """Return current identity communities memberships."""
+    memberships = current_communities.service.members.read_memberships(g.identity)
```
zzacharo, line 290: Thanks for that @ppanero! I totally missed that....

PR #2050, opened 2023-01-30
ui: MARCXML export option

TLGINO:
* added UI option for exporting to MARCXML
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1165

PR #2064, opened 2023-02-07
oai-pmh: UI export for marcxml

TLGINO:
* added UI for exporting to marcxml for oai
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1173

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -459,6 +459,11 @@ OAISERVER_METADATA_FORMATS = {
         "schema": "http://schema.datacite.org/oai/oai-1.1/oai.xsd",
         "namespace": "http://schema.datacite.org/oai/oai-1.1/",
     },
+    "oai_marcxml": {
+        "serializer": "invenio_rdm_records.oai:oai_marcxml_etree",
+        "schema": "https://www.loc.gov/standards/marcxml/schema/MARC21slim.xsd",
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/config.py
```diff
@@ -635,6 +640,14 @@ def files_rest_permission_factory(obj, action):
         "content-type": "application/x-dc+xml",
         "filename": "{id}.xml",
     },
+    "marcxml": {
```
ppanero, line 643: minor: this is duplicated but will be fixed when https://github.com/inveniosoftware/invenio-app-rdm/pull/2050 is merged and this PR rebased.

invenio_app_rdm/config.py
```diff
@@ -651,8 +656,17 @@ def files_rest_permission_factory(obj, action):
         "content-type": "application/vnd.geojson+json",
         "filename": "{id}.geojson",
     },
+    "marcxml": {
```
ppanero, line 659: major: I think something went wrong with the rebasing, if you see the resulting file it has the `marcxml` entry duplicated. There should be only one (the one from the previous PR).

https://github.com/inveniosoftware/invenio-app-rdm/blob/357b0ef0c22d0cdb7b0d6a1a29c9e12794e891f3/invenio_app_rdm/config.py#L643

PR #2080, opened 2023-02-10
Statistics integration

max-moser:
This PR introduces the following pieces in order to integrate `Invenio-Stats` into InvenioRDM:
* search index templates for **events** and **aggregations**
* custom parts of the stats pipeline (i.e. **event builders**), adjusted to the basic needs of InvenioRDM
* default configuration that makes use of these custom parts

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -4,6 +4,7 @@
 # Copyright (C) 2019-2020 Northwestern University.
 # Copyright (C) 2021 Graz University of Technology.
 # Copyright (C) 2022 KTH Royal Institute of Technology
+# Copyright (C) 2023 TU Wien
 #
 # Invenio App RDM is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -36,6 +37,10 @@ from celery.schedules import crontab
 from flask_principal import Denial
 from invenio_access.permissions import any_user
 from invenio_rdm_records.config import RDM_NAMESPACES
+from invenio_stats.aggregations import StatAggregator
+from invenio_stats.contrib.event_builders import build_file_unique_id
+from invenio_stats.processors import EventsIndexer, anonymize_user, flag_robots
+from invenio_stats.queries import TermsQuery
 from invenio_vocabularies.config import (
     VOCABULARIES_DATASTREAM_READERS,
     VOCABULARIES_DATASTREAM_TRANSFORMERS,
@@ -62,6 +67,8 @@ from invenio_vocabularies.contrib.names.datastreams import (
 from invenio_vocabularies.contrib.names.datastreams import (
     VOCABULARIES_DATASTREAM_WRITERS as NAMES_WRITERS,
 )
+
+from .stats.event_builders import build_record_unique_id
 
 # TODO: Remove when records-rest is out of communities and files
 RECORDS_REST_ENDPOINTS = {}
@@ -312,6 +319,22 @@ CELERY_BEAT_SCHEDULE = {
     "file-integrity-report": {
         "task": "invenio_app_rdm.tasks.file_integrity_report",
         "schedule": crontab(minute=0, hour=7),  # Every day at 07:00 UTC
+    },
+    # indexing of statistics events & aggregations
+    "stats-process-events": {
+        "task": "invenio_stats.tasks.process_events",
+        "schedule": timedelta(minutes=30),
+        "args": [("record-view", "file-download")],
+    },
+    "stats-aggregate-events": {
+        "task": "invenio_stats.tasks.aggregate_events",
+        "schedule": timedelta(hours=1),
+        "args": [
+            (
+                "record-view-agg",
+                "file-download-agg",
+            )
+        ],
     },
 }
 """Scheduled tasks configuration (aka cronjobs)."""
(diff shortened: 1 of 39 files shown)


invenio_app_rdm/config.py
```diff
@@ -829,3 +852,169 @@ def files_rest_permission_factory(obj, action):
...
+        "cls": ESTermsQuery,
+        "permission_factory": None,
+        "params": {
+            "index": "stats-file-download",
+            "doc_type": "file-download-day-aggregation",
+            "copy_fields": {
+                "recid": "recid",
+                "parent_recid": "parent_recid",
+            },
+            "query_modifiers": [],
```
max-moser, line 989: Having a query modifier of `lambda q, **_: q.filter("term", via_api=False)` here resulted in zero views all the time (even regardless of whether `via_api=True` or `via_api=False`)

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -152,6 +158,9 @@ def record_file_preview(
         filename=file_metadata.data["key"],
         preview=1 if is_preview else 0,
     )
+
+    # TODO: should we count file preview as a download?
```
max-moser, line 162: i guess not?

max-moser:
A few observations that I've made yesterday while testing out the `via_api` field:

## Lost events with nonzero double click window

The landing page creates an API call for the citation text export, so every visit will trigger an API _and_ UI `record-view` event, so we have two similar events in quick succession.
With the default `EventIndexer` implementation and a `double_click_window` > 0, this will discard one of the events (sometimes the UI event). If we ignore the API events in a later step by default, this will result in lost statistics events.

For now, I'm simply discarding the `via_api` events in the `EventIndexer`, before the `double_click_window` event triggers.


## Aggregating API and UI events separately isn't trivial

Further, I saw that the `StatsAggregator` will (unsurprisingly) override older aggregations with the same `unique_id`.
So this is the first point where we need some sort of decision on how to proceed with the UI and API events:

We could simply discard API events at this point by using a more general `unique_id` and filtering (per query modifier) `via_api=False`. By customizing the query modifiers, we could create only the aggregations that we care about.
However, this will opinionate the aggregations and reduce flexibility in the queries.

For yesterday's testing, I've built the `unique_id` by adding a `ui` or `api` prefix to the recid and used two separate aggregators with different query modifiers (`via_api=True/False`).
This did work and AFAIK the `unique_id` doesn't have any requirements regarding semantics and only serves as an arbitrary unique ID for the aggregator.
However, this _does_ incorporate the same information as the `via_api` field and makes the latter redundant.

invenio_app_rdm/config.py
max-moser: Update: This was due to the `via_api` field missing from the templates.

invenio_app_rdm/stats/templates/aggregations/aggr_record_view/os-v2/aggr-record-view-v1.json
```diff
@@ -0,0 +1,46 @@
...
+      },
+      "parent_recid": {
+        "type": "keyword"
+      },
+      "unique_id": {
+        "type": "keyword"
+      },
+      "via_api": {
+        "type": "boolean"
+      }
```
max-moser, line 40: depending on the way we want to go, not all fields are going to be used... however, not having them mapped to the correct types can make them unusable

invenio_app_rdm/stats/event_builders.py
```diff
@@ -0,0 +1,96 @@
...
+    """Add metadata of the record in question."""
+
+    # NOTE: the selection of fields is based on the Zenodo stats module
+    #       (see: zenodo.modules.stats.utils:extract_event_record_metadata)
+    assert "record" in kwargs
+    record = kwargs["record"]
+    event.update(
+        {
+            "recid": record.get("id", None),
+            "parent_recid": record.parent.get("id", None),
```
max-moser, line 83: should we add information about the records' communities here, similar to what zenodo stores?
how about other fields, like owners, etc.?

also, this could probably be merged together with the `record_view_event_builder`... i don't think anybody would want either of these functions separately :eyes:

invenio_app_rdm/stats/event_builders.py
```diff
@@ -0,0 +1,96 @@
...
+        }
+    )
+    return event
+
+
+def build_record_unique_id(event):
+    """Build record unique identifier."""
+    # NOTE: the 'unique_id' is used by the aggregators to know which events
+    #       should be aggregated together... since we want to distinguish between
+    #       API and UI events, this needs to be incorporated into the 'unique_id'
```
slint, line 93: minor: it might be that this is not entirely necessary for aggregators, since we can apply filters to the query, or even "split" the aggregated counts inside the aggregated document, to produce something like:


```python
{
  ...
  "via_api": [
    {"key": true, "doc_count": 123},
    {"key": false, "doc_count": 567},
  ]
}
```

slint (commented):
Nice work! Some minor comments/confirmations on merging functions.

Maybe it would be good to merge this one as is, and address the additional event fields to be captured in a separate issue/PR (since deciding what and how to capture might become a big topic).

invenio_app_rdm/stats/event_builders.py
max-moser: oh wow, that would potentially solve the entire aggregations issue as i've documented above in one go

invenio_app_rdm/stats/event_builders.py
max-moser: Merged the builders, created a follow-up issue: https://github.com/inveniosoftware/invenio-app-rdm/issues/2091

invenio_app_rdm/stats/templates/aggregations/aggr_file_download/os-v1/aggr-file-download-v1.json
```diff
@@ -0,0 +1,69 @@
+{
+  "index_patterns": ["__SEARCH_INDEX_PREFIX__stats-file-download-*"],
```
ppanero, line 2: Shouldn't `__SEARCH_INDEX_PREFIX__` be taken care of in the search/indexer/stats level? (i.e. make it aware of the prefix?) Right now we are centralizing that in invenio-search. Maybe we need to move it somewhere different to have it all in one place.

ppanero (approved):
Few minors/questions, otherwise LGTM. Very nice job!

- How is the SEARCH_PREFIX actually handled?
- Should we use major.minor.patch for search indeces/templates? as we do for other's in rdm-records (e.g. records)

invenio_app_rdm/stats/templates/aggregations/aggr_file_download/os-v1/aggr-file-download-v1.json
slint: This is actually [an `invenio-search` feature](https://github.com/inveniosoftware/invenio-search/blob/18f6ea3adfec2f8e3e551d787df9b08001903d4b/invenio_search/ext.py#L413-L423) 🤯, so it's already centralized.

ppanero: Thanks! 🤯 👁️‍🗨️ 👁️‍🗨️  (I'll retreat to process this)

invenio_app_rdm/config.py
```diff
@@ -36,6 +37,10 @@
 from flask_principal import Denial
 from invenio_access.permissions import any_user
 from invenio_rdm_records.config import RDM_NAMESPACES
+from invenio_stats.aggregations import StatAggregator
+from invenio_stats.contrib.event_builders import build_file_unique_id
+from invenio_stats.processors import EventsIndexer, anonymize_user, flag_robots
+from invenio_stats.queries import ESTermsQuery
```
kpsherva, line 43: nit: what does `ES` prefix stand for in `ESTermsQuery` ?

slint: That would be short for "**ES** is the only search engine we ever imagine supporting"...

/s

I guess this could be dropped to be consistent with our generalizations for search backends.

kpsherva (approved):
LGTM!

invenio_app_rdm/config.py
kpsherva: in this case I think now it is the time, since we are transitioning, @max-moser is it possible to change the naming at this point ?

max-moser: oops, yeah, missed that one while i was at it :sweat_smile: 
sure, i'd suggest renaming it to `SearchTermsQuery` etc. and maybe providing back-compat aliases similar to [this guy here](https://github.com/inveniosoftware/invenio-stats/blob/master/invenio_stats/receivers.py#L75-L76)?

max-moser: on second thought, let's ~~not go to camelot – it's a silly place~~ just strip the `ES` prefix?
here's the [PR](https://github.com/inveniosoftware/invenio-stats/pull/131) for the renaming, with a bunch of other elasticsearch references being fixed too

invenio_app_rdm/stats/templates/aggregations/aggr_file_download/os-v1/aggr-file-download-v1.0.0.json
```diff
@@ -0,0 +1,69 @@
...
+      "bucket_id": {
+        "type": "keyword"
+      },
+      "volume": {
+        "type": "double"
+      },
+      "unique_id": {
+        "type": "keyword"
+      },
+      "record_id": {
```
max-moser, line 49: The `{record,parent}_id` mappings aren't used (anymore) and could be removed.
Having them defined as extras doesn't hurt too much, but may confuse developers.
Should we remove them?

General question: Should we allow dynamic mappings, and set the default type to "keyword"?
On one hand, this may make additions later on a bit easier, but on the other it might mask errors in the mappings.

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -75,6 +76,43 @@ def open(self):
...
+    """Fetch the statistics for the given record."""
+    try:
+        views = current_stats.get_query("record-view").run(recid=recid)
+        views_all = current_stats.get_query("record-view-all-versions").run(
+            parent_recid=parent_recid
+        )
+        downloads = current_stats.get_query("record-download").run(recid=recid)
+        downloads_all = current_stats.get_query("record-download-all-versions").run(
+            parent_recid=parent_recid
+        )
```
max-moser, line 89: The cached `current_stats.get_query()` is introduced in https://github.com/inveniosoftware/invenio-stats/pull/133/files
This would reduce wastefulness a bit, but require a `3.0.1` (or `3.1.0`?) release of `invenio-stats`. Then the bump would probably need to be done in my other PR (https://github.com/inveniosoftware/invenio-records-resources/pull/439) 

If you prefer, we can leave the `3.0.0` requirement and add the caching here.

Issue #2088, opened 2023-02-15
Add banner macro to base template


anikachurilova:
Make use of `invenio-banners` functionality

PR #2089, opened 2023-02-15
header: add banner macro

anikachurilova:
* add latest invenio-banners to the requirements
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2088

Preview:
Screenshot 2023-02-15 at 16 53 50


ppanero (approved):
Nice job!

Note to self: the default banners template has no structure (e.g. divs), so default should not affect the display if there is no active banners. [Ref](https://github.com/inveniosoftware/invenio-banners/blob/master/invenio_banners/templates/semantic-ui/invenio_banners/banner.html)

PR #2107, opened 2023-02-24
serializers: ui export for bibtex

TLGINO:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1208


ppanero (approved):
requires rdm-records 2.5.0 released

Issue #1998, opened 2022-12-09
Misspelled file names in download links lead to internal server errors (500)


max-moser:
Package version (if known): 10.0.0

## Context

We've noticed that our server sometimes produces 500 responses for file downloads, because the file names in the accessed download links are transformed to lower-case.
It seems that those requests are coming from [bing](https://www.bing.com/) bots.

## Example

The following is an example of a valid file download link: https://test.researchdata.tuwien.ac.at/records/rardm-wb730/files/lurkin.jpg?download=1

If the file name part of the URL is misspelled however (e.g. by using upper-case), it becomes invalid: https://test.researchdata.tuwien.ac.at/records/rardm-wb730/files/LURKIN.JPG?download=1

Accessing the second link results in a "500 internal server error" page, and the following stack trace on the server:
```
Exception on /records/rardm-wb730/files/LURKIN.JPG [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2525, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1822, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1820, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1796, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/usr/local/lib/python3.8/site-packages/invenio_app_rdm/records_ui/views/decorators.py", line 79, in view
    return f(**kwargs)
  File "/usr/local/lib/python3.8/site-packages/invenio_app_rdm/records_ui/views/decorators.py", line 159, in view
    return f(**kwargs)
  File "/usr/local/lib/python3.8/site-packages/invenio_app_rdm/records_ui/views/records.py", line 171, in record_file_download
    return file_item.send_file(as_attachment=download)
AttributeError: 'NoneType' object has no attribute 'send_file'
```

## Expected behavior

I would expect one of the two statements to be correct:
* The file name in download links is case-insensitive
* A misspelled file name in download links leads to a 404 page

IMO, the second one would be more reasonable.
I couldn't find any spot where Invenio would generate misspelled links.
It's not Invenio's fault if clients perform any kind of text manipulation on links before accessing them.

[...]

PR #2116, opened 2023-03-01
File download: Return 404 if file to download isn't found

max-moser:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1998
Depends on https://github.com/inveniosoftware/invenio-records-resources/pull/454

### DIFF
--- a/invenio_app_rdm/records_ui/views/decorators.py
+++ b/invenio_app_rdm/records_ui/views/decorators.py
@@ -112,20 +112,16 @@ def pass_record_or_draft(expand=False):
         def view(**kwargs):
             pid_value = kwargs.get("pid_value")
             is_preview = kwargs.get("is_preview")
-
-            def get_record():
-                """Retrieve record."""
-                return service().read(id_=pid_value, identity=g.identity, expand=expand)
+            read_kwargs = {"id_": pid_value, "identity": g.identity, "expand": expand}
 
             if is_preview:
                 try:
-                    record = service().read_draft(
-                        id_=pid_value, identity=g.identity, expand=expand
-                    )
+                    record = service().read_draft(**read_kwargs)
                 except NoResultFound:
-                    record = get_record()
+                    record = service().read(**read_kwargs)
             else:
-                record = get_record()
+                record = service().read(**read_kwargs)
+
             kwargs["record"] = record
             return f(**kwargs)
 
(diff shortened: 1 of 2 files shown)


fenekku (commented):
*PR triage note*

Hmm I can see there is an issue for it in the main sprint board, but it seems apart from the rest of the sprint to me. Could/should this be part of the PR triage board? It's 
- slightly more independent (in both content and submitter)
- something we ran into as well
- I have time to review 
- placing it there publicizes the 2 previous points to others

So I will give my comments ;)

*Review*

I am leaning towards throwing the exception at the service level without even the no raise option. It simply seems more explicit: service is trying to access a non-existent "resource", so throw exception and let the client (here either the resource layer via error handler or a CLI potentially) handle it. The constant check and potential `None` return: 

```python
if file_key not in record.files
    return None
``` 
throughout the `FileService` is a little shady to me. Returning two different return types (None and FileResultItem here and `None` and something else more generally) gets us in these bugs in the first place.

Maybe there is a convincing reason to return None, though, it'd be interesting to hear.

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -158,6 +158,11 @@ def get_record_file_content():
                 item = get_record_file_content()
         else:
             item = get_record_file_content()
+
+        # note: the file service returns None if the record doesn't have the file
+        if item is None:
```
ppanero, line 163: moderate: I agree with @fenekku on raising at the service level, and I'd even create a custom exception (if doesn't exist already). This would allow a more fine grained error handling/reporting. `KeyError` is a bit too generic.  I think this might also allow to simplify the logic above.

max-moser: Perfect, then I'll rework it to match this!

Oopsy, sorry regarding the PR triage!
We're getting spammed with these error messages lately and I was a bit distracted, so I didn't pay the attention to the process I should've :sweat_smile:

fenekku: No worries about the triaging, I think it's a good discussion point to bring up with the group. Maybe the triage group will have some good thoughts on this.

rekt-hard (commented):
minor: sometimes the helper functions expect a service object, other times a function. should this be aligned? since these are only helper functions within the methods, feel free to ignore :)

max-moser:
@rekt-hard i've tried out aligning all the helper functions to take a function as argument (in the last commit called `test`), but honestly that feels like it takes away some readability again.
Personally, I'd drop the commit again and live with this little inconsistency.
WDYT?

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -113,19 +113,18 @@ def view(**kwargs):
...
-                    record = service().read_draft(
-                        id_=pid_value, identity=g.identity, expand=expand
-                    )
+                    record = _get_record(service().read_draft)
                 except NoResultFound:
-                    record = get_record()
+                    record = _get_record(service().read)
             else:
-                record = get_record()
+                record = _get_record(service().read)
```
fenekku, line 126: My "readability"/"dryness" take here would be to ask what these functions are trying to abstract or reduce duplication of. The answer is the passing of a same set of arguments, i.e. not a specific function. So something like: 

```python
read_kwargs = dict(id_=pid_value, identity=g.identity, expand=expand)
service = service()

if is_preview:
    try:
        record = service.read_draft(**read_kwargs)
    except NoResultFound:
        record = service.read(**read_kwargs)
else:
    record = service.read(**read_kwargs)
```

seems more straightforward / true (passing functions around is probably harder to follow / any function could have been passed to `_get_record` not just a function to get a record ...).

The same pattern can be reused for the other functions.

fenekku (approved):
I'd encourage the adoption of the refactoring suggested below. But overall the PR is now good to go (as soon as the other one is merged), so approved.

invenio_app_rdm/records_ui/views/decorators.py
max-moser: Actually yes, _much_ better :smile:

ppanero (approved):

rekt-hard (approved):

PR #2133, opened 2023-03-08
OAI-PMH: UI export for DCAT-AP

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1234

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -488,6 +488,11 @@ OAISERVER_METADATA_FORMATS = {
         "schema": "https://www.loc.gov/standards/marcxml/schema/MARC21slim.xsd",
         "namespace": "https://www.loc.gov/standards/marcxml/",
     },
+    "oai_dcat": {
+        "serializer": "invenio_rdm_records.oai:oai_dcat_etree",
+        "schema": "http://schema.datacite.org/meta/kernel-4/metadata.xsd",
+        "namespace": "https://www.w3.org/ns/dcat",
+    },
 }
 
 OAISERVER_LAST_UPDATE_KEY = "updated"


rekt-hard (approved):
LGTM

jrcastro2 (approved):

invenio_app_rdm/config.py
```diff
@@ -464,6 +464,11 @@ def files_rest_permission_factory(obj, action):
         "schema": "https://www.loc.gov/standards/marcxml/schema/MARC21slim.xsd",
         "namespace": "https://www.loc.gov/standards/marcxml/",
     },
+    "oai_dcat": {
+        "serializer": "invenio_rdm_records.oai:oai_dcat_etree",
+        "schema": "http://schema.datacite.org/meta/kernel-4.1/metadata.xsd",
```
ppanero, line 469: I think we are using 4.3 now, also the `datacite` has `nonexistant` in the schema... shouldn't it actually be 4.3?

slint: We might be able to actually point to [just the `kernel-4` schema.](https://blog.datacite.org/the-release-of-metadata-schema-4-4/#:~:text=Version%204.4%20is%20backwards%20compatible%2C%20and%20as%20of%20version%204.0%20you%20only%20need%20to%20specify%20the%20major%20version%20in%20your%20metadata%20submission%20(using%20http%3A//datacite.org/schema/kernel%2D4)%20so%20you%20can%20already%20take%20advantage%20of%20these%20updates%20without%20having%20to%20change%20anything%20else%20about%20how%20you%20send%20metadata%20to%20us.), i.e. `http://schema.datacite.org/meta/kernel-4/metadata.xsd`.

This is because the v4.x schemas are backwards compatible. We should update this in other DataCite serializations (we can create a separate task for this).

Issue #2143, opened 2023-03-10
sort by: most viewed


kpsherva:
- [ ] hook up the  usage statistics to search sorting, similarly to current state on zenodo:
![image](https://user-images.githubusercontent.com/38131488/224329581-2115919a-2de8-45d3-98a1-953c549b85dd.png)
- [ ] display usage statistics on the search items as shown on the mockup:
<img width="2193" alt="Screenshot 2023-03-10 at 14 39 14" src="https://user-images.githubusercontent.com/38131488/224330478-0650c87b-a968-498c-9bac-8e126077d502.png">

max-moser:
A bit of discussion regarding this happened in [Discord](https://discord.com/channels/692989811736182844/1073182429365424178/1079792482507829389).
My findings:

so i just did a bit of digging to check how the search works under the hood and how easy or hard it would be to implement a sorting by anything that's not indexed in the SE (search engine)
on the search pages, the top-right "Sort by" drop-down determines the value for the `sort` arg in the `/api/records?sort=<sort-by-value>` call that's performed to the REST API
there, it's the `resource` who handles the API call, and here the `request_search_args` parses the search-related request args (via a schema for type checking and validation, etc.) and puts them into the `resource_requestctx` that's passed as params to `service.search()`: https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/resources/records/resource.py#L80
this then goes through a bunch of service methods to figure out which parameters should be applied to the search and construct an appropriate SE query, which will then be executed and turned into a `result_list`: https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/service.py#L225 
so i wonder, can we even ever fully support a true "sort by views" if the stats aren't indexed in the SE?
because given there's enough records in the search results of that query to not yield all records in one single go, we will always be limited to the records that are in the current result page
if the record with the most views is (for whatever reason) not contained in the search result page, we can't really add it

PR #2156, opened 2023-03-21
Display views & downloads in search results

max-moser:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2143
Requires https://github.com/inveniosoftware/invenio-rdm-records/pull/1256


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
```diff
@@ -122,6 +124,9 @@ class RecordsResultsListItem extends Component {
                   </small>
                 </p>
               )}
+              <div>
+                {views} views, {downloads} downloads
```
ntarocco, line 128: Can you please add translation and also above (it was forgotten).

max-moser:
Search: 
image

Landing page:
image

slint (approved):

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
```diff
@@ -135,6 +137,11 @@ class RecordsResultsListItem extends Component {
                   </small>
                 </p>
               )}
+              <div>
+                {`${uniqueViews} ${i18next.t("views")}, ${uniqueDownloads} ${i18next.t(
```
ntarocco, line 141: Can we add singular/plural support? And also the variable should go inside the translated string, not via concatenation. We should always avoid concatenation, each language can have a different way to formulate a sentence, different order of words.

```suggestion
                {`${i18next.t("{{count}} views", {count: uniqueViews})}, ${i18next.t({{count}} downloads, {count: uniqueDownloads}}
```
Actually, ideally, it should be one string only:
`18next.t("{{uniqueViews}} views, ({{uniqueDownloads}} downloads, {uniqueViews: uniqueViews, uniqueDownloads: uniqueDownloads}}` but it looks like that [i18next requires {{count}}](https://www.i18next.com/translation-function/plurals), so it should be tested.

I see that in your screenshots you have skipped the `1` ;)

max-moser: I hoped you wouldn't catch me :unamused: 

Jokes aside, tbh I'm clueless about translations – thanks for the tip!
Will rework accordingly.

max-moser:
Update for the search:
image

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
max-moser: Made it work with the `i18next` features for [plurals](https://www.i18next.com/translation-function/plurals) and [nesting](https://www.i18next.com/translation-function/nesting), and given that we're still on v20 I use the translation keys for the [v3 JSON](https://www.i18next.com/misc/json-format#i18next-json-v3).

ppanero (approved):

ntarocco (approved):

PR #2164, opened 2023-03-29
overrides: adjust custom css

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-communities/issues/929


kpsherva:
could you add screenshots ?

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/checkbox.overrides
```diff
@@ -1,3 +1,9 @@
 /***********************************************
     Invenio App RDM Checkbox Overrides
 ***********************************************/
+
+.ui.checkbox {
+  &.confirm-delete-modal {
+    padding-top: 0.7em;
```
kpsherva, line 7: can we use existing class rel-pt-1 instead? I prefer not to maintain custom CSS

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/modal.overrides
```diff
@@ -104,6 +104,12 @@
     border-top: @actionBorder;
     text-align: @actionAlign;
   }
+
+  .content.delete-community-modal {
+    .input {
+      padding-top: 0.5em;
```
kpsherva, line 110: same here

PR #2168, opened 2023-03-30
search-ui: grammatical pedantry

dfdan:
Quick grammar fix - should be "1 more version exists" in the singular / "n more versions exist" in the plural


dfdan:
I give up, why does that fail? :(

fenekku:
I think the linter wants "this" to be on the following line because the line is too long now :/ .

fenekku (approved):
Apart from the silly linting, it's fine by me.

dfdan:
Closing as with comment on #2188 as it deals with i18n side of this issue.

PR #2179, opened 2023-04-03
Fix sidebar boxes

kpsherva:
Screenshot 2023-04-04 at 10 44 02

### DIFF
--- a/invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-App-RDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Record migration script from InvenioRDM 11.0 to 12.0.
+
+Disclaimer: This script is intended to be executed *only once*, namely when
+upgrading from InvenioRDM 11.0 to 12.0!
+If this script is executed at any other time, probably the best case scenario
+is that nothing happens!
+"""
+
+from click import secho
+from invenio_communities.communities.records.api import Community
+from invenio_communities.communities.records.systemfields.access import ReviewPolicyEnum
+from invenio_db import db
+
+
+def execute_upgrade():
+    """Execute the upgrade from InvenioRDM 11.0 to 12.0.0.
+
+    Please read the disclaimer on this module before thinking about executing
+    this function!
+    """
+
+    def migrate_review_policy(community_record):
+        review_policy = community_record["access"].get(
+            "review_policy", ReviewPolicyEnum.CLOSED.value
+        )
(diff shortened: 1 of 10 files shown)


invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
```diff
@@ -0,0 +1,47 @@
...
+
+def execute_upgrade():
+    """Execute the upgrade from InvenioRDM 11.0 to 12.0.beta
+
+    Please read the disclaimer on this module before thinking about executing
+    this function!
+    """
+
+    def migrate_review_policy(community_record):
+        review_policy = community_record["access"].get("review_policy", "closed")
```
ntarocco, line 29: very minor:
```suggestion
        review_policy = community_record["access"].get("review_policy", ReviewPolicyEnum.CLOSED.value)
```

invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-App-RDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Record migration script from InvenioRDM 10.0 to 11.0.
```
ntarocco, line 8: Wrong version numbers in the title.

invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
```diff
@@ -0,0 +1,47 @@
...
+is that nothing happens!
+"""
+
+from click import secho
+from invenio_db import db
+from invenio_communities.communities.records.api import Community
+
+
+def execute_upgrade():
+    """Execute the upgrade from InvenioRDM 11.0 to 12.0.beta
```
ntarocco, line 22: the final version 12 is to be defined

invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    def migrate_review_policy(community_record):
+        review_policy = community_record["access"].get("review_policy", "closed")
+        community_record["access"]["review_policy"] = review_policy
+
+    secho("Starting data migration...", fg="green")
+    communities = Community.model_cls.query.all()
+
+    for community_metadata in communities:
```
ntarocco, line 35: Maybe I am bit mistaken, but why this recreating the community with the `.data`?

Isn't it enough to do:
```
for community in CommunityMetadata.query.all():
    migrate_review_policy(community)
    community.commit()
```

kpsherva: I think it is not enough because communities return the db rows, not the metadata api class, so without that I cannot do community.commit(). I think simply the inner for loop object name was misleading, I changed it

ntarocco (approved):

PR #2187, opened 2023-04-05
fix: untranslatable relative path

utnapischtim:
* this bug prevented the migration to webpack 5, because this line was
  translated to a relative path, which could not be converted to an
  absolute path. this is bad, because webpack 5 introduced a check about
  relative paths and removed those and printed an error.

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/theme.less
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/theme.less
@@ -16,7 +16,7 @@
 @import '@{themesFolder}/default/globals/site.variables';
 
 /* Packaged site.variables */
-@import '@{themesFolder}/@{siteFolder}/globals/site.variables';
+@import '@{themesFolder}/@{site}/globals/site.variables';
 
 /* Component's site.variables */
 @import (optional) '@{themesFolder}/@{theme}/globals/site.variables';


max-moser:
The change in question led to a different error for me, so I did a bit of digging.

### Webpack 5
Indeed, Webpack 5 will check if the file dependencies found by the configured loaders are [absolute paths](https://github.com/webpack/webpack/blob/main/lib/NormalModule.js#L1017).
A closer look (with some `console.log`s at some spots like [here](https://github.com/webpack/webpack/blob/main/lib/NormalModule.js#L859) and configuration experimentation) did show that there was only one offending entry in `fileDependencies`, from the `less-loader`.

### Directory structure
Note that the relevant part of the directory structure in the virtualenv should look like the following:
```
/path/to/venv/var/instance/assets
├── node_modules
│   └── semantic-ui-less
│       └── themes
└── less
    └── site
        └── globals
            └── site.variables
```

### Configuration
The [relevant variables defined in `theme.config`](https://github.com/inveniosoftware/cookiecutter-invenio-rdm/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/assets/less/theme.config#L81) have the following values:
* `@themesFolder : '~semantic-ui-less/themes';`
* `@siteFolder  : '../../less/site';`

This means that the line `@import '@{themesFolder}/@{siteFolder}/globals/site.variables';` would result in `~semantic-ui-less/themes/../../less/site/globals/site.variables`.
However, two directories up will only land in `node_modules` rather than `assets`, which requires another step up.

### Results
Changing the value to `@siteFolder  : '../../../less/site';` in the cookiecutter `theme.config` made the Webpack build without issues for me.
Especially, there's no more errors regarding any non-absolute paths pop up anymore, which hints at Webpack being able to handle relative paths as long as they can be resolved.

However, I'm not 100% sure if that's the correct change, and if other values should be changed as well.
* The `@imagesFolder : '../../images';` might have a similar issue, but there was no `images` directory to be found anywhere in my `/var/instance/assets` at all, so I can't say for sure.
* Similar for the `@fontPath : "../../../themes/@{theme}/assets/fonts";` – the value looks a lot like [aliases](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/webpack.py#L59) to me, but from a quick glance I can't see an alias (in `build/config.json`) that would match this pattern.
* Alternatively, the approach from this PR could be taken to change the variables' usage rather than their definition.

### Further findings
The [value for `@site`](https://github.com/inveniosoftware/cookiecutter-invenio-rdm/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/assets/less/theme.config#L21) is set to `rdm` in the default cookiecutter, causing the change in this PR (`@import '@{themesFolder}/@{site}/globals/site.variables';`) to result in `@import '~semantic-ui-less/themes/rdm/globals/site.variables';`, which doesn't exist (at least for me, with a fresh `my-site`).
Instead, the `rdm` theme is located in the directory `less/invenio_app_rdm/theme`, which is accessible via the [`themes/rdm` alias](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/webpack.py#L62).

kpsherva (approved):

Issue #2162, opened 2023-03-28
Fix wrong translations


kpsherva:
String should not be concatenated when translating. Each language can have different ways of expressing something, or a different words order.
❌ Wrong:
`{i18next.t("View all")} {totalCommunities} {i18next.t("communities")}`
✅ Correct:
`{i18next.t("View all {{count}} communities", {count: totalCommunities})`

Singular/plural should be also take into account:
❌ Wrong:
`{total} {i18next.t("result(s) found")}`
✅ Correct:
`{i18next.t(`View all {{count}} versions`, {count: recordVersions.total})}`
Apparently, the variable MUST be called `count`, see [here](https://www.i18next.com/translation-function/plurals).

**WARNING** The `_plural` translation should be added. This depends on `i18next` version!
- `_plural`: https://www.i18next.com/misc/json-format#i18next-json-v3
-`one`/`_other`: https://www.i18next.com/misc/json-format#i18next-json-v4

Wrong occurrences (not complete list, please look for `i18next.t(`):
- https://github.com/inveniosoftware/invenio-app-rdm/blob/submission-multiple-communities/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCommunitiesListModal/RecordCommunitiesListModal.js#L43
- https://github.com/inveniosoftware/invenio-app-rdm/blob/submission-multiple-communities/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/communityRecordsSearch/CommunityRecordsSearchAppLayout.js#L38
- https://github.com/inveniosoftware/invenio-app-rdm/blob/submission-multiple-communities/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js#L110
- https://github.com/inveniosoftware/invenio-app-rdm/blob/submission-multiple-communities/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js#L118
- https://github.com/inveniosoftware/invenio-app-rdm/blob/submission-multiple-communities/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js#L34
[...]

PR #2188, opened 2023-04-11
translations: fix string concatenation and pluralizaton

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2162

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
@@ -115,15 +115,18 @@ class RecordsResultsListItem extends Component {
                   <p>
                     {createdDate && (
                       <>
-                        {i18next.t("Uploaded on")} <span>{createdDate}</span>
+                        {i18next.t("Uploaded on {{uploadDate}}", {
+                          uploadDate: createdDate,
+                        })}
                       </>
                     )}
                     {createdDate && publishingInformation && " | "}
 
                     {publishingInformation && (
                       <>
-                        {i18next.t("Published in: ")}{" "}
-                        <span>{publishingInformation}</span>
+                        {i18next.t("Published in: {{publishInfo}}", {
+                          publishInfo: publishingInformation,
+                        })}
                       </>
                     )}
                   </p>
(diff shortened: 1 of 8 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
```diff
@@ -115,25 +115,28 @@ class RecordsResultsListItem extends Component {
...
                     )}
                   </p>
 
                   {!allVersionsVisible && versions.index > 1 && (
                     <p>
                       <b>
-                        {numOtherVersions} more{" "}
-                        {numOtherVersions > 1 ? "versions" : "version"}{" "}
-                        {i18next.t("exist for this record")};
+                        {i18next.t("{{count}} more versions exist for this record", {
```
dfdan, line 137: I've tried to solve another issue here (#2168) but wasn't aware of the efforts here... This is a bit tricky to handle. The correct phrasing is "1 more version exist**s**" in the singular and "n more version**s** exist" in the plural. 

It might be easiest to drop 'exist(s)' and simply say "n more version(s)" of this record? 

I will close #2168 to avoid conflict with efforts here.

anikachurilova: The important thing is to [mark the string](https://inveniordm.docs.cern.ch/develop/howtos/i18n/#marking-strings-for-translation) for pluralized translation. In that case, the whole string will be taken into account in translation process. 

Having the `count` property marks the string to be pluralized later on. See here: https://inveniordm.docs.cern.ch/develop/howtos/i18n/#singular-plural

Here is the result of generated string marked for translation using the code sample you provide (from the [pr](https://github.com/inveniosoftware/invenio-app-rdm/pull/2168)):
`"{{count}} more versions exist for this record": "",`

And here is the string marked for pluralization:
`  "{{count}} more versions exist for this record": "",
"{{count}} more versions exist for this record_plural": ""`

anikachurilova: Thank you for asking!

I apologize for the confusion, the generated key will look like this:
```
  "{{count}} more versions exist for this record": "{{count}} more versions exist for this record",
  "{{count}} more versions exist for this record_plural": "{{count}} more versions exist for this record",
```
which copy-pastes the key and adds the `_plural` marker.

This makes the whole string translatable for all the languages in singular and plural, BUT for English language those examples have to be fixed manually - this will be the template for the translation of all the other languages.

ntarocco (approved):

Issue #2205, opened 2023-05-02
Custom fields are not displayed in the deposit form


alejandromumo:
After migrating custom fields components to `invenio-rdm-records`, they are not being loaded in the deposit form.

PR #2206, opened 2023-05-02
deposit: import custom fields from rdm-records

alejandromumo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2205


slint:
I guess we can close this? It looks like there's something similar now there, causing a merge conflict.

PR #2222, opened 2023-05-17
deposit_form: fixed deleting draft not possible unless refresh

TLGINO:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1293

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -366,6 +366,7 @@ def deposit_create(community=None):
         permissions=get_record_permissions(
             [
                 "manage_files",
+                "delete_draft",
                 "manage_record_access",
             ]
         ),


TLGINO:
Before saving

image



After Saving

image

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -612,18 +612,16 @@ export class RDMDepositForm extends Component {
                       showMetadataAccess={permissions?.can_manage_record_access}
                     />
                   </Overridable>
-                  {permissions?.can_delete_draft && (
```
alejandromumo, line 615: isn't this check still needed? 
I am not an expert on permissions but I think they can be configured in a way that not everyone can delete a draft.

TLGINO: I'll do some more debugging on this,
The strange thing is that removing this option doesn't change the fact that you cannot delete the draft unless the record is saved
If you look at the picture you can see that the colour is lighter and that the button is not clickable, unless the record is saved, in which case the button becomes clickable.

rekt-hard: The reason for this might be in the `DeleteButton`component itself https://github.com/inveniosoftware/react-invenio-deposit/blob/040b698967e7fee14ac7fe466e49f883efbe8c47/src/lib/components/DeleteButton.js#L103

We probably need to pass the permission from the backend here: https://github.com/inveniosoftware/invenio-app-rdm/blob/342c9b04ac13177feb0454224da20f6edbe5c3ad/invenio_app_rdm/records_ui/views/deposits.py#L366-L371  similar to the edit template (https://github.com/inveniosoftware/invenio-app-rdm/blob/342c9b04ac13177feb0454224da20f6edbe5c3ad/invenio_app_rdm/records_ui/views/deposits.py#L390-L397)

TLGINO: You're right that fixed it! :rocket: 
Have updated PR

rekt-hard (approved):
LGTM

jrcastro2 (approved):

ppanero (approved):

ppanero:
Blocked until tests are passing, seems to be due to the bad releases of invenio-db that were mentioned?
@TLGINO  do they pass locally?

PR #2228, opened 2023-05-23
pids: redirection for parent_id 

TLGINO:
* closes inveniosoftware/invenio-rdm-records#1282

### DIFF
--- a/invenio_app_rdm/records_ui/views/decorators.py
+++ b/invenio_app_rdm/records_ui/views/decorators.py
@@ -20,6 +20,16 @@ from invenio_rdm_records.proxies import current_rdm_records
 from invenio_records_resources.services.errors import PermissionDeniedError
 from sqlalchemy.orm.exc import NoResultFound
 
+from ..utils import get_id_from_parent_id
+
+
+class ExceptionResult(Exception):
+    """Exception class for handling thrown result."""
+
+    def __init__(self, message):
+        """Constructor."""
+        self.message = message
+
 
 def service():
     """Get the record service."""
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -120,7 +134,13 @@ def view(**kwargs):
                 except NoResultFound:
                     record = service().read(**read_kwargs)
             else:
-                record = service().read(**read_kwargs)
+                try:
+                    read_kwargs["front_end_call"] = True
```
ntarocco, line 138: what is the need of this `front_end_call`?

ntarocco (commented):
I think that we need to re-discuss how to design this feature. If possible, we should make more clear when we fetch a record by its PID or by parent PID (returning the newest version).

For the discussion:
```
parent = RDMParent.pid.resolve(<parent-pid>)
RDMRecord.get_records_by_parent(parent)
```

invenio_app_rdm/records_ui/views/decorators.py
TLGINO: So essentially
when going through /api/
the only function in `invenio-records-resources `which is called is search (and this works fine)
the issue is that when the call is done through `app-rdm`
it first calls `records.search` on `invenio-records-resources`, then calls `files.search` 

if we throw an error with the correct pid in `records.search`, then the call to /api wil fail (but `app-rdm` will not as the error is caught)
if we don't throw an error, then `app-rdm` will try again to access a record using the wrong id (ie not of the most recent parent) and this crashes in` files.search `(one solution would be to also resolve in `files.search` the parent pid in order to find the most recent), the issue is that this could also overlap on other files

finaly going back to the original question
if we add a flag to the function, we are able to distinguish wether it is a backend (/api) or frontend call (app-rdm)

slint:
Closing in favor of the solution in https://github.com/inveniosoftware/invenio-drafts-resources/pull/237

PR #2235, opened 2023-05-30
config: add post logout view

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/16

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -263,11 +263,14 @@ MUST ensure that NGINX (or other proxies) removes these headers again before
 sending the response to the client. Set to False, in case of doubt.
 """
 
-# Invenio-Security-Invenio
+# Flask-Security
 # ========================
 
 SECURITY_EMAIL_SUBJECT_REGISTER = _("Welcome to Invenio App RDM!")
 """Email subject for account registration emails."""
+
+SECURITY_POST_LOGOUT_VIEW = "/oauth/logout"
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/config.py
```diff
@@ -263,12 +263,15 @@ def files_rest_permission_factory(obj, action):
...
 """
 
-# Invenio-Security-Invenio
+# Flask-Security
 # ========================
 
 SECURITY_EMAIL_SUBJECT_REGISTER = _("Welcome to Invenio App RDM!")
 """Email subject for account registration emails."""
 
+SECURITY_POST_LOGOUT_VIEW = "/oauth/logout"
```
ntarocco, line 272: I would avoid putting this here by default, given that maybe no oauth is configured (only local login). It probably does not hurt, but maybe better to add it to the doc [here](https://inveniordm.docs.cern.ch/customize/authentication/)?

kpsherva: with only local accounts I think this would break, to be tested?

ntarocco: no it should not, because if you have no remotes configured, it will not redirect (if I am not mistaken). However, I will remove it, and document it instead.

jrcastro2: Yes, it wouldn't break, but I will document it instead :+1:

jrcastro2:
Closed as requested and replaced by the docs: https://github.com/inveniosoftware/docs-invenio-rdm/pull/560

Issue #2229, opened 2023-05-23
requests do not displayed correctly when users do not have username


ppanero:
From the migration we have noticed that requests fail to display with the following stacktrace. Note that username is not a mandatory field (`nullable=True`, https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/models.py#L93) therefore it cannot be required to display requests (might need to default to user email or name if there is a profile).

```python
File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1799, in dispatch_request
return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
File "/usr/local/lib/python3.9/site-packages/flask_resources/resources.py", line 65, in view
return view_meth()
File "/usr/local/lib/python3.9/site-packages/flask_resources/content_negotiation.py", line 116, in inner_content_negotiation
return f(*args, **kwargs)
File "/usr/local/lib/python3.9/site-packages/flask_resources/parsers/decorators.py", line 51, in inner
return f(self, *args, **kwargs)
File "/usr/local/lib/python3.9/site-packages/flask_resources/parsers/decorators.py", line 51, in inner
return f(self, *args, **kwargs)
File "/usr/local/lib/python3.9/site-packages/flask_resources/parsers/decorators.py", line 51, in inner
return f(self, *args, **kwargs)
File "/usr/local/lib/python3.9/site-packages/flask_resources/responses.py", line 39, in inner
res = f(*args, **kwargs)
File "/usr/local/lib/python3.9/site-packages/invenio_communities/communities/resources/resource.py", line 94, in search_community_requests
return hits.to_dict(), 2File "/usr/local/lib/python3.9/site-packages/invenio_requests/services/requests/results.py", line 172, in to_dict
fields = self._fields_resolver.expand(self._identity, hit)
File "/usr/local/lib/python3.9/site-packages/invenio_records_resources/services/records/results.py", line 415, in expand
output = field.pick(identity, resolved_rec)
File "/usr/local/lib/python3.9/site-packages/invenio_requests/services/results.py", line 39, in pick
return self.entity_proxy.pick_resolved_fields(identity, resolved_rec)
File "/usr/local/lib/python3.9/site-packages/invenio_users_resources/entity_resolvers.py", line 52, in pick_resolved_fields
"username": resolved_dict["username"],
KeyError: 'username'
```

Steps to replicate:

- Create a user (i.e. register)
- Create a community
[...]

PR #2236, opened 2023-05-30
ui : fixing username not appearing

TLGINO:
* no longer shows None when neither full_name or user_name are set
 * closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2229

### DIFF
--- a/invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html
+++ b/invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html
@@ -15,7 +15,7 @@
           <img src="{{ user_avatar }}" alt="" />
         </div>
         <div class="content rel-pl-1">
-          <h4 class="header">{{ current_userprofile.full_name or current_user.username }}</h4>
+          <h4 class="header">{{ current_userprofile.full_name or current_user.username or current_userprofile.email or _('Anonymous user')}}</h4>
           <div class="meta">{{ current_userprofile.affiliations or "" }}</div>
         </div>
       </div>
@@ -29,4 +29,3 @@
   {% endfor %}
 </div>
 </div>
-


TLGINO:
When no `full_name` or `username` is provided (or deleted from database), desplay name will default to email.

image

kpsherva:
thanks for fixing it!

what about the Receiver/sender field  of a request? could you check if does it have the same issue ? 
if yes, then we cannot simply display an e-mail due to privacy concerns, we need to check if the receiver's profile allows to display the email (it is in profile visibility preferences)

invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html
```diff
@@ -15,7 +15,7 @@
           <img src="{{ user_avatar }}" alt="" />
         </div>
         <div class="content rel-pl-1">
-          <h4 class="header">{{ current_userprofile.full_name or current_user.username }}</h4>
+          <h4 class="header">{{ current_userprofile.full_name or current_user.username or (current_userprofile.email if current_userprofile.preferences.email_visibility != 'restricted' else '')}}</h4>
```
ntarocco, line 18: See my comment in the other PR ;)

TLGINO:
## Requests as seen by the community manager (in this case  martin.lettry@cern.ch) from aaa@gmail.com

### Request with a now deleted user

image

### Request with a valid user, with no name, nor username with **email visibility set to restricted**

image

### Request with a valid user, with no name, no username and with **email visibility set to public**

image

### Request as seen by the submitter of the request
Note: permision checking regarding email visibility is irrelevant as this is the current logged in user (ie if no username, nor display name then will always show email)
image

invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html
```diff
@@ -15,7 +15,7 @@
           <img src="{{ user_avatar }}" alt="" />
         </div>
         <div class="content rel-pl-1">
-          <h4 class="header">{{ current_userprofile.full_name or current_user.username }}</h4>
+          <h4 class="header">{{ current_userprofile.full_name or current_user.username or current_userprofile.email or 'Unknown user'}}</h4>
```
anikachurilova, line 18: `Unknown user` should be made translatable here as well

anikachurilova (changes requested):

anikachurilova (approved):

invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html
```diff
@@ -15,7 +15,7 @@
           <img src="{{ user_avatar }}" alt="" />
         </div>
         <div class="content rel-pl-1">
-          <h4 class="header">{{ current_userprofile.full_name or current_user.username }}</h4>
+          <h4 class="header">{{ current_userprofile.full_name or current_user.username or current_userprofile.email or _('Unknown user')}}</h4>
```
slint, line 18:
```suggestion
          <h4 class="header">{{ current_userprofile.full_name or current_user.username or current_user.email or _('Unknown user')}}</h4>
```

minor: unless I'm missing some nesting that takes place

kpsherva: I think we should go with `Anonymous user` instead of Unknown, it is better for the user experience

TLGINO: > I think we should go with `Anonymous user` instead of Unknown, it is better for the user experience

I agree

TLGINO: > minor: unless I'm missing some nesting that takes place

did some testing, both the current option and the proposed solution work

jrcastro2 (approved):

zzacharo (approved):

zzacharo (approved):

PR #2243, opened 2023-06-02
frontpage: rewrite communities carousel to python and jinja

jennur:
addresses https://github.com/inveniosoftware/invenio-communities/issues/874

See screen recording at https://github.com/inveniosoftware/invenio-communities/pull/966

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/transition.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/transition.overrides
@@ -27,41 +27,41 @@
 @keyframes carouselSlideInLeft {
   0% {
     opacity: 0;
-    transform: translate3d(100%, 0, 0);
+    transform: translateX(100vw);
   }
   100% {
-    transform: none;
+    transform: translateX(0);
   }
 }
 
 @keyframes carouselSlideOutLeft {
   0% {
     opacity: 1;
-    transform: translate3d(-100%, 0, 0);
+    transform: translateX(0);
   }
   100% {
     opacity: 0;
-    transform: translate3d(-200%, 0, 0);
+    transform: translateX(-100vw);
   }
 }
 
 @keyframes carouselSlideInRight {
   0% {
     opacity: 0;
-    transform: translate3d(-100%, 0, 0);
+    transform: translateX(-100vw);
   }
   100% {
-    transform: none;
+    transform: translateX(0);
   }
 }
 
 @keyframes carouselSlideOutRight {
   0% {
     opacity: 1;
-    transform: translate3d(-100%, 0, 0);
+    transform: translateX(0);
   }
   100% {
     opacity: 0;
-    transform: translate3d(0%, 0, 0);
+    transform: translateX(100vw);
   }
 }
--- a/invenio_app_rdm/theme/views.py
+++ b/invenio_app_rdm/theme/views.py
@@ -19,6 +19,16 @@ from invenio_i18n import lazy_gettext as _
 from invenio_users_resources.forms import NotificationsForm
 
 
+def create_url_rule(rule, default_view_func):
+    """Generate rule from string or tuple."""
+    if isinstance(rule, tuple):
+        path, view_func = rule
+
+        return {"rule": path, "view_func": view_func}
+    else:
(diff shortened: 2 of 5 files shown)


invenio_app_rdm/theme/views.py
```diff
@@ -62,11 +64,20 @@ def init_menu():
 #
 # Views
 #
+
+
 def index():
     """Frontpage."""
+    featured_communities = current_communities.service.featured_search(
```
ntarocco, line 71: what if an instance will not have feature communities in the home page? This extra search should be avoided in this case.

zzacharo: Plus one on this!

invenio_app_rdm/config.py
```diff
@@ -649,7 +649,7 @@ class UserPreferencesNotificationsSchema(UserPreferencesSchema):
 }
 
 APP_RDM_ROUTES = {
-    "index": "/",
+    "index": ("/", None),
```
zzacharo, line 652: by default this should be left as `/` right?

invenio_app_rdm/theme/views.py
```diff
@@ -33,7 +33,19 @@ def create_blueprint(app):
...
+            path, view_func = rule
+
+            if view_func == None:
+                view_func = default_view_func
+
+            return path, name, view_func
+        else:
+            return rule, name, default_view_func
+
+    blueprint.add_url_rule(*create_url_rule(rule=routes["index"], name="index", default_view_func=index))
```
zzacharo, line 48: I think it makes sense to use the same way of constructing this for the rest of the routes i.e robots, help_search, help_statistics to be consistent...

invenio_app_rdm/theme/views.py
```diff
@@ -62,6 +74,8 @@ def init_menu():
 #
 # Views
 #
+
```
zzacharo, line 77: leftover

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/transition.overrides
```diff
@@ -2,6 +2,26 @@
     Invenio App RDM Transition Overrides
 ***********************************************/
 
+#carousel-slides {
```
zzacharo, line 5: there is no carousel anymore in invenio-app-rdm so maybe is better to move these in zenodo-rdm? Or are these changes needed for the existing macros?

jennur: All the styling is for the macros and/or react component in `invenio-communities`. `zenodo-rdm` is just importing the macro from `invenio-communities`

invenio_app_rdm/theme/views.py
jennur: I added the `name` there because I couldn't specify which of the arguments were the `view_func` in `add_url_rule` through the other function (or is there a way?), so it was failing. But I can remove it from the function parameters and just add `None` in the return value?

invenio_app_rdm/config.py
jennur: I'm not sure what you mean?

zzacharo (approved):

invenio_app_rdm/theme/views.py
```diff
@@ -18,6 +18,18 @@
 from invenio_i18n import lazy_gettext as _
 from invenio_users_resources.forms import NotificationsForm
 
+# Generate rule from string or tuple
+def create_url_rule(rule, default_view_func):
+    if isinstance(rule, tuple):
+        path, view_func = rule
+
+        if view_func == None:
```
zzacharo, line 26: this can be removed as a tuple should always be used with a view_func

invenio_app_rdm/theme/views.py
```diff
@@ -18,6 +18,18 @@
 from invenio_i18n import lazy_gettext as _
 from invenio_users_resources.forms import NotificationsForm
 
+# Generate rule from string or tuple
+def create_url_rule(rule, default_view_func):
+    if isinstance(rule, tuple):
```
zzacharo, line 23: you need a docsting here :)

zzacharo (approved):

PR #2254, opened 2023-06-09
global: upgrade to invenio-assets v3

kpsherva:
* upgrades the instance to webpack 5
* fix theme import path


max-moser (approved):
that was a short-lived change :smile:

ntarocco (approved):
🚀

PR #2275, opened 2023-06-27
deposit: allow to configure the default value for files enabled

jrcastro2:
* closes https://github.com/zenodo/who-rdm/issues/71

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -343,7 +343,9 @@ def get_search_url():
 def new_record():
     """Create an empty record with default values."""
     record = dump_empty(RDMRecordSchema)
-    record["files"] = {"enabled": True}
+    record["files"] = {
+        "enabled": current_app.config.get("APP_RDM_DEPOSIT_DEFAULT_FILES_ENABLED")
+    }
     if "doi" in current_rdm_records.records_service.config.pids_providers:
         record["pids"] = {"doi": {"provider": "external", "identifier": ""}}
     else:
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -343,9 +343,7 @@ def get_search_url():
 def new_record():
     """Create an empty record with default values."""
     record = dump_empty(RDMRecordSchema)
-    record["files"] = {
-        "enabled": current_app.config.get("APP_RDM_DEPOSIT_DEFAULT_FILES_ENABLED")
-    }
+    record["files"] = {"enabled": current_app.config.get("RDM_DEFAULT_FILES_ENABLED")}
     if "doi" in current_rdm_records.records_service.config.pids_providers:
         record["pids"] = {"doi": {"provider": "external", "identifier": ""}}
     else:
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.variables
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.variables
@@ -7,3 +7,4 @@
 // Tablets/phones were left as-is because they have less screen estate
 @selectionComputerMaxMenuHeight : 575px;
 @selectionWidescreenMaxMenuHeight : 575px;
+@selectionIconPadding: (@selectionIconHitbox / @dropdownIconSize);
--- a/invenio_app_rdm/__init__.py
+++ b/invenio_app_rdm/__init__.py
@@ -16,6 +16,6 @@
 #
 # See PEP 0440 for details - https://www.python.org/dev/peps/pep-0440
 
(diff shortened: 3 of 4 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.variables
```diff
@@ -7,3 +7,4 @@
 // Tablets/phones were left as-is because they have less screen estate
 @selectionComputerMaxMenuHeight : 575px;
 @selectionWidescreenMaxMenuHeight : 575px;
+@selectionIconPadding: calc(@selectionIconHitbox / unit(@dropdownIconSize));
```
jrcastro2, line 10: This variable is defined in [semantic ui less](https://github.com/Semantic-Org/Semantic-UI-LESS/blob/master/themes/default/modules/dropdown.variables#L163) but it's not properly compiled. I see that we have done some major upgrades lately ([here the commit](https://github.com/inveniosoftware/invenio-assets/commit/656f0fe4943b748bcec54e2f6c4bce4b1b3bb22a#diff-e0c1de0eb18d1877eced2250b1c0583e65790027d37eed938d8e8f80aa8dea0aL14-L66)). I've quickly checked the changelog for some packages that might be involving in the loading of the css such as `mini-css-extract-plugin`, `less-loader` and `css-loader` but couldn't find the reason why this is failing now. Any advice on what else to check is welcomed!

jrcastro2: Dsicussed IRL, we fix the inconsistencies that we find due to the upgrade

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -343,7 +343,9 @@ def get_search_url():
 def new_record():
     """Create an empty record with default values."""
     record = dump_empty(RDMRecordSchema)
-    record["files"] = {"enabled": True}
+    record["files"] = {
+        "enabled": current_app.config.get("APP_RDM_DEPOSIT_DEFAULT_FILES_ENABLED")
```
kpsherva, line 347: I think this should be at the record service level, ping @ntarocco and @slint to confirm
potentially it can be improved later

ppanero: Naming was I think is too long, maybe removing the `DEPOSIT`. On the other hand, this would **only** affect UI and people could overpass it via REST since then it would be enabled by default (see https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/services/records/components/files.py#L39). Which leads to Karolina's conclusion, you can set it at service-component level (and then the variable naming without `DEPOSIT` makes sense. Full circle hehe

slint: Agree, it could be something similar to what we have [for the download-all-files link](https://github.com/slint/invenio-rdm-records/blob/4c9237e90af0dd3c2ddb50a22ac393356f13081a/invenio_rdm_records/resources/config.py#L181), which uses the `FromConfig` descriptor.

I feel though that we will soon run into the classic "config granularity" discussion of single-field config vars vs. overriding the entire config class. I'm not sure how we all arrived at the exact same time to need similar functionalities, but I would say the best place to centralize where record files behaviour config is consumed would be the [`FileOptionsComponent`](https://github.com/inveniosoftware/invenio-records-resources/blob/58503e2a6fdc7bd110f7d3bf3a51e7ca05fd63f0/invenio_records_resources/services/records/components.py#L82). Either that or we merge/inherit this class in [`DraftFilesComponent`](https://github.com/slint/invenio-drafts-resources/blob/3e6d6b5d289ff88e9e094e3ff419cb5fd7b6cdd6/invenio_drafts_resources/services/records/components/files.py#L21) and call it a day.

Maybe we go ahead though with the individual implementation and refactor/merge later on just so that we have something working first.

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.variables
```diff
@@ -7,3 +7,4 @@
 // Tablets/phones were left as-is because they have less screen estate
 @selectionComputerMaxMenuHeight : 575px;
 @selectionWidescreenMaxMenuHeight : 575px;
+@selectionIconPadding: (@selectionIconHitbox / @dropdownIconSize);
```
zzacharo, line 10: is this needed?

zzacharo (approved):

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/dropdown.variables
jrcastro2: Yes, to fix the icons of the dropdowns

Issue #2029, opened 2023-01-20
Update the GUI for share-by-link


max-moser:
# Original report in [Discord](https://discord.com/channels/692989811736182844/704625518552547329/1065754430529613884)

Greetings fellow RDM’ers.  I’ve recently discovered a feature in RDM that we consider a security vulnerability.  It’s the link that comes from the share modal from a records landing page.  We noted that this scenario is possible which resulted in a record’s author unwittingly providing write access to a record to someone she thought only has read:

1.    Click Share link from landing page
2.    Click Get a link, leave ‘Can view’ as selected option in drop down
3.    Copy link, send that ‘view’ link to Joe Schmo.
4.    Change ‘can view’ to ‘Can edit’ in drop down
5.    Copy link, send that ‘edit’ link to Jane Doe 

At this point Joe can edit the record just as Jane can.  Our instance of RDM does not currently need this share-able link feature and so we’re simply removing that button from the UI to address the vulnerability.  However, I wanted to share this concern to make others aware. 

![image](https://user-images.githubusercontent.com/6437519/213653198-21cca579-faf1-41b2-9505-ee4c5e624771.png)

# My take

The current version of the UI is not clear on the fact that changing the permissions of the _existing_ token, rather than creating a new one – especially given that the URL in the text box changes slightly.
This makes it easy to accidentally elevate the permissions given to people who already have the existing link.

Further, the back-end supports additional features that are not exposed via the GUI yet.
This includes the setting of an expiration date for newly generated links, and the management of multiple links per record.

# Proposed changes

Changing the permissions of an existing link should ask for confirmation and make the outlined issue clear.

Further, it should be possible to manage multiple links for the record in the front-end, to further disincentivize updates to the existing links' permissions.

PR #2286, opened 2023-07-06
Rework the share modal

max-moser:
Closes  https://github.com/inveniosoftware/invenio-app-rdm/issues/2029.

This PR updates the "share modal" to a tabbed modal with three tabs according to the [access requests RFC](https://codimd.web.cern.ch/U6_FMlwPS9ShGVtYDBjIUQ?view#Share-modal-People-and-groups-tab):
* Access grants (shared with users & groups)
* Shared links
* Access request settings

It allows management of _all_ (i.e. multiple) secret links that have been generated for a record rather than just the first one.
Further, it allows sharing access to the record with specific users and groups in the system via so-called grants.


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -0,0 +1,66 @@
...
+      pane: (
+        <Tab.Pane className="borderless" attached="top" key="accessLinks">
+          <LinksTab record={record} accessLinksSearchConfig={accessLinksSearchConfig} />
+        </Tab.Pane>
+      ),
+    },
+  ];
+
+  return (
+    <Modal
```
jennur, line 28: **A11y**
The modal trigger button should have
```
aria-expanded={open}
aria-haspopup="dialog"
aria-controls={modalId},
```
and the modal
```
role="dialog"
id={modalId}
```
The first interactive element inside the modal should get focus when opening. (If I understand correctly, you have already implemented this, just make sure to test that it works as expected when using the keyboard.)

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/modal.overrides
```diff
@@ -126,3 +126,10 @@
     border-color: @mutedTextColor transparent transparent;
   }
 }
+
+
+.access-link-modal.scrolling.content {
+  overflow-y: auto !important;
+  overflow-x: hidden !important;
```
jennur, line 133: Instead of hiding the x-overflow, does it work to e.g. set the `max-width` of the table to `100%`? To make sure we don't crop content.

PR #2319, opened 2023-07-24
ui: render access request form based on settings

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1374

Case 1: `accept_conditions_text = null` or absent
Screenshot 2023-07-24 at 10 42 32

Case 2: `accept_conditions_text = "Buy me a unicorn"`
Screenshot 2023-07-24 at 10 42 54

Case 3: `accept_conditions_text = "<h1>Buy me a unicorn</h1>"`
Screenshot 2023-07-24 at 10 43 20

Case 4: `allow_user_requests` or `allow_guest_requests` is set to false:
Screenshot 2023-07-24 at 10 45 00

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -24,9 +24,10 @@
 {%- set is_preview_submission_request = preview_submission_request or false %}
 {%- set show_record_management_menu = can_manage_record and (not is_preview or is_preview_submission_request) %}
 
-{#- TODO check if the record allows requests for users/guests #}
-{%- set user_request_enabled = not current_user.is_anonymous %}
-{%- set guest_request_enabled = current_user.is_anonymous %}
+{%- if record.parent.access.settings %}
+  {%- set allow_user_requests = not current_user.is_anonymous and record.parent.access.settings.allow_user_requests %}
+  {%- set allow_guest_requests = current_user.is_anonymous and record.parent.access.settings.allow_guest_requests %}
+{%- endif %}
 
 {%- block page_body %}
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -256,16 +257,24 @@ <h2 id="files-heading">{{ _('Files') }}</h2>
...
-                                  <!-- TODO: Once implemented in the backend display the access description provided nad use this as fallback -->
-                                  If you would like to request access to these files, please fill out the form below.
+                                  {{ _("If you would like to request access to these files, please fill out the form below.") }}
                                 </p>
+                                <div>
+                                  {%- if record.parent.access.settings %}{%- set accept_conditions_text = record.parent.access.settings.accept_conditions_text %}{%- endif %}
+                                  {%- if accept_conditions_text %}
+                                    <p>
+                                      <h5>{{ _("You need to satisfy these conditions in order for this request to be accepted:") }}</h5>
+                                      {{ _(accept_conditions_text) | safe }}
```
zzacharo, line 274: this is redundant I think because it will not mark the resolved text as translatable but rather the variable `accept_conditions_text` but maybe to be checked

zzacharo (approved):

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
anikachurilova: Saw the example here: https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/header.html#L33
But you're right, seems like we can't translate everything that user can put into that field

zzacharo (approved):

PR #2336, opened 2023-08-07
decorator: updates pass_record_or_draft

jrcastro2:
* allows to pass the parent id and returns the latest version
* requires https://github.com/inveniosoftware/invenio-drafts-resources/pull/240
* closes https://github.com/zenodo/rdm-project/issues/174

### DIFF
--- a/invenio_app_rdm/records_ui/views/decorators.py
+++ b/invenio_app_rdm/records_ui/views/decorators.py
@@ -11,7 +11,7 @@
 
 from functools import wraps
 
-from flask import g, request
+from flask import g, redirect, request, url_for
 from invenio_communities.communities.resources.serializer import (
     UICommunityJSONSerializer,
 )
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -128,10 +128,30 @@ def view(**kwargs):
                 try:
                     record = service().read_draft(**read_kwargs)
                 except NoResultFound:
-                    record = service().read(**read_kwargs)
+                    try:
```
ntarocco, line 131: it feels like a lot of duplicated code?

```
record = None
if is_preview:
                try:
                    record = service().read_draft(**read_kwargs)
                except NoResultFound:
                    record = None

if not record:
    try:
                    record = service().read(**read_kwargs)
                except NoResultFound:
                    # If the parent pid is being used we can get the id of the latest record and redirect
                    latest_version = service().read_latest(**read_kwargs)
                    return redirect(
                        url_for(
                            "invenio_app_rdm_records.record_detail",
                            pid_value=latest_version.id,
                        )
                    )
```
I also think that it should never happen we reach this decorator, with a parent_pid and is_preview=True, WDYT?

PR #2340, opened 2023-08-09
views: add signposting link to landing+download response headers

fenekku:
See https://github.com/inveniosoftware/rfcs/pull/73

### DIFF
--- a/invenio_app_rdm/urls.py
+++ b/invenio_app_rdm/urls.py
@@ -1,0 +1,62 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 Northwestern University.
+#
+# This file is lifted from Invenio-RDM-Records .
+#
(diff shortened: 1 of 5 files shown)


invenio_app_rdm/urls.py
```diff
@@ -0,0 +1,63 @@
...
+
+Design decisions:
+
+- Generated urls are absolute for now
+"""
+
+from flask import current_app
+
+
+def record_url_for(_app="ui", pid_value=""):
```
lnielsen, line 34: I have the same comment here as on inveniosoftware/invenio-rdm-records#1404

fenekku: Not sure what can be changed immediately for this case versus the one in invenio-rdm-records, so I just rebased this one.

fenekku:
Rebased. Ready for re-review.

lnielsen (approved):

PR #2351, opened 2023-08-17
jump to top button: move event to button only

jennur:
Landing page jump to top button: 
- Fixes a small but annoying bug where clicking the entire button container (full width) triggers the jump-to-top event. Moves the trigger class to the actual button.

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -366,13 +366,11 @@
             {# Jump up button #}
             {%- block jump -%}
               <div class="ui grid">
-                <div class="jump-to-top centered row rel-mt-1" id="jump-btn">
-                  <a style="cursor: pointer;">
-                    <button class="ui button labeled icon" aria-label="{{ _('Jump to top of page') }}">
-                      <i class="arrow alternate circle up outline icon"></i>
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -366,13 +366,11 @@ <h2 id="files-heading">{{ _('Media files') }}</h2>
...
               <div class="ui grid">
-                <div class="jump-to-top centered row rel-mt-1" id="jump-btn">
-                  <a style="cursor: pointer;">
-                    <button class="ui button labeled icon" aria-label="{{ _('Jump to top of page') }}">
-                      <i class="arrow alternate circle up outline icon"></i>
-                      {{ _("Jump up") }}
-                    </button>
-                  </a>
+                <div class="centered row rel-mt-1">
+                  <button id="jump-btn" class="jump-to-top ui button labeled icon" aria-label="{{ _('Jump to top of page') }}">
```
kpsherva, line 370: minor: do we always have only one jump up button per page ? otherwise I wouldn't assign the js event to element id, since element id should be unique

jennur: The js event is assigned to the class. (But we also just have one button per page.)

kpsherva (approved):

PR #2363, opened 2023-08-25
Record deletion: Display tombstone page

max-moser:
Requires https://github.com/inveniosoftware/invenio-rdm-records/pull/1428

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/tombstone.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/tombstone.html
@@ -1,5 +1,6 @@
 {#
   Copyright (C) 2021 CERN.
+  Copyright (C) 2023 TU Wien.
 
   Invenio App RDM is free software; you can redistribute it and/or modify it
   under the terms of the MIT License; see LICENSE file for more details.
@@ -7,18 +8,41 @@
 {% extends config.THEME_ERROR_TEMPLATE %}
 
 {%- set title = _("Tombstone") + " | " + config.THEME_SITENAME -%}
+{%- set removal_reason = _("Reason unknown") %}
+{%- if record.ui.tombstone.removal_reason %}
+  {%- set removal_reason = record.ui.tombstone.removal_reason.title_l10n %}
+{%- endif %}
 
-{% block message %}
-  <h1><i class="fa fa-flash"></i> {{_('Gone')}}</h1>
+{%- block message %}
+  <h1><i class="fa fa-flash"></i> {{ _('Gone') }}</h1>
   <p>
-  {% trans sitename=config.THEME_SITENAME %}
-    The record  you are trying to access was removed from {{sitename}}. The
+  {%- trans sitename=config.THEME_SITENAME %}
+    The record  you are trying to access was removed from {{ sitename }}. The
     metadata of the record is kept for archival purposes.
-  {% endtrans %}
+  {%- endtrans %}
   </p>
-  {% if record and record.removal_reason %}
-  <p>
-    <strong>{{ _("Reason for removal:") }}</strong> {{record.removal_reason}}
-  </p>
-  {% endif %}
+  <div class="ui segment">
+    <p>
+      <strong>{{ _("Reason for removal:") }}</strong> {{ removal_reason }}
+    </p>
+    <p>
+      <strong>{{ _("Removed by:") }}</strong> {{ record.ui.tombstone.removed_by }}
+    </p>
+    {%- if record.ui.tombstone.note %}
+    <p>
+      <strong>{{ _("Removal note:") }}</strong> {{ record.ui.tombstone.note }}
+    </p>
+    {%- endif %}
+    <p>
+      <strong>{{ _("Date of removal:") }}</strong> {{ record.ui.tombstone.removal_date_l10n_long }}
+    </p>
+    <p>
+      <strong>{{ _("Citation:") }}</strong> {{ record.ui.tombstone.citation_text }}
+    </p>
+    {%- if record.pids.doi %}
+    <p>
+      <strong>{{ _("Identifier:") }}</strong> {{ record.pids.doi.identifier }}
+    </p>
+    {%- endif %}
+  </div>
 {% endblock message %}
--- a/invenio_app_rdm/records_ui/views/records.py
+++ b/invenio_app_rdm/records_ui/views/records.py
@@ -2,7 +2,7 @@
 #
 # Copyright (C) 2019-2021 CERN.
 # Copyright (C) 2019-2021 Northwestern University.
-# Copyright (C)      2021 TU Wien.
+# Copyright (C) 2021-2023 TU Wien.
 #
 # Invenio App RDM is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -279,7 +279,27 @@ def not_found_error(error):
 
 def record_tombstone_error(error):
     """Tombstone page."""
-    return render_template("invenio_app_rdm/records/tombstone.html"), 410
+    # the RecordDeletedError will have the following properties,
+    # while the PIDDeletedError won't
+    record = getattr(error, "record", None)
+    if (record_ui := getattr(error, "result_item", None)) is not None:
+        if record is None:
+            record = record_ui._record
+
+        record_ui = UIJSONSerializer().dump_obj(record_ui.to_dict())
+
+    # render a 404 page if the tombstone isn't visible
+    if not record.tombstone.is_visible:
+        return not_found_error(error)
+
+    # we only render a tombstone page if there is a record with a visible tombstone
+    return (
+        render_template(
+            "invenio_app_rdm/records/tombstone.html",
+            record=record_ui,
+        ),
+        410,
+    )
 
 
 def record_permission_denied_error(error):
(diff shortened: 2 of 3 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/tombstone.html
```diff
@@ -8,17 +8,46 @@
...
+  {%- endtrans %}
   </p>
-  {% if record and record.removal_reason %}
-  <p>
-    <strong>{{ _("Reason for removal:") }}</strong> {{record.removal_reason}}
-  </p>
-  {% endif %}
+  <div class="ui segment">
+  {%- if record and record.tombstone and record.tombstone.is_visible %}
+    {%- set removal_reason = record.tombstone.removal_reason.title.en or "None" %}
```
kpsherva, line 21:
```suggestion
    {%- set removal_reason = record.tombstone.removal_reason.title.en or "Reason unknown" %}
```
wouldn't it be better for user experience ?

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -279,7 +279,20 @@ def not_found_error(error):
...
 def record_tombstone_error(error):
     """Tombstone page."""
-    return render_template("invenio_app_rdm/records/tombstone.html"), 410
+    # the RecordDeletedError will have the following properties,
+    # while the PIDDeletedError won't
+    record = getattr(error, "result_item", None)
+    raw_record = getattr(error, "record", None)
+
+    if raw_record and raw_record.tombstone and not raw_record.tombstone.is_visible:
+        return not_found_error(error)
```
kpsherva, line 288:
```suggestion
        raise not_found_error(error)
```

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/tombstone.html
```diff
@@ -8,17 +8,46 @@
...
+    {%- if record.tombstone.removed_by.user == "system" %}
+      {%- set removed_by = "System (automatic)" %}
+    {%- elif record.tombstone.removed_by.user %}
+      {%- set removed_by = "User" %}
+    {%- else %}
+      {%- set removed_by = "Unknown" %}
+    {%- endif %}
+    <p>
```
kpsherva, line 29: could you provide a screenshot to see how it displays?

invenio_app_rdm/records_ui/views/records.py
max-moser: just as a bit of context:
the `not_found_error` i'm using here is the function defined a few lines above, which renders the 404 template and not an exception.
this view function (`record_tombstone_error`) is called when the error handler catches a `RecordDeletedException` already, so raising another exception here to go back to the error handler might be a bit complicated.

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/tombstone.html
```diff
@@ -8,17 +8,46 @@
...
-  {% endtrans %}
+  {%- endtrans %}
   </p>
-  {% if record and record.removal_reason %}
-  <p>
-    <strong>{{ _("Reason for removal:") }}</strong> {{record.removal_reason}}
-  </p>
-  {% endif %}
+  <div class="ui segment">
+  {%- if record and record.tombstone and record.tombstone.is_visible %}
```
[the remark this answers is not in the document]
max-moser: good point!

i've updated the check in the tombstone page, since it was not logically equivalent to the negation of this check here though.
the main difference was the case if there _is_ a (deleted) record that _does not have_ tombstone information.
previously, this would cause the `tombstone.html` template to be rendered, but only the the "heading" part.
now, this case will result in a 404.

note, however:
with the normal flow, a record should never get into the "deleted" status without having tombstone information attached.
so maybe you would actually prefer me to remove the checks for `if record and record.tombstone`, as they potentially mask deeper issues?

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/tombstone.html
max-moser: yes, that's definitely a clearer text :smile:

max-moser: i've wrapped the hard-coded string in a translation call, too.

however, i'm not sure how i would go about displaying the "correct" translation for the vocabulary title, if set?
right now the english value `record.tombstone.removal_reason.title.en` is hard-coded.

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/tombstone.html
max-moser: sure :)

max-moser:
# L10N update

I figured out that the translations of vocabulary titles is usually done in the UI serializer, and taken the liberty of moving most of that logic into there (now in the RDM-Records PR).
New screenshots:

Record deleted because "duplicate", by user (ID: 1):
image

Record deleted because of unknown reason, by the system (without note):
image

Record deleted by unknown entity:
image

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/tombstone.html
max-moser: found that this is usually done through the UI serializer; see the comment with new screenshots below.

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -279,7 +279,27 @@ def not_found_error(error):
...
+    # while the PIDDeletedError won't
+    record = getattr(error, "record", None)
+    if (record_ui := getattr(error, "result_item", None)) is not None:
+        if record is None:
+            record = record_ui._record
+
+        record_ui = UIJSONSerializer().dump_obj(record_ui.to_dict())
+
+    # render a 404 page if the tombstone isn't visible
+    if not record.tombstone.is_visible:
```
max-moser, line 292: i've removed the pretty defensive `if not record or not record.tombstone` here, as I understood the InvenioRDM philosophy is more to let unexpected cases crash.
And arriving at this point without a `record`, or with one that doesn't have a tombstone, is indeed very unexpected.

Any thoughts about that?
Ping @ntarocco @slint

invenio_app_rdm/records_ui/views/records.py
kpsherva: I think in general it would be nicer to handle it with the handler, not sure what others think

then I would propose to change the name of the function, otherwise it is suggesting we deal with error/exception

max-moser: I agree with the naming, it really sounds like an error more than a handler function.
I didn't touch any of the naming though, in order to avoid unnecessary churn and stay consistent with what's already there.

About the error raising from the error handler, I personally feel that this would introduce a layer of indirection that makes the logic harder to follow.

max-moser: Further, exceptions raised in the error handlers don't seem to go back up to where the errors are routed to their registered handlers – I just tried raising a `KeyError("kekw")` instead of `return not_found_error(error)`, with the following result:

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mmoser/.local/share/virtualenvs/my-site-OggevyoG/lib/python3.9/site-packages/flask/app.py", line 2552, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/mmoser/.local/share/virtualenvs/my-site-OggevyoG/lib/python3.9/site-packages/werkzeug/middleware/proxy_fix.py", line 187, in __call__
    return self.app(environ, start_response)
  File "/home/mmoser/.local/share/virtualenvs/my-site-OggevyoG/lib/python3.9/site-packages/werkzeug/middleware/dispatcher.py", line 78, in __call__
    return app(environ, start_response)
  File "/home/mmoser/.local/share/virtualenvs/my-site-OggevyoG/lib/python3.9/site-packages/flask/app.py", line 2532, in wsgi_app
    response = self.handle_exception(e)
  File "/home/mmoser/.local/share/virtualenvs/my-site-OggevyoG/lib/python3.9/site-packages/flask/app.py", line 2529, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/mmoser/.local/share/virtualenvs/my-site-OggevyoG/lib/python3.9/site-packages/flask/app.py", line 1825, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/mmoser/.local/share/virtualenvs/my-site-OggevyoG/lib/python3.9/site-packages/flask/app.py", line 1676, in handle_user_exception
    return self.ensure_sync(handler)(e)
  File "/home/mmoser/rdm/modules/invenio-app-rdm/invenio_app_rdm/records_ui/views/records.py", line 293, in record_tombstone_error
    raise KeyError("kekw")
KeyError: 'kekw'
```

Even though there is an [error handler registered for `KeyError`s](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/views/__init__.py#L140).

invenio_app_rdm/records_ui/views/records.py
zzacharo: I would say that if is unexpected then it should not happen thus we will see it failing sooner ;)

PR #2366, opened 2023-08-29
Remove whitespaces as default text

max-moser:
The line break after the `<textarea>` opening tag in the HTML seems to introduce the indent before the closing tag as default text for the request message.

Before:

image

After:

image


max-moser:
Outdated

PR #2375, opened 2023-09-04
menu: fix mobile display

jrcastro2:
* closes https://github.com/zenodo/who-rdm/issues/79

## BEFORE 
Screenshot from 2023-09-04 16-35-49

## NOW
Screenshot from 2023-09-04 16-38-22

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
@@ -157,6 +157,7 @@
 
     &.search-bar {
       padding-right: 1rem;
+      margin-right: 1rem;
     }
 
     &:first-of-type {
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/menu.overrides
```diff
@@ -222,6 +222,8 @@
       z-index: 100;
       padding: 2rem 1.5rem 1.5rem 1.5rem;
       overflow-y: scroll;
+      overflow-x: clip;
```
jennur, line 225: The x overflow seems to be caused by the `rel-mr-1` on the search bar element
Screenshot 2023-09-13 at 17 35 45
Instead of clipping the overflow (risky business), can we maybe remove that class from the element, and rather add the margin to the css like this: 

```
.search-bar {
      margin-right: 1rem;

      @media screen and (max-width: @largestTabletScreen) {
        margin-right: 0;
      }
    }
```

jrcastro2: Right! Thank you very much!

Issue #2374, opened 2023-09-04
A11y errors for burger menu


jennur:
The burger menu needs an accessibility scan. Known issue:

![Screenshot 2023-09-04 at 10 16 23](https://github.com/inveniosoftware/invenio-app-rdm/assets/21052053/a0bd1164-f467-4d67-a945-aa0a94dcb726)

PR #2376, opened 2023-09-04
address accessibility errors on global header and landing page

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2374

- Fixes all errors detected in the global header
- Removes duplicate preview id's from landing page, not needed
- Tested with VoiceOver in Safari & axeDevTools
- No visible changes in the UI


fenekku (approved):
For what our review is worth, we had some version of those changes ourselves and were eyeing those bigger ones. Thanks for working on accessibility lately.

zzacharo (approved):

PR #2390, opened 2023-09-11
landing page: improve a11y for accordions, change add details view

jennur:
Closes https://github.com/zenodo/rdm-project/issues/229

- Removes "Description" header from the UI (the element remains in the DOM but with a new `sr-only` class, to keep it available for screen readers.
- Changes Additional Details view from tabs to list (accordion if sub content longer than 5 items)
    - The accordion is open by default
- Improves accessibility for all files accordions
- Custom fields are linked to search for namespace terms
- External url added for namespace docs
- Added "Notes" description to warning message box
- Added acronym to awards

Screenshot 2023-09-12 at 10 19 50

127 0 0 1_5000_records_wp2dk-aqj34
127 0 0 1_5000_records_kmevf-50k32
127 0 0 1_5000_records_kmevf-50k32_expanded

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
@@ -29,19 +29,28 @@
 
 
 {% macro preview_file_box(file, pid, is_preview, record) %}
-<div class="">
-  <div class="ui accordion panel mb-10 {{record.ui.access_status.id}}" href="#collapsablePreview">
-    <div class="active title trigger panel-heading {{record.ui.access_status.id}} truncated" tabindex="0" aria-label="{{ _('File preview') }}">
-      <span id="preview-file-title">{{file.key}}</span>
-      <i class="ui angle right icon"></i>
-    </div>
-    <div id="collapsablePreview" class="active content pt-0">
+  <div class="ui accordion panel mb-10 {{record.ui.access_status.id}}" href="#files-preview-accordion-panel">
+    <h3 class="active title panel-heading {{ record.ui.access_status.id }} m-0">
+      <div
+        role="button"
+        id="files-preview-accordion-trigger"
+        aria-controls="files-preview-accordion-panel"
+        aria-expanded="true"
+        tabindex="0"
+        class="trigger"
+        aria-label="{{ _('File preview') }}"
+      >
+        <span id="preview-file-title">{{ file.key }}</span>
+        <i class="angle right icon" aria-hidden="true"></i>
+      </div>
+    </h3>
+
+    <div role="region" id="files-preview-accordion-panel" aria-labelledby="files-preview-accordion-trigger" class="active content preview-container pt-0">
       <div>
         {{ preview_file('invenio_app_rdm_records.record_file_preview', pid_value=pid, filename=file.key, is_preview=is_preview) }}
       </div>
     </div>
   </div>
-</div>
 {%- endmacro %}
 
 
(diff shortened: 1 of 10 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
```diff
@@ -113,17 +122,20 @@
...
 {%- set binary_sizes = not config.APP_RDM_DISPLAY_DECIMAL_FILE_SIZES %}
-<div class="">
-  <div class="ui accordion panel mb-10 {{record.ui.access_status.id}}" href="#collapsablePreview">
-    <div class="active title trigger panel-heading {{record.ui.access_status.id}}" tabindex="0">
-      {{ _("Files") }}
-      <small class="text-muted">{% if files %} ({{files|sum(attribute='size')|filesizeformat(binary=binary_sizes)}}){% endif %}</small>
-      <i class="angle right icon"></i>
-    </div>
-    <div class="active content pt-0">
+{%- set files_access = record.access.files if not is_media else record.access.media_files %}
```
zzacharo, line 125: I think @anikachurilova changed this so probably align with her first if this change makes sense?

invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -148,3 +148,29 @@ def truncate_number(value, max_value):
...
+    return namespaces[namespace] + namespace_value
+
+
+def custom_fields_search(field, field_value):
+    """Get custom field search url."""
+    namespace_array = field.split(":")
+    namespace = namespace_array[0]
+    namespaces = current_app.config.get("RDM_NAMESPACES")
+
+    if namespaces[namespace] is None:
```
zzacharo, line 172: I think here `namespaces` can be empty string and this will fail

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
anikachurilova: I removed the line `set files_access` (see the commit https://github.com/inveniosoftware/invenio-app-rdm/commit/47194f416a2379b69675611d9dc2ce85df3dd1f9) because the var `files_access` is not used anywhere. If it's still needed, then the  `is_media` should be put back to the `file_list_box` params (and apply it to where it's used) as it is used there. 
Also, I see some changes to the UI, does it make sense to also apply them to the ` media_file_list_box`? Let me know if you need help with adding media files to the record to test it.

invenio_app_rdm/records_ui/views/filters.py
slint:
```suggestion
    if not namespaces[namespace]:
```

We now try to use `None` rather than `""`, but maybe to keep things safe and backwards compat, let's do the above?

slint:
```suggestion
    if not namespaces.get(namespace):
```

I would also use `namespaces.get()` also, in case we get any weird data in accidentally.

invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -148,3 +148,29 @@ def truncate_number(value, max_value):
...
+        return None
+
+    namespace_string = "\:".join(namespace_array)
+    return f"/search?q=custom_fields.{namespace_string}:{field_value}"
```
slint, line 176:
```suggestion
    return url_for("<search-endpoint>", q=f"custom_fields.{namespace_string}:{field_value}")
```

To also get the URL encoding right.

slint (commented):
LGTM, some minor comments on URL generation and namespace checking

PR #2398, opened 2023-09-12
requests: render request details page empty topic

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/243

### DIFF
--- a/invenio_app_rdm/requests_ui/views/requests.py
+++ b/invenio_app_rdm/requests_ui/views/requests.py
@@ -119,8 +119,9 @@ def user_dashboard_request_view(request, **kwargs):
     request_type = request["type"]
     request_is_accepted = request["status"] == AcceptAction.status_to
 
-    has_record_topic = "record" in request["topic"]
-    has_community_topic = "community" in request["topic"]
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/requests_ui/views/requests.py
```diff
@@ -119,8 +119,9 @@ def user_dashboard_request_view(request, **kwargs):
     request_type = request["type"]
     request_is_accepted = request["status"] == AcceptAction.status_to
 
-    has_record_topic = "record" in request["topic"]
-    has_community_topic = "community" in request["topic"]
+    has_no_topic = request["topic"] is None
```
slint, line 122:
```suggestion
    has_topic = request["topic"] is not None
```

nit: I somewhat prefer "positive" versions of boolean statements (since they can be easily converted to the negative by prefixing `not ...`). If one wanted e.g. to reuse this variable to say "if has topic", it would mean reading out a double-negative `not has_no_topic` (like below :) ).

slint (commented):
LGTM, one nit on naming for better readability

PR #2401, opened 2023-09-13
Update robots.txt

max-moser:
Update the `robots.txt` file to discourage bots from crawling some routes that aren't relevant for them, but present in all InvenioRDM deployments.
Some of the routes may be a bit overkill because they're only available to authenticated users (which should not apply to random web crawlers).

### DIFF
--- a/invenio_app_rdm/theme/static/robots.txt
+++ b/invenio_app_rdm/theme/static/robots.txt
@@ -2,4 +2,11 @@ User-agent: *
 Disallow: /search
 Disallow: /api
 Disallow: /administration
+Disallow: /account
+Disallow: /uploads
+Disallow: /me
+Disallow: /login
+Disallow: /logout
+Disallow: /records/*/files
+Disallow: /records/*/preview
 Crawl-delay: 10


invenio_app_rdm/theme/static/robots.txt
```diff
@@ -2,4 +2,11 @@ User-agent: *
 Disallow: /search
 Disallow: /api
 Disallow: /administration
+Disallow: /account
+Disallow: /uploads
+Disallow: /me
```
max-moser, line 7: These are generally only available when the user is authenticated:
* `/account/*`: profile settings, tokens, etc.
* `/me/*`: the user dashboard
* `/uploads/*`: the deposit form

invenio_app_rdm/theme/static/robots.txt
```diff
@@ -2,4 +2,11 @@ User-agent: *
...
+Disallow: /login
+Disallow: /logout
+Disallow: /records/*/files
+Disallow: /records/*/preview
```
max-moser, line 11: These are probably the most relevant routes here, also for unauthenticated guests.

slint (approved):

invenio_app_rdm/theme/static/robots.txt
slint: Google datasets actually does also index some [files for preview](https://datasetsearch.research.google.com/search?src=3&query=canada%20water%20quality&docid=L2cvMTFqbnoyanRwNA%3D%3D) (e.g. CSVs). I'll open a PR to remove this line specifically.

Issue #2416, opened 2023-09-15
Delete community modal styling


jennur:
![image](https://github.com/inveniosoftware/invenio-app-rdm/assets/21052053/c688412b-9d79-4275-a4ff-6bef67a6b277)

PR #2417, opened 2023-09-15
theme: add more specific selector to .ui.modal .content to prevent in…

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2416

- Adding `>` selector as the `.content` styling was inherited by e.g. `.message .content` which is not desired.
- A11y and styling fixes
Screenshot 2023-09-15 at 17 32 16

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/modal.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/modal.overrides
@@ -115,25 +115,27 @@
     border-top: @actionBorder;
     text-align: @actionAlign;
   }
+
   .content {
-  display: block;
-  width: 100%;
-  font-size: @contentFontSize;
-  line-height: @contentLineHeight;
-  padding: @contentPadding;
-  background: @contentBackground;
+    display: block;
+    width: 100%;
+    font-size: @contentFontSize;
+    line-height: @contentLineHeight;
+    padding: @contentPadding;
+    background: @contentBackground;
+  }
 }
-}
+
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/modal.overrides
```diff
@@ -115,19 +115,22 @@
...
-  .content {
-  display: block;
-  width: 100%;
-  font-size: @contentFontSize;
-  line-height: @contentLineHeight;
-  padding: @contentPadding;
-  background: @contentBackground;
-}
+
+  & > .content {
```
kpsherva, line 119: this was done on purpose, it will break other modals, check the comment in line 110.
by doing this change you make it identical with original semantic ui code

jennur: Do you know what it was fixing? Because it's causing a lot of inheritance issues in the modals, since `.content`is also a sub class of other classes.

jennur:
## Updates

- The previous change was breaking some of the modals, and the intended SUI element nesting could not be maintained in these cases, so I have reverted the css changes.
- One suggestion was to add `:not(.message)` to the `.ui.modal .content` selector, but this simply wouldn't pick up  unless I specified `.ui.modal **.ui**:not(.message) .content`, so I decided to rather remove the `<Message.Content>`, as this doesn't affect the styling. I.e. the UI looks the same as in the posted screenshots.

kpsherva:
LGTM!

kpsherva (approved):

PR #2453, opened 2023-09-26
citation: fix citation display for soft deleted records

jrcastro2:


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCitationField.js
```diff
@@ -52,14 +52,15 @@ export class RecordCitationField extends Component {
...
-    return await axios(
-      `${record.links.self}?locale=${navigator.language}&style=${style}`,
-      {
-        headers: {
-          Accept: "text/x-bibliography",
-        },
-      }
-    );
+    const includeDeleted = record.deletion_status.is_deleted
+      ? "&include_deleted=1"
```
kpsherva, line 56: I would try to avoid hardcoding it to be set to true. What if we don't pass it in the landing page view ? Why would it be set here? Maybe we can see if the original view has the argument in the URL somehow?

jrcastro2: It should not be possible to arrive at this cmp with a deleted record without passing `include_delete` param in the URL. However, I agree that it is better to get it from the view, I have adressed this. I am not sure how we can  avoid hardcoding the parameter in the URL

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCitationField.js
```diff
@@ -51,25 +51,26 @@ export class RecordCitationField extends Component {
...
-      `${record.links.self}?locale=${navigator.language}&style=${style}`,
-      {
-        headers: {
-          Accept: "text/x-bibliography",
-        },
-      }
-    );
+  fetchCitation = async (record, style, includeDeleted) => {
+    const includeDeletedParam = includeDeleted ? "&include_deleted=1" : "";
+    const url = `${record.links.self}?locale=${navigator.language}&style=${style}${includeDeletedParam}`;
```
kpsherva, line 56: did you test this change with both "0" and "1" ? Are they string values?
my feeling is that with this approach if someone passess `include_deleted` with value `"0"` (string), you will likely end up with js evaluation of the string, which will be true. see:
Screenshot 2023-09-27 at 14 00 18

jrcastro2: The value `include_deleted` is parsed in the view by the `pass_include_deleted` decorator which returns a boolean and then passed down to the cmp and the cmp has set the Proptype of `includeDeleted` as `bool`, so I would say that we are quite safe, however someone could reuse the cmp wronlgy (as any other cmp I guess), therefore I have applied Zachs suggetion above.

kpsherva (approved):

PR #2457, opened 2023-09-27
ui: break up long community titles

fenekku:
**Before**
community_overflow

**After**
community_overflow_fixed

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/card.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/card.overrides
@@ -83,6 +83,10 @@
       }
     }
   }
+
+  .content .header {
+    overflow-wrap: break-word;
+  }
 }
 
 @media only screen and (min-width: @largestSmallMonitor) {
@@ -162,5 +166,3 @@ a.ui.card {
 .ui.card > .field > .content > .header + .description  {
   margin-top: @descriptionDistance;
 }
-
-


jennur (commented):
Thanks for fixing! I'm wondering if an even better solution is to add the `truncate-lines-2` class to the header element. Then in the case of line break, we'll take advantage of the extra space available. Not 100% sure it will not cause the same kind of issues, so needs to be tested.

WDYT?

fenekku:
Ah, :thinking: so here is what this gives me by only using `truncate-lines-2`:

image

If I add both, I get the same as the original fix which indeed doesn't display the full title either even though it has enough space. So, hypothetically the `truncate` call: https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/assets/semantic-ui/js/invenio_communities/community/frontpage.js#L79C66-L79C66 could be removed, the word-break kept and the `truncate-lines-2` added to get a header that is at most on 2 lines, uses up most of that 2 line and ends with an ellipsis. Is that what you were thinking of?

Unfortunately, I can't set up an environment to test this with v12 that just works given the high rate of change right now (I tried twice in the past and it's too time intensive for me and with our instance running v11 I don't have the bandwidth).    

We could pause this PR and wait for things to settle or merge it and make a ticket for that nicer change to tackle later. I am fine with either since we were able to apply the css change on our instance.

jennur:
Yes, it's what I meant. It's probably better to just keep this fix for now then!

fenekku:
I had to fix another occurrence of text breaking out of its boundaries  ("Additional Details" section of the record landing page"). This prompted a colleague of mine to ask: 

"why isn't there a blanket `overflow-wrap: break-word;` thrown somewhere up the element hierarchy to cover all these cases?" 

... Yeah, why isn't there? It would solve a lot of these that I am sure will continue to crop up.

jennur:
> I had to fix another occurrence of text breaking out of its boundaries ("Additional Details" section of the record landing page"). This prompted a colleague of mine to ask:
> 
> "why isn't there a blanket `overflow-wrap: break-word;` thrown somewhere up the element hierarchy to cover all these cases?"
> 
> ... Yeah, why isn't there? It would solve a lot of these that I am sure will continue to crop up.

Agreed. I think it would be good to just set it overall since we keep finding new cases,  @kpsherva WDYT

PR #2471, opened 2023-10-04
navbar: add cog menu

jrcastro2:
* closes https://github.com/zenodo/rdm-project/issues/262

Screenshot from 2023-10-06 10-49-00

Screenshot from 2023-10-06 10-48-54

### DIFF
--- a/invenio_app_rdm/records_ui/views/records.py
+++ b/invenio_app_rdm/records_ui/views/records.py
@@ -174,6 +174,9 @@ def record_detail(
     if record is not None and emitter is not None:
         emitter(current_app, record=record._record, via_api=False)
 
+    # NOTE: this should maybe be an expandable field instead
+    record_owner = record._record.parent.access.owner.resolve()
+
     resolved_community, _ = get_record_community(record_ui)
     return render_template(
         current_app.config.get("APP_RDM_RECORD_LANDING_PAGE_TEMPLATE"),
@@ -191,6 +194,7 @@ def record_detail(
                 "review",
                 "view",
                 "media_read_files",
+                "moderate",
             ]
         ),
         custom_fields_ui=custom_fields["ui"],
@@ -200,6 +204,9 @@ def record_detail(
         community=resolved_community,
         external_resources=get_external_resources(record_ui),
         user_avatar=avatar,
+        record_owner_username=record_owner.username
+        if record_owner is not None
+        else None,  # record created with system_identity have not owners e.g demo
     )
 
 
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordManagement.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordManagement.js
@@ -15,6 +15,7 @@ import { ShareButton } from "./ShareOptions/ShareButton";
 import { NewVersionButton } from "@js/invenio_rdm_records/";
 import PropTypes from "prop-types";
 import Overridable from "react-overridable";
+import { ManageButton } from "./ManageButton";
 
 export class RecordManagement extends Component {
   constructor(props) {
@@ -32,10 +33,10 @@ export class RecordManagement extends Component {
       isPreviewSubmissionRequest,
       currentUserId,
       accessLinksSearchConfig,
+      recordOwnerUsername,
     } = this.props;
     const { error } = this.state;
     const { id: recid } = record;
-
     const handleError = (errorMessage) => {
       console.error(errorMessage);
       this.setState({ error: errorMessage });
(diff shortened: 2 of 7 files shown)


invenio_app_rdm/theme/views.py
```diff
@@ -75,6 +90,34 @@ def init_menu():
...
+                "q": f"id:{request.view_args.get('pid_value')}"
+            },
+            visible_when=_can_manage_record,
+        )
+
+        current_menu.submenu("cog.user").register(
+            "administration.users",
+            _("Manage user"),
+            order=2,
+            endpoint_arguments_constructor=lambda: {"q": f"id:{current_user.id}"},
```
jrcastro2, line 107: Comment for reviewers, in the issue is specified to use the email. However if the preferences of the visibility are not set to public we cannot find the user by using the email, therefore I added the id that will always find the user (when searching as admin)

invenio_app_rdm/theme/views.py
```diff
@@ -75,6 +90,34 @@ def init_menu():
...
+            visible_when=_can_manage_record,
+        )
+
+        current_menu.submenu("cog.community").register(
+            "administration.communities",
+            _("Manage community"),
+            order=3,
+            endpoint_arguments_constructor=lambda: {
+                "q": f"slug:{request.view_args.get('pid_value')}"
```
jrcastro2, line 116: Comment for reviewers: In the [issue](https://github.com/zenodo/rdm-project/issues/262) is specified to user UUID, however in the request we do only have access to the slug, to get more infor we would need to resolve the community, if I am not mistaken. Since the slug is unique we should be good IMHO.

invenio_app_rdm/theme/views.py
```diff
@@ -29,6 +30,20 @@ def create_url_rule(rule, default_view_func):
...
+    """Function used to check if a user has permissions to manage a community."""
+    return administration_permission.can() and request.url_rule.rule.startswith(
+        "/communities/"
+    )
+
+
+def _can_manage_record():
+    """Function used to check if a user has permissions to manage a record."""
+    return administration_permission.can() and request.url_rule.rule.startswith(
+        "/records/"
```
kpsherva, line 43: this will add it to search pages as well if I am not mistaken? is this what we want?

jrcastro2: It should only be displayed when we have the context of a record, which if I am not mistaken happens all the time under `/records/`. I couldn't find any search page that starts with `/records/`.
The general search starts with `/search` and the community records search starts with `/communities/`.
If you find any please share it and I will think of alternatives.

invenio_app_rdm/theme/views.py
jrcastro2: We can see all the users in the admin page but we cannot search by email unless they have set the Profile visibility to Public.

## List all users
Screenshot from 2023-10-05 10-41-33

## Search for user with Public profile
Screenshot from 2023-10-05 10-41-53

## Search for user with hidden profile
Screenshot from 2023-10-05 10-42-02

kpsherva: Nice finding! thanks for checking this. 

in this case I think we are just missing User manager generator (or permission?) in a right place for searching emails:

https://github.com/inveniosoftware/invenio-users-resources/blob/master/invenio_users_resources/services/permissions.py#L39

due to https://github.com/inveniosoftware/invenio-users-resources/blob/dbb9126b725b26501f4921a7d73fe8e15a3d2ea8/invenio_users_resources/services/schemas.py#L48

I think it is better to handle it by permissions rather than trying to circumvent it via ID. maybe makes sense to discuss during the standup?

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -126,6 +126,7 @@ def record_detail(
     pid_value, record, files, media_files, is_preview=False, include_deleted=False
 ):
     """Record detail page (aka landing page)."""
+    record_owner = current_rdm_records.records_service.get_owner(pid_value)
```
jrcastro2, line 129: Comment for reviewers:This returns the User model object which has no dict representation. As I commented [here](https://github.com/inveniosoftware/invenio-rdm-records/pull/1519/files#diff-d79cbaf668935422b2743d81ca9906d9b72ac8b3456c2280ba5a5edacb72796bR157), maybe we should consider creating an interface for the owner as it could potentially be something different than a user.

invenio_app_rdm/theme/views.py
jrcastro2: The email is not indexed if the user profile is set to hidden, therefore no one can search for it as it's not available, agreed over [chat](https://discord.com/channels/692989811736182844/1092367535141289984/1159802012288958494) that we will use the username instead, which unique and indexed all the time and only admins can search for it.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordManagement.js
```diff
@@ -32,19 +33,24 @@ export class RecordManagement extends Component {
...
-
     const handleError = (errorMessage) => {
       console.error(errorMessage);
       this.setState({ error: errorMessage });
     };
 
     return (
       <Grid columns={1} className="record-management">
-        {permissions.can_edit && !isDraft && (
+        {permissions.can_manage && (
```
zzacharo, line 47: shouldn't be this `can_manage_communities`? Are we exposing the manage button to record owners/curators ?

invenio_app_rdm/records_ui/views/records.py
zzacharo: record is already resolved here, can't we resolve the owner directly?

PR #2504, opened 2023-10-19
access settings: populate parent with default settings if not present

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/392

The flow for records: on render of landing page, if `record.access.settings` are empty, they are getting prefilled with default values. **Not saved to db yet**. In case of user managing settings and clicking the "Save" button - all the data get's sent to the db.

The flow for drafts: settings are getting populated on publish of a draft even without this fix


kpsherva (approved):
Approved as temp solution

PR #2517, opened 2023-10-26
views: retrieve published record if inclusion request

zzacharo:
closes https://github.com/zenodo/rdm-project/issues/467

### DIFF
--- a/invenio_app_rdm/requests_ui/views/requests.py
+++ b/invenio_app_rdm/requests_ui/views/requests.py
@@ -49,9 +49,19 @@ def _resolve_topic_record(request):
     # parse the topic field to get the draft/record pid `record:abcd-efgh`
     entity = ResolverRegistry.resolve_entity_proxy(request["topic"])
     pid = entity._parse_ref_dict_id()
+
+    request_type = request["type"]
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/requests_ui/views/requests.py
```diff
@@ -84,28 +94,40 @@ def _resolve_topic_record(request):
...
         try:
-            files = draft_files_service().list_files(
-                id_=record_pid, identity=g.identity
-            )
+            if is_record_inclusion:
+                files = files_service().list_files(id_=record_pid, identity=g.identity)
+            else:
+                files = draft_files_service().list_files(
+                    id_=record_pid, identity=g.identity
+                )
```
slint, line 109: minor/shelve: it feels like the logic is a bit duplicated and disconnected here. It should be that based on the `record`'s type, we have a way to fetch the files (without having to depend on the request to determine what the `record` is). In any case, let's shelve and refactor later.

slint (approved):
LGTM, some concerns about the duplicate/disconnected switch logic on the request type, but should be addressed at https://github.com/zenodo/rdm-project/issues/482.

PR #2532, opened 2023-11-08
UserActions: Fix wrong endpoint for drafts + UI improvements

jennur:
Closes https://github.com/zenodo/rdm-project/issues/500 

- UI improvements (default input in GB, toggle to use bytes), more info displayed in modals.
- Updating apiUrl for record drafts.

Screenshot 2023-11-08 at 2 30 51 PM
Screenshot 2023-11-08 at 2 31 01 PM
Screenshot 2023-11-08 at 2 30 40 PM

Screenshot 2023-11-08 at 3 09 17 PM
Screenshot 2023-11-08 at 3 09 23 PM
Screenshot 2023-11-08 at 3 09 35 PM

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/SetQuotaForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/SetQuotaForm.js
@@ -44,17 +44,34 @@ export class SetQuotaForm extends Component {
   handleSubmit = async (values) => {
     this.setState({ loading: true });
     const { addNotification } = this.context;
-    const { resource, actionSuccessCallback, apiUrl } = this.props;
-    const payload = { ...values };
-    this.cancellableAction = withCancel(http.post(apiUrl, payload));
+    const { resource, actionSuccessCallback, setQuotaInBytes, apiUrl } = this.props;
+    let { notes, quota_size: quotaSize, max_file_size: maxFileSize } = values;
+
+    const quotaInGb = !setQuotaInBytes && quotaSize;
+
+    if (!setQuotaInBytes) {
+      quotaSize *= 10 ** 9;
+      maxFileSize *= 10 ** 9;
+    }
+
+    this.cancellableAction = withCancel(
+      http.post(apiUrl, {
+        notes,
+        quota_size: quotaSize,
+        max_file_size: maxFileSize,
+      })
+    );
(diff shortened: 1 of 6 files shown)


Samk13:
Closes [#1526](https://github.com/inveniosoftware/invenio-rdm-records/issues/1526)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserActions.js
```diff
@@ -101,11 +101,18 @@ export class UserActions extends Component {
...
       displayQuota,
       useDropdown,
+      recordId,
     } = this.props;
     const { loading } = this.state;
     const isUserBlocked = !isEmpty(user.blocked_at);
     const isUserActive = user.active;
     const isUserVerified = !isEmpty(user.verified_at);
+    const apiUrl = recordId
+      ? `/api/records/${recordId}/quota`
```
kpsherva, line 111: Since the component is tied to the user, it should not be aware of a record.
I think this is the moment to split the functionalities and either
1) have a separate very similar component `RecordActions` where we implement `<SetQuotaAction` and `<UserActions` returned in render function
OR: 2) we take out the `<Set quota` from this component and we place it directly in respective SearchResultItemLayout or users and record

I would prefer to avoid adding more responsibilities to `UserActions` since we aim to have single responsibility components and it is already quite overloaded

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/SetQuotaForm.js
```diff
@@ -44,17 +44,33 @@ export class SetQuotaForm extends Component {
   handleSubmit = async (values) => {
     this.setState({ loading: true });
     const { addNotification } = this.context;
-    const { resource, actionSuccessCallback, apiUrl } = this.props;
-    const payload = { ...values };
-    this.cancellableAction = withCancel(http.post(apiUrl, payload));
+    const { resource, actionSuccessCallback, setQuotaInBytes, apiUrl } = this.props;
+    let { notes, quota_size, max_file_size } = values;
+    let quota_in_gb = !setQuotaInBytes && quota_size;
```
kpsherva, line 49: eslint unhappy

kpsherva (approved):

PR #2534, opened 2023-11-10
record details: handle files in pending/failed state

anikachurilova:
* don't include render of pending/failed files and media files in landing page and in request view
* closes https://github.com/zenodo/rdm-project/issues/471

all the files are broken:
Screenshot 2023-11-10 at 16 54 07
Screenshot 2023-11-10 at 16 56 45

1/2 file is broken - shows only healthy files
Screenshot 2023-11-10 at 16 56 13


invenio_app_rdm/requests_ui/views/requests.py
```diff
@@ -109,7 +109,15 @@ def _resolve_record_or_draft_files(record, request):
...
-        return files.to_dict()
+
+        files_dict = None if files is None else files.to_dict()
+        if files_dict is not None:
+            files_dict["entries"] = [
+                entry
+                for entry in files_dict.get("entries", [])
+                if entry.get("status", None) == "completed"
+            ]
+        return files_dict
```
slint, line 120: minor: there are a couple of ways to think/design this filtering of files:

- mid-level service: we pass a `status=...` parameter to `FileService.list_files`, which will do the filtering
  - this could also go "lower" by propagating the filtering to `FilesField` (to avoid fetching more files than necessary).
- presentation: we filter in the template using a Jinja filter. This place here is also presentation-layer technically, but I would prefer to have it in Jinja so that it doesn't feel ad-hoc for each view that deals with files.

@kpsherva any other ideas on where to handle this? I think going too low-level (e.g. on the `FilesField` and files manager) wouldn't be easy to propagate through all the layers, and we're already looking into bigger changes now with all the edit/publish/delete optimizations.

slint (commented):
Some comments on where/how the filtering should take place.

slint (approved):

PR #2538, opened 2023-11-16
ui: add mathjax to record details page

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/548

Landing page:
Screenshot 2023-11-15 at 17 07 02

Request page:
Screenshot 2023-11-15 at 17 07 40

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -360,6 +360,9 @@
 {%- endblock page_body %}
 
 {%- block javascript %}
+  {% if config.THEME_MATHJAX_CDN %}
+    <script type="text/javascript" src="{{ config.THEME_MATHJAX_CDN }}"></script>
+  {% endif %}
   {% if use_theme_basic_template|default(true) %}
     {{ super() }}
   {% else %}


invenio_app_rdm/config.py
```diff
@@ -234,6 +234,11 @@ def _(x):
 
 THEME_JAVASCRIPT_TEMPLATE = "invenio_app_rdm/javascript.html"
 
+THEME_MATHJAX_CDN = (
```
kpsherva, line 237: was it agreed on to use the cdn in the invenioRDM core? should we use it only in zenodo? to be discussed.

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -360,6 +360,7 @@ <h4>{{ _("Request access") }}</h4>
 {%- endblock page_body %}
 
 {%- block javascript %}
+  <script type="text/javascript" src="{{ config.THEME_MATHJAX_CDN }}"></script>
```
slint, line 363:
```suggestion
  {% if config.THEME_MATHJAX_CDN %}
  <script type="text/javascript" src="{{ config.THEME_MATHJAX_CDN }}"></script>
  {% endif %}
```

As suggested above.

slint: Now that I think about it, shouldn't this bee already defined like this in `invenio-theme`? Where did the `THEME_` prefix come from?

anikachurilova: There's  mathjax defined in `invenio-theme`. Took this constant name from the legacy https://github.com/zenodo/zenodo/blob/master/zenodo/config.py#L1427

PR #2548, opened 2023-12-06
search: signal if more than 3 authors

### LINKED ISSUE #1873: Truncated creator name list in search results
The search results truncate to the first three creators. There is no indication that there are other creators on the record. This is particularly confusing when you have a record with four creators, since it looks like data is missing.

An indication that other creators are present but not listed. The simplest would be "..." linked to the record landing page

## Describe alternatives you've considered
Showing more creators and wrapping is possible but tricky on small screens. "Et al." is the traditional approach, but not sure whether we have to think about translation.

## Additional context
Mockup:
Screen Shot 2022-09-14 at 9 49 15 AM

fenekku:
- closes #1873

**Screenshot (most recent)**

search_and_other_s_v2

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
@@ -27,7 +27,8 @@ class RecordsResultsListItem extends Component {
       "ui.created_date_l10n_long",
       "No creation date found."
     );
-    const creators = result.ui.creators.creators.slice(0, 3);
+
+    const creators = result.ui.creators.creators;
 
     const descriptionStripped = _get(
       result,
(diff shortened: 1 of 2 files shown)


tmorrell:
Awesome! Could we get a screenshot? I'm nervous that "and other(s)" might get a bit long

fenekku:
Oups sorry I took the screenshot but forgot to add it and add some discussion here :sweat_smile: .

So I tried with `...` too and it didn't stand out clearly enough (a bit too small and squeezed) - felt like a rendering artifact. I tried `more` and `et al.`, but `more` wasn't quite the right word in the end and `et al.` is typically strictly for "and others" plural, there is the translation thing your mentioned, and perhaps it's a bit too high brow.

This morning I just thought about a more nuanced version, which might be better, which is the "counted others". Something like:

for 1 more author

image

for more than 1 more author:

image

I saw translation mechanisms that should allow this. I will change it and wait for more feedback.

tmorrell:
That looks really nice! One thought-could you link it to the record landing page? Since the other authors are linked it would be nice if the "n others" led you to where you see who they are.

fenekku:
> That looks really nice! One thought-could you link it to the record landing page? Since the other authors are linked it would be nice if the "n others" led you to where you see who they are.

Yes, that makes sense!

fenekku:
Done :santa:

jennur (approved):
LGTM! Thanks for fixing this!

PR #2581, opened 2024-02-15
theme: add some css classes to display communities list on the record

anikachurilova:
* closes https://github.com/zenodo/zenodo-rdm/issues/715

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
@@ -462,6 +462,12 @@ dl.details-list {
   margin-left: .1rem;
 }
 
+.comma-separated:not(:last-child)::after {
+  content: @listCommaSeparator;
+  display: inline-block;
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -462,6 +484,12 @@ dl.details-list {
   margin-left: .1rem;
 }
 
+.comma_separated:not(:last-child)::after {
```
kpsherva, line 487:
```suggestion
.comma-separated:not(:last-child)::after {
```
please stick to one convention ;)

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -416,6 +416,28 @@ dl.details-list {
   }
 }
 
+.community-link {
```
kpsherva, line 419: this custom CSS should not be needed at all if you are using `<a>` tag

PR #2583, opened 2024-02-16
css: add muted color link style

anikachurilova:
* closes https://github.com/zenodo/zenodo-rdm/issues/727


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides
```diff
@@ -299,6 +299,10 @@ a.no-text-decoration:hover {
   text-decoration: none;
 }
 
+a.no-blue-link-underline:hover {
```
kpsherva, line 302:
```suggestion
a.remove-default-style:hover {
```
just a naming suggestion

PR #2595, opened 2024-02-26
administration: display record actions as dropdown

slint:


ntarocco (approved):

slint:
In the end, I just reverted partially changes made in https://github.com/inveniosoftware/invenio-app-rdm/commit/37b2a255dec5b4cb1fda165c9a8f2900e9347f62. The way actions are currently built requires some planning for composing dropdowns. For now the fix is just to display the set quota button again.

PR #2600, opened 2024-03-04
Bump react invenio forms

ntarocco:


slint (approved):
LGTM! Checked all dependencies that are now pinned to majors and there are no breaking changes. I guess we're fully SemVer compatible now :)

PR #2606, opened 2024-03-08
communities-ui: fix grants metric type

slint:


invenio_app_rdm/communities_ui/views/communities.py
```diff
@@ -71,7 +71,7 @@ def communities_home(pid_value, community, community_ui):
                 "metrics": {
                     "total_grants": {
                         "name": "total_grants",
-                        "type": "value_count",
+                        "type": "cardinality",
```
alejandromumo, line 74: question: what was the issue with `value_count`? Is it a performance thing?

lnielsen: I think it simply count the number of values which means it will count duplicates, where was we need it as s et.
Screenshot 2024-03-08 at 13 22 11

Sandbox has 4 records, but all records are related to the same grant.

alejandromumo (approved):

PR #2623, opened 2024-03-22
records-ui: urlize links in citation text

slint:
If there are DOI links serialized in the citation text, they should be clickable links (cc @tmorrell).

Before:

image

After:

image

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCitationField.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCitationField.js
@@ -7,6 +7,7 @@
 // under the terms of the MIT License; see LICENSE file for more details.
 
 import _debounce from "lodash/debounce";
+import _escape from "lodash/escape";
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import { Placeholder, Dropdown, Message } from "semantic-ui-react";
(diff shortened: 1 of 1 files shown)


tmorrell (approved):
This is awesome! It works great and I've already rolled it into production

slint:
Two fixes:

- HTML-escaping the original citation using [`_.escape`](https://lodash.com/docs/4.17.15#escape)) to prevent XSS attacks
- Updated the regex to avoid matching some punctuation points (dots, commas, semicolons) so that we don't end up with links like <https://doi.org/10.1234/foobar.>

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCitationField.js
```diff
@@ -100,10 +101,21 @@ export class RecordCitationField extends Component {
       };
     });
 
+    // convert links in text to clickable links (ignoring punctuations at the end)
+    const escapedCitation = _escape(citation); // escape html characters
+    const urlRegex = /(https?:\/\/[^\s.,;]+)/g;
```
tmorrell, line 106: This seems to stop too early. `const urlRegex = /(https?:\/\/[^\s,;]+(?=[^\s.,;]*\S))/g;` seems to do the trick.

slint: Thanks, I think I've committed the wrong iteration after playing at https://regex101.com/ 😅

PR #2632, opened 2024-04-03
config: update limiter config

zzacharo:
Update flask-limiter config according to [invenio-app](https://github.com/inveniosoftware/invenio-app/blob/master/invenio_app/ext.py#L123C25-L123C46). The change in the latter keeps backwards compatibility. This PR fixes the ability to be able to use the newly created config as currently invenio-app-rdm sets a default to a local redis storage.


alejandromumo (approved):
LGTM

Issue #2592, opened 2024-02-26
access-requests: add UI tab for groups


anikachurilova:
Manage groups grants in the "Groups" tab of "Share access" modal

PR #2649, opened 2024-04-16
share modal: add groups tab by feature flag

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2592

See most of the screenshots here https://github.com/inveniosoftware/invenio-communities/pull/1127

Some error handlers:
On delete:
Screenshot 2024-04-18 at 16 47 36
Screenshot 2024-04-18 at 16 48 35


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/AccessLinks/LinksSearchItem.js
```diff
@@ -66,67 +78,82 @@ export const LinksSearchItem = ({ result, record, fetchData }) => {
...
-        />
-      </Table.Cell>
-      <Table.Cell width={4}>
-        <Button
-          loading={loading}
-          disabled={loading}
-          onClick={handleDelete}
-          icon
-          labelPosition="left"
+      {error ? (
```
kpsherva, line 81: minor: could we discuss not using ternary operators with big blocks of code? it is a bit unreadable

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -6,59 +6,255 @@
...
+      groupsError,
+    } = this.state;
+    // const usersClient = new UsersApi(); uncomment together with the code below
+    const groupsClient = new GroupsApi();
+    const panes = [
       // hiding user access for until the groups tab is added and fully tested
       // {
-      //   menuItem: { icon: "users", content: "People" },
-      //   pane: (
+      //   menuItem: { icon: "user", content: "People" },
```
kpsherva, line 174: is it still needed to be commented out?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/AccessUsersGroups/AccessUsersGroups.js
```diff
@@ -0,0 +1,123 @@
...
+            icon="exclamation"
+            className="text-align-left"
+            negative
+            size="mini"
+          />
+        )}
+        <Modal.Content className="share-content">
+          {loading && <Loader isLoading active />}
+          {!loading && results !== undefined && (
+            <UserGroupAccessSearchResult
```
kpsherva, line 61: understanding just from props, I have a feeling that this component has too many responsibilities, could it be simplified and divided to several components?
when we have too many responsibilities on one component, it becomes impossible to decouple functionalities and reuse the code. 
Some of the props are redundant, for example `record` and `recOwner`, you can retrieve the owner inside the component using record. callbackEndpoint and endpoint would be potentially another example.
some of the other props are just passed down from the parent component to UserGroupAccessSearchResult, maybe that would be an indicator to split this component to couple of new ones?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/AccessUsersGroups/UserGroupAccessSearchResult.js
```diff
@@ -0,0 +1,313 @@
...
+  };
+
+  render() {
+    const { result, permissions, endpoint, searchType, updatedCallback } = this.props;
+    const { loading, success, error } = this.state;
+    const entity = result.expanded.subject;
+    return (
+      <Table.Row>
+        {error ? (
+          <ErrorMessage
```
kpsherva, line 66: same comment as somewhere before: do we need the error message to be displayed inside the table?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -6,59 +6,255 @@
...
+      results = groupsResults;
+    }
+    if (entityType === this.linksType) {
+      const { linksResults } = this.state;
+      results = linksResults;
+    }
+
+    results.forEach((result) => {
+      if (result.id === id) {
+        result.permission = permission;
```
kpsherva, line 70: just for my understanding on how it works: I am not sure I fully understand why permission is set on the result, shouldn't the permission be already there?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -6,59 +6,255 @@
...
+    if (entityType === this.groupType) {
+      const { groupsResults } = this.state;
+      results = groupsResults;
+    }
+    if (entityType === this.linksType) {
+      const { linksResults } = this.state;
+      results = linksResults;
+    }
+
+    results.forEach((result) => {
```
kpsherva, line 68: if there are many results, could it have issues with performance?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -6,59 +6,255 @@
...
+      results = usersResults;
+    }
+    if (entityType === this.groupType) {
+      const { groupsResults } = this.state;
+      results = groupsResults;
+    }
+    if (entityType === this.linksType) {
+      const { linksResults } = this.state;
+      results = linksResults;
+    }
```
kpsherva, line 66: these ifs could be simplified by using map

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -6,59 +6,255 @@
...
+        result.permission = permission;
+      }
+    });
+  };
+
+  addedOrDeletedCallback = async (callbackEndpoint, entityType) => {
+    if (entityType === this.userType || entityType === this.groupType)
+      await this.fetchUsersGroups(callbackEndpoint, entityType, true);
+    if (entityType === this.linksType) await this.fetchLinks(callbackEndpoint, true);
+  };
```
kpsherva, line 79: for readability of the code we could use `isUserType` and `isLinksType` pattern (precalculated variables) :
```javascript
const isUserGroupType = entityType === this.userType || entityType === this.groupType

const isLinksType = (entityType === this.linksType)

if (isUserGroupType){
...
}
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -6,59 +6,255 @@
...
+
   panes = (record, accessLinksSearchConfig, permissions) => {
-    const { handleClose } = this.props;
-    return [
+    const { handleClose, groupsEnabled } = this.props;
+    const {
+      linksResults,
+      linksLoading,
+      linksError,
+      usersResults,
```
kpsherva, line 162: please check the lint warnings below

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -6,59 +6,255 @@
...
+    });
+  };
+
+  addedOrDeletedCallback = async (callbackEndpoint, entityType) => {
+    if (entityType === this.userType || entityType === this.groupType)
+      await this.fetchUsersGroups(callbackEndpoint, entityType, true);
+    if (entityType === this.linksType) await this.fetchLinks(callbackEndpoint, true);
+  };
+
+  fetchUsersGroups = async (endpoint, entityType, isDataChanged) => {
```
kpsherva, line 81: general observation for this component: it has too many responsibilities, especially fetching different types of data, it should be split into smaller components

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/AccessLinks/LinksSearchItem.js
anikachurilova: I put the message inside the row because it's an error that occurred on deleting a specific secret link. Do you mean we should put the error message on top of the modal?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/AccessLinks/AccessDropdown.js
```diff
@@ -83,4 +84,6 @@ AccessDropdown.propTypes = {
   result: PropTypes.object.isRequired,
   dropdownOptions: PropTypes.array.isRequired,
   updateEndpoint: PropTypes.string.isRequired,
+  entityType: PropTypes.string.isRequired,
+  updatedCallback: PropTypes.func.isRequired,
```
ntarocco, line 88: can I suggest a rename for something like:
```suggestion
  onItemUpdated: PropTypes.func.isRequired,
```
?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/AccessLinks/LinksSearchResultContainer.js
```diff
@@ -174,5 +175,6 @@ export class LinksSearchResultContainer extends Component {
 LinksSearchResultContainer.propTypes = {
   results: PropTypes.array.isRequired,
   record: PropTypes.object.isRequired,
-  fetchData: PropTypes.func.isRequired,
+  addedOrDeletedCallback: PropTypes.func.isRequired,
```
ntarocco, line 178:
```suggestion
  onItemAddedOrDeleted: PropTypes.func.isRequired,
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -6,59 +6,255 @@
...
+    const { record } = this.state;
+    updatedRecord.expanded = record.expanded;
     this.setState({ record: updatedRecord });
   };
 
+  userType = "user";
+  groupType = "group";
+  linksType = "links";
+
+  updatedCallback = (id, permission, entityType) => {
```
ntarocco, line 53: Ok checking its usage, I would rename it to:
```suggestion
  onPermissionChanged = (entityId, entityType, newPermission) => {
```
And I would add a docstring comment to explain what this func is doing, and why

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/modules/modal.overrides
```diff
@@ -1,12 +1,9 @@
 
 .share-modal {
-  .share-content {
-    display: flex;
+  min-height: 35em;
```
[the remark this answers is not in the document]
anikachurilova: `share-content` class was not used anywhere, so no side effect!

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -6,59 +6,255 @@
...
+      await this.fetchUsersGroups(callbackEndpoint, entityType, true);
+    if (entityType === this.linksType) await this.fetchLinks(callbackEndpoint, true);
+  };
+
+  fetchUsersGroups = async (endpoint, entityType, isDataChanged) => {
+    if (entityType === this.userType) await this.fetchUsers(endpoint, isDataChanged);
+    if (entityType === this.groupType) await this.fetchGroups(endpoint, isDataChanged);
+  };
+
+  fetchUsers = async (endpoint, isDataChanged) => {
```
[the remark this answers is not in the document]
anikachurilova: The purpose is to mark that data was added/deleted and the table of results needs refetching. 
`isDataChanged` will be equal to `false` when clicking between tabs for more than 1 time. As a result the whole list is not being fetched every time user clicks on the tab.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
anikachurilova: I don't think we are expecting to have more than ~10-20 results in each tab, so I think this should be ok!

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
anikachurilova: Split the responsibilities between the components! Does it look better now?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/AccessLinks/LinksSearchItem.js
kpsherva: ok I see, that was not obvious from the code, thanks for explaining.
What about ternary operators?

anikachurilova: Replaced ternany operators with 
```
{error && <ErrorMessage />}
{!error && <... />}
```
I agree looks more readable now!

kpsherva:
tests outcome (mainly UX improvements, can be implemented in a separate PR):

1. Active tab label should have brand colour (pinkish), otherwise it looks strange, like inactive tab
Screenshot 2024-04-29 at 16 34 42

2. Owner should be counted in the number on the label, otherwise the numbers don't add up:
Screenshot 2024-04-29 at 16 37 28

3. The permission level field does not stand out, therefore someone can just choose users and click "Add", which results in:
Screenshot 2024-04-29 at 14 54 42
We need to either block the add button, until all required fields are filled or improve the error message, because it does not state which field is missing - the user won't associate permission with `Access` field

4. Search needs improvement, ticket already created https://github.com/CERNDocumentServer/cds-rdm/issues/114

5. When adding multiple users, the search works quite badly, it should favour users not yet added to the access grants: 
Screenshot 2024-04-29 at 14 55 32 ( I was expecting to see user `Test 2`, which showed up only when typing the full name, otherwise the dropdown is taken over by the already added entries)

6. when clicking remove, we should have confirmation monit ("Are you sure you want to remove...?"), like everywhere else. The button should be red, since our actions are color-coded.
Screenshot 2024-04-29 at 16 37 28

7. When clicking on remove, the whole modal goes to loading state, we should only display loading on the specific button, why to block all the rows, is there any specific reason? Otherwise the UI "jumps" due to unspecified minimal height while loader is displaying
Screenshot 2024-04-29 at 16 37 28

PR #2650, opened 2024-04-17
views: fix missing `flash` import

slint:


zzacharo (approved):

slint:
I've also run a `ruff check | grep F821` on all InvenioRDM repos to make sure we haven't missed any other imports after the Flask v2.3 refactor/moves.

ntarocco:
Shall we add `ruff` to CI?

PR #2661, opened 2024-04-27
records-ui: fix trailing dot issue in citation text

slint:
* Closes https://github.com/zenodo/zenodo-rdm/issues/847.


tmorrell (approved):
This works. Trailing dots are such a pain!

PR #2664, opened 2024-05-06
fix: RecordVersionsList fetch versions only once

ptamarit:

### Description

Opening the landing page of any record currently fetches 4 times the list of versions from the API:

---

Versions API being fetched multiple times

---

The problem is due to the `fetchVersions` and `cancellableFetchVersions` (wrongly named `cancellableFetchCitation`) functions being recreated at every render (when the state changes).

This pull request moves the fetching functions inside `useEffect`, so that they are created only within the scope of `useEffect` and don't need to be declared as a dependency. The version is now only fetched once (since the props on which `recordDeserialized.links.versions` and `recid` don't change).

### Checklist

- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).

**Frontend**

**Reminder**

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js
@@ -1,5 +1,5 @@
 // This file is part of InvenioRDM
-// Copyright (C) 2020-2021 CERN.
+// Copyright (C) 2020-2024 CERN.
 // Copyright (C) 2020-2021 Northwestern University.
 // Copyright (C) 2021 Graz University of Technology.
 //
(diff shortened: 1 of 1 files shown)


jrcastro2 (approved):
:rocket: Nice fix!

ptamarit:
> LGTM, I assume passing the correct dependencies to `useEffect` is the main fix :)

* In theory, an empty array means no dependency and no need to re-run the effect again, so it's not 100% clear to me why it was running on every render.
* Had we not disabled the the ESLint warning, we would have had to define the fetching function as a dependency, and then it would have been clear that the `useEffect` would always be triggered due to the function being new at every render.
* Moving the fetch function inside the `useEffect` was enough to fix the problem, which means that the fact that this function was always a new version (and therefore different) was what was triggering the useEffect (even though it was not declared as a dependency).
* I decided to define the dependencies as it did not hurt and is the right thing to do.

slint (approved):

PR #2669, opened 2024-05-10
views: pass draft and record for lock files check

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/852


slint (approved):
LGTM, but it might be worth moving the calculation of `files_locked` outside of the decorator and in the view function where it's used.

PR #2677, opened 2024-05-27
fix 403 tombstone page

anikachurilova:


kpsherva (approved):

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -419,15 +419,16 @@ def record_permission_denied_error(error):
...
-            ),
-            403,
-        )
+    if record:
+        is_restricted = record.get("access", {}).get("record", None) == "restricted"
+        has_doi = "doi" in record.get("pids", {})
+        if is_restricted and has_doi:
+            return (
+                render_template(
+                    "invenio_app_rdm/records/restricted_tombstone.html",
```
ntarocco, line 428: Unrelated: given that this error will be shown only if a DOI exist, I would have called the HTML file something with `doi` in the middle ☺️

ntarocco (approved):

Issue #2684, opened 2024-05-30
secret links: CSRF token not set on editing record as anonymous


anikachurilova:
Steps to reproduce the bug:

1. Create a record
2. Create an edit access level link in the "Share access" modal and copy it
3. Open the link as anonymous user and try to click on "Edit" or "New version" buttons
4. Notice that "CSRF cookie not set" error appears (problem 1)
5. Click on the button again and notice the redirection to login page (problem 2)

PR #2687, opened 2024-05-30
deposit: remove login required from deposit edit

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2684

### DIFF
--- a/invenio_app_rdm/records_ui/views/decorators.py
+++ b/invenio_app_rdm/records_ui/views/decorators.py
@@ -11,7 +11,8 @@
 
 from functools import wraps
 
-from flask import g, make_response, redirect, request, url_for
+from flask import g, make_response, redirect, request, session, url_for
+from flask_login import login_required
 from invenio_communities.communities.resources.serializer import (
     UICommunityJSONSerializer,
 )
@@ -381,3 +382,24 @@ def add_signposting(f):
         return response
 
     return view
+
+
+def secret_link_or_login_required():
+    """Skip login redirection check for requests with secret links.
+
+    If access has been granted via a secret link, then permissions are checked
+    in the dedicated view.
+    """
+
+    def decorator(f):
+        @wraps(f)
+        def view(**kwargs):
+            secret_link_token_arg = "token"
+            session_token = session.get(secret_link_token_arg, None)
+            if session_token is None:
+                login_required(f)
+            return f(**kwargs)
+
+        return view
+
+    return decorator
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -33,7 +33,12 @@ from marshmallow_utils.fields.babel import gettext_from_dict
 from sqlalchemy.orm import load_only
 
 from ..utils import set_default_value
-from .decorators import pass_draft, pass_draft_community, pass_draft_files
+from .decorators import (
+    pass_draft,
+    pass_draft_community,
+    pass_draft_files,
+    secret_link_or_login_required,
+)
 from .filters import get_scheme_label
 
 
@@ -409,7 +414,7 @@ def deposit_create(community=None):
     )
 
 
-@login_required
+@secret_link_or_login_required()
 @pass_draft(expand=True)
 @pass_draft_files
 def deposit_edit(pid_value, draft=None, draft_files=None, files_locked=True):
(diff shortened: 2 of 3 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/ShareOptions/ShareModal.js
```diff
@@ -31,7 +31,7 @@ export class ShareModal extends Component {
       usersResults: undefined,
       groupsResults: undefined,
       linksResults: undefined,
-      activeTab: 0,
+      activeTabKey: "",
```
zzacharo, line 34: are these relevant changes regarding login required?

anikachurilova: No, changes in this file are the bug fix of "Settings" tab reported by Lars

zzacharo: Can you please split them in a separate commit just for clarity please? Basically, keep in one commit only the removal of the `login_required`...

anikachurilova: Done!

invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -409,7 +409,6 @@ def deposit_create(community=None):
     )
 
 
-@login_required
```
kpsherva, line 412: since we remove this, I think we should consider also this scenario:
1. User was logged in, editing the draft
2. User's session expired
3. User refreshed the page
4. Got permission denied, and the fact of expired session is not very visible in the interface, so it can be interpreted as "I had access before and now I don't, what is going on?"

shouldn't we instead check the token in the link if it allows for edits? And then:
if no token-> redirect to login 
token with insufficient permissions -> raise PermissionDenied()
login with insufficient permissions -> raise PermissionDenied()
login with OK permissions or token with OK permissions -> show the view

ntarocco: We briefly discussed, we should indeed have an enhanced decorator as suggested by @kpsherva that will simply do:
```
def token_or_login_required():
  if not token:
    login_required
```

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -381,3 +382,20 @@ def view(*args, **kwargs):
         return response
 
     return view
+
+
+def secret_link_or_login_required():
+    """Skip login check for requests with secret links."""
```
zzacharo, line 388:
```suggestion
    """Skip login redirection check for requests with secret links.
    
    If access has been granted via a secret link, then permission checks are tested 
    in the decorated view.
    """
```
A bit more explanatory description?

zzacharo (approved):

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -381,3 +382,24 @@ def view(*args, **kwargs):
...
+
+    If access has been granted via a secret link, then permissions are checked
+    in the dedicated view.
+    """
+
+    def decorator(f):
+        @wraps(f)
+        def view(**kwargs):
+            secret_link_token_arg = "token"
+            session_token = session.get(secret_link_token_arg, None)
```
kpsherva, line 398: one clarification, I think we should leave a comment on this in the code on how the token is already in the session, since more "obvious" to understand how this works is to check if token is in the request args

general thoughts: it was not clear to me how at this stage the token is already in the session, so I started digging and maybe it makes sense that it was already loaded there by identity change signal (are we sure the signal will be processed before we enter this code block? I guess we can assume the identity loader always goes first, otherwise sometimes shared links would fail - not sure if this is possible)

an alternative would be checking if token is there as a request argument, like it is done in verify_token function, but it might be redundant

anikachurilova: Token is not present here as a request argument, as the request at this point is created by clicking on "Edit" or "New version" buttons, which is a simple create draft request (not the secret link that was copied). We could retrieve token from the request's `Referer` header, but it is not a reliable source of information at this point IMHO.
And yes, you are right, we can assume that the `on_identity_loaded` will be called before the deposit edit view

PR #2696, opened 2024-06-10
Backport fix unsafe HTML

alejandromumo:


alejandromumo:
The fix was backported to v9.1, this PR is not needed.

PR #2724, opened 2024-07-01
ui: CSS fixes

anikachurilova:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/80

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/label.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/label.overrides
@@ -1,6 +1,7 @@
 .ui.label.access-status {
   color: @white;
   width: max-content;
+  z-index: 100;
 
   &.open {
     background-color: @positiveColor;


ntarocco (approved):

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/label.overrides
```diff
@@ -1,6 +1,7 @@
 .ui.label.access-status {
   color: @white;
   width: max-content;
+  z-index: 100;
```
kpsherva, line 4: I have checked the issue and the underlying problem is not the ui label elemnt but data-tooltip.
an example of why this is not optimal solution is: If you set the z-index of the label and you have a sticky header on top of the page, you will end up with labels "hovering" over the header (or other sticky elements)
I would suggest to go with

```
[data-tooltip]:after {
   z-index:3
}
```

kpsherva:
closing in favour of https://github.com/inveniosoftware/invenio-app-rdm/pull/2728/files#diff-c4c330faab677d7c41b090ce5d9526ed283ca48128cfc1dc66c2d2b3f9522ab0R533

PR #2733, opened 2024-07-03
deposit: provide and check publish permission for visibility

### LINKED ISSUE #2732: Publish button visible without permission to publish
Package version (if known): 11.0.0 - 13.0.0b0.dev3

## Describe the bug
The publish button on the deposit page is always visible - even if a user does not fulfill the requirements for the `can_publish` permission.
Pressing the publish button will then result in a `Permission denied` error.

## Steps to Reproduce
1. Setup a permission policy which only allows a user with administrator role to publish 
2. As a user without this role, create a new deposit
3. Fill out all required fields
4. Push publish button
5. See `Permission denied` error

## Expected behavior
Publish button is only visible if a user fulfills the requirements for the `can_publish` permission.

rekt-hard:

### Description

depends on https://github.com/inveniosoftware/invenio-rdm-records/pull/1768
closes #2732 
Provide `can_publish` permission when rendering the deposit page. Also reordered the permission alphabetically.

Front end check is performed in: https://github.com/inveniosoftware/invenio-rdm-records/pull/1768

With permission:
Screenshot from 2024-07-03 15-36-25

Without permission:
Screenshot from 2024-07-03 15-35-43

### Checklist

**Frontend**

**Reminder**

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -403,10 +403,11 @@ def deposit_create(community=None):
         files_locked=False,
         permissions=get_record_permissions(
             [
+                "delete_draft",
                 "manage",
                 "manage_files",
-                "delete_draft",
                 "manage_record_access",
+                "publish",
             ]
         ),
     )
@@ -469,11 +470,12 @@ def deposit_edit(pid_value, draft=None, draft_files=None, files_locked=True):
         files_locked=files_locked,
         permissions=draft.has_permissions_to(
             [
-                "manage",
-                "new_version",
                 "delete_draft",
+                "manage",
                 "manage_files",
                 "manage_record_access",
+                "new_version",
+                "publish",
             ]
         ),
     )
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
@@ -637,7 +637,11 @@ export class RDMDepositForm extends Component {
                           </Grid.Column>
 
                           <Grid.Column width={16} className="pt-10">
-                            <PublishButton fluid record={record} />
+                            <PublishButton
+                              fluid
+                              permissions={permissions}
+                              record={record}
+                            />
                           </Grid.Column>
 
                           <Grid.Column width={16} className="pt-0">


invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -471,11 +472,12 @@ def deposit_edit(pid_value, draft=None, draft_files=None, files_locked=True):
...
         permissions=draft.has_permissions_to(
             [
-                "manage",
-                "new_version",
                 "delete_draft",
+                "manage",
                 "manage_files",
                 "manage_record_access",
+                "new_version",
+                "publish",
```
max-moser, line 480: this seems to be the only new value, the rest is just reordered?

max-moser:
Since this is a quite minimal and obvious code change (but with large positive impact), I dare say LGTM

max-moser (approved):

invenio_app_rdm/records_ui/views/deposits.py
rekt-hard: Exactly. Reordered them alphabetically and added `publish`

fenekku (commented):
I am perhaps iconoclastic here: we found that our users can be confused by disappearing buttons like these. What about disabling the button and having a hover-over that explains why instead of removing the button completely?

I'll defer to any 3rd-party that chimes in though :)

max-moser:
good point, perhaps the button should simply be "greyed out" with an on-hover tooltip.
WDYT @rekt-hard ?

Samk13:
> good point, perhaps the button should simply be "greyed out" with an on-hover tooltip. WDYT @rekt-hard ?

I agree with @max-moser, I did something similar to this [PR](invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/controls/PublishButton/PublishButton.js) which hasn't been merged yet. Here's a screen record: 
https://github.com/inveniosoftware/invenio-rdm-records/pull/1283#issuecomment-1546049040

rekt-hard:
@fenekku @max-moser @Samk13 thank you for the input - very valid reasons. Just trying to depict the current state in InvenioRDM.

tldr; Setting the button active/inactive makes sense. Help on choosing the correct help text is welcome.

For requests, only the actions a user can perform are shown. For a user, a permission to perform an action usually also does not change by updating the request in some way. So it does not depend on some user action (like missing a required field) whether or not the action can be performed.

For rdm records on the deposit page, there is the share and publish button. These are always shown since from the back end, a user does fulfill the permissions to perform these actions. However, there is no sense in clicking on these buttons if the record is in an invalid state since this would lead to an error any way. So the front end is performing some extra checks for usability reasons. As soon as the record is in a certain state, the buttons becomes active (through user actions).

It looks like there are two possibilities:
- A user **_is not_** allowed to perform an action from the back end perspective -> **_do not_** display the button
- A user **_is_** allowed to perform an action from the back end perspective but the front end has additional checks for usability -> display the button but set it active depending on some local state

In my opinion, this issue and PR are more related to the second situation. The publish button will not become active just by filling out some fields. Another other action, triggered by some other user or the system will have to happen before the user will even be allowed to perform the publish action.

### HOWEVER
In the case of a request, a user usually has enough context to understand that only certain actions are available anyway for them.
In the case of a new record, there might be several reasons why a user is not allowed to publish which might be unknown to the user.

I guess we can all envision the frustration of people wondering where the publish button went off to. Especially if it is present for record A but not for record B. In order to prevent this, simply letting the UI decide the state of the button seems to be more appropriate here - given that there will be meaningful hints/help texts explaining why a button is not active (which is not the case now by the way).

Since the possibility to click the button might be front end or back end related, there has to be some logic involved and probably also some additional payload from the back end. Ideas on how to conditionally set the correct help texts are very welcome :)

kpsherva:
I hope you don't mind if I jump in with some UX suggestions, even if it is late.
From our experience, hiding elements of the interface is confusing for the end user - we receive support tickets stating "My colleague sees the button and I am not, there is a bug"
In order to avoid the confusion, in all of the cases we disable the button instead of hiding it, and we are adding the tooltip explaining why the button is disabled.
Could you align the changes to follow this approach to make it consistent?

rekt-hard:
@fenekku @kpsherva Thank you for the input - we are on the same page now and reached the same conclusion :smile: 
I will update the implementation accordingly

rekt-hard:
@fenekku @kpsherva @max-moser @Samk13 
Updated this PR to pass the permissions in the front end.
Also created this PR for the publish button component: https://github.com/inveniosoftware/invenio-rdm-records/pull/1768

fenekku (approved):
@rekt-hard thanks for the work on this! If you need/want/can version / merge / release all the better or we can piggyback on other releases of these modules.

rekt-hard:
@fenekku I do not have the permissions to do that. Would be great to get this into v12 though if possible :)

max-moser:
i'd like to see this in v12 too, but i think in yesterday's telecon it was agreed to freeze non-critical bug fixes etc. for v12 for now – so maybe v12.1?
ping @utnapischtim

fenekku:
What @max-moser said - defer to @utnapischtim + @kpsherva for this release's management. However I really think we should aim to have a 12.1, .2 ... series for v12 and have so with a faster turnaround. I'd put that there definitely. v11 was released in January of *2023*. A discussion about pace/content/approach of releases would be warranted.

kpsherva (approved):

max-moser:
like its sibling PR, could you please rebase it? @rekt-hard 
then i'll merge it

Issue #2746, opened 2024-07-15
Improve diacritic search


carlinmack:
## Example

can't find this record by searching for fruhwirth

https://sandbox.zenodo.org/records/87676

PR #2763, opened 2024-07-26
config: Bump records index version to v7.0.0

sakshamarora1:
Closes: https://github.com/inveniosoftware/invenio-app-rdm/issues/2746

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -665,7 +665,7 @@ SEARCH_HOSTS = [{"host": "localhost", "port": 9200}]
 # ===============
 # See https://invenio-indexer.readthedocs.io/en/latest/configuration.html
 
-INDEXER_DEFAULT_INDEX = "rdmrecords-records-record-v6.0.0"
+INDEXER_DEFAULT_INDEX = "rdmrecords-records-record-v7.0.0"
 """Default index to use if no schema is defined."""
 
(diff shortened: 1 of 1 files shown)


utnapischtim:
does this mean an instance has to upgrade there records to use the new schema? so an upgrade script is necessary?

zzacharo:
> does this mean an instance has to upgrade there records to use the new schema? so an upgrade script is necessary?

If I am not mistaken, that means that one must reindex the records so they are searchable again.

invenio_app_rdm/config.py
```diff
@@ -665,7 +665,7 @@ def files_rest_permission_factory(obj, action):
 # ===============
 # See https://invenio-indexer.readthedocs.io/en/latest/configuration.html
 
-INDEXER_DEFAULT_INDEX = "rdmrecords-records-record-v6.0.0"
+INDEXER_DEFAULT_INDEX = "rdmrecords-records-record-v7.0.0"
```
kpsherva, line 668: will the index be created automatically after deployment or we need to add this to an upgrade recipe>

sakshamarora1: We will create the index and do reindexing before but Alex and I discussed that records and communities indices bump up need to be strategised to prevent downtime if possible.

zzacharo (approved):

PR #2769, opened 2024-08-01
custom_fields: allow to search vocab

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/180
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/143

image

### DIFF
--- a/invenio_app_rdm/records_ui/views/filters.py
+++ b/invenio_app_rdm/records_ui/views/filters.py
@@ -15,6 +15,7 @@ import idutils
 from babel.numbers import format_compact_decimal, format_decimal
 from flask import current_app, url_for
 from invenio_base.utils import obj_or_import_string
+from invenio_i18n import get_locale
 from invenio_previewer.views import is_previewable
 from invenio_records_files.api import FileObject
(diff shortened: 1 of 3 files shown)


kpsherva:
the dropdown icon looks a little bit strange to me, what is it referring to?

jrcastro2:
> the dropdown icon looks a little bit strange to me, what is it referring to?

These are custom fields, I added them without section name for now. We will need to find a proper name for this.

kpsherva:
could we add a name? it looks broken otherwise

invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -182,6 +183,25 @@ def custom_fields_search(field, field_value):
     )
 
 
+def custom_vocab_fields_search(field, field_value, sub_field):
```
kpsherva, line 186: do we need a split from custom_fields_search? could we instead add a flag if is it a vocab and use a function above?

jrcastro2:
> could we add a name? it looks broken otherwise, also the changes in react-invenio-forms are only to handle the fact we do not have a name?

Sure, name is temporal though
image

PR #2773, opened 2024-08-05
templates: landing page custom field macro refactor

kpsherva:


invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -177,26 +177,14 @@ def custom_fields_search(field, field_value):
...
-    locale = get_locale()
-    if not locale:
-        locale = current_app.config.get("BABEL_DEFAULT_LOCALE", "en")
-
-    namespace_string = "\:".join(namespace_array) + f".{sub_field}.{locale}"
+    localised_title = field_cfg.get("locale")
+    if localised_title:
+        locale = get_locale()
+        if not locale:
+            locale = current_app.config.get("BABEL_DEFAULT_LOCALE", "en")
```
slint, line 184: minor: isn't that always returned at the end [from `get_locale()`](https://github.com/inveniosoftware/invenio-i18n/blob/721fc1d50e7f53184955501f110ffaae70e822e3/invenio_i18n/selectors.py#L74)?

invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -177,26 +177,14 @@ def custom_fields_search(field, field_value):
...
+        namespace_string = "\:".join(namespace_array) + f".{localised_title}.{locale}"
```
slint, line 185: nit: I have some trouble imagining what the final value will be here, maybe add a comment above with an example? Or if it's too much it can also go in the function docstring.

slint (approved):
LGTM, two minor comments

.github/workflows/tests.yml
```diff
@@ -31,52 +31,12 @@ on:
...
-      - name: Run tests
-        run: ./run-tests.sh
+  Python:
+      uses: inveniosoftware/workflows/.github/workflows/tests-python.yml@master
+      with:
+        extras: "tests"
+  JS:
+    uses: inveniosoftware/workflows/.github/workflows/tests-js.yml@master
+    with:
+      js-working-directory: ./invenio_app_rdm/assets/semantic-ui/js/invenio_app_rdm
```
slint, line 41:
```suggestion
      js-working-directory: invenio_app_rdm/assets/semantic-ui/js/invenio_app_rdm
```

Might fix the issue in the tests

invenio_app_rdm/records_ui/views/filters.py
kpsherva: I assume @jrcastro2 checked this and added for a reason, I didn't verify myself

slint: Ok, I guess better safe than sorry then 😅

.github/workflows/tests.yml
```diff
@@ -31,52 +31,11 @@ on:
...
-      - name: Run tests
-        run: ./run-tests.sh
+  Python:
+      uses: inveniosoftware/workflows/.github/workflows/tests-python.yml@master
+      with:
+        extras: "tests"
+  JS:
+    uses: inveniosoftware/workflows/.github/workflows/tests-js.yml@master
+    with:
+      translations-working-directory: ./invenio_app_rdm/assets/semantic-ui/translations/invenio_app_rdm
```
slint, line 41:
```suggestion
      translations-working-directory: ./invenio_app_rdm/theme/assets/semantic-ui/translations/invenio_app_rdm
```

There's are only `theme` assets apparently.

zzacharo (approved):

Issue #2767, opened 2024-07-30
Vocabularies: Improve subjects field


jrcastro2:
Subjects field should display all values on the list and allow to serach as well, not displaying all the values is confusing.
If only one term we should omit the first dropdown

PR #2776, opened 2024-08-06
DepositForm: Add search_on_focus attribute to subjects RemoteSelectField

sakshamarora1:
closes: https://github.com/inveniosoftware/invenio-app-rdm/issues/2767

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
@@ -366,6 +366,7 @@ export class RDMDepositForm extends Component {
                       fieldPath="metadata.subjects"
                       initialOptions={_get(record, "ui.subjects", null)}
                       limitToOptions={this.vocabularies.metadata.subjects.limit_to}
+                      searchOnFocus
                     />
                   </Overridable>
 


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -366,6 +366,7 @@ export class RDMDepositForm extends Component {
                       fieldPath="metadata.subjects"
                       initialOptions={_get(record, "ui.subjects", null)}
                       limitToOptions={this.vocabularies.metadata.subjects.limit_to}
+                      searchOnFocus={this.vocabularies.metadata.subjects.search_on_focus}
```
kpsherva, line 369:
```suggestion
                      searchOnFocus={true}
```
I think adding a backend config variable is a bit an overkill, we can control this entirely on the frontend

sakshamarora1: This switch won't be as accessible in that case. Also if we want to add a new functionality similar to this, it would be easy with the config. Especially for the inbuilt props which could be passed to the config without updating the component. WDYT?

kpsherva: lets dicuscuss IRL, I am not sure I understand what you mean so let's clarify

kpsherva (approved):

PR #2788, opened 2024-08-12
dependencies: bump invenio-cache minimum for v12+

fenekku:
invenio-cache >= 1.3.0 is needed in InvenioRDM v12+ for invenio-vocabularies for example (see https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/services/facets.py#L12 ) .

This should be backported to maint-v12 when accepted. I can do that then or wait for more patches to come in since the workaround is to specify invenio-cache >= 1.3.0 in instance's Pipfile.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ install_requires =
     # Invenio core modules
     invenio-app>=1.4.0,<2.0.0
     invenio-base>=1.3.0,<2.0.0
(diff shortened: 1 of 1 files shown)


kpsherva (approved):

fenekku:
Merging it in since low impact. I will make a backport tomorrow and mention it here when it's done.

fenekku:
Backported to v12.0.2 .

PR #2793, opened 2024-08-19
migrate to v12: emit non-zero exit code on error

fenekku:
This is a convenience for automated upgrades to be able to detect
errored command from the outside.


max-moser (approved):
that should help with automation via scripts!

fenekku:
backported to maint-v12.x

PR #2794, opened 2024-08-20
landing page: support different MathJax delimeters

anikachurilova:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/133

Add support for `$` and `\` MathJax delimeters. 

Examples of records that are using different types:
https://repository.cern/records/fp18d-jc149
https://zenodo.org/records/8292839

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -407,7 +407,7 @@
 
 {%- block javascript %}
   {% if use_theme_basic_template|default(true) %}
-    {{ super.super() }}
+    {{ super() }}
   {% else %}
     {% include config.THEME_JAVASCRIPT_TEMPLATE %}
   {% endif %}
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -406,11 +406,8 @@
 {%- endblock page_body %}
 
 {%- block javascript %}
-  {% if config.THEME_MATHJAX_CDN %}
-    <script type="text/javascript" src="{{ config.THEME_MATHJAX_CDN }}"></script>
-  {% endif %}
   {% if use_theme_basic_template|default(true) %}
-    {{ super() }}
+    {{ super.super() }}
   {% else %}
     {% include config.THEME_JAVASCRIPT_TEMPLATE %}
   {% endif %}
--- a/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/javascript.html
+++ b/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/javascript.html
@@ -6,6 +6,18 @@
   Invenio is free software; you can redistribute it and/or modify it
   under the terms of the MIT License; see LICENSE file for more details.
 #}
+{% if config.THEME_MATHJAX_CDN %}
+  <script type="text/javascript">
+    window.MathJax = {
+      tex: {
+        inlineMath: [['$', '$'], ['\\(', '\\)']],
+        processEscapes: true // Allows escaping $ signs if needed
+      }
+    };
+  </script>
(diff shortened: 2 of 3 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -407,6 +407,14 @@ <h4>{{ _("Request access") }}</h4>
 
 {%- block javascript %}
   {% if config.THEME_MATHJAX_CDN %}
+    <script type="text/javascript">
```
ntarocco, line 410: move the entire block to `invenio-theme`, but pay attention to the `else` below `include config.THEME_JAVASCRIPT_TEMPLATE`

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/page.html
```diff
@@ -34,3 +34,6 @@
   {%- endblock page_footer %}
 
 {%- endblock body_inner %}
+{%- block javascript %}
+  {{ super() }}
+{%- endblock javascript %}
```
jrcastro2, line 39: Minor: If we simply call super we can simply remove the block and it would be the same, no?
```suggestion
```

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/javascript.html
```diff
@@ -6,6 +6,18 @@
...
+    window.MathJax = {
+      tex: {
+        inlineMath: [['$', '$'], ['\\(', '\\)']],
+        processEscapes: true // Allows escaping $ signs if needed
+      }
+    };
+  </script>
+  <script type="text/javascript" src="{{ config.THEME_MATHJAX_CDN }}"></script>
+{% endif %}
+
```
jrcastro2, line 20: Minor: I think this is duplicated, if we are adding it to the base page.html (as done in [this PR](https://github.com/inveniosoftware/invenio-theme/pull/394/files#diff-6b3948d6e559b49f40b74e035d308f26ee3f8d78e7f83c709660098ac90edcd6R118-R130)) it shouldn't be needed here? unless I am missing something

anikachurilova: This piece of code will be called in the `else` of this condition (in reality this is record detail view in the requests): https://github.com/inveniosoftware/invenio-app-rdm/pull/2794/files#diff-d74bd02e8b7c3bedcec8f02daae657ffd97db51aea164a8a80ca5dfc88b39658R412

And the code from base page.html is called from here: https://github.com/inveniosoftware/invenio-app-rdm/pull/2794/files#diff-d74bd02e8b7c3bedcec8f02daae657ffd97db51aea164a8a80ca5dfc88b39658R410

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/page.html
anikachurilova: Now that in the child component I'm calling `super.super()`, some tests were failing because it couldn't find javascript block in here.

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/javascript.html
jrcastro2: Hmmm I would expect that if it's already added in `invenio_theme/templates/semantic-ui/invenio_theme/page.html` it's present in all the views as it's the global template we use, unless I am wrong.

I am also suprised about the `super.super()`, I think that this is skipping the parent super and going to the grandfather, see this for reference: https://jinja.palletsprojects.com/en/3.1.x/templates/

I find it a bit confusing to follow this and would expect to have a simpler way to enable it.

I think that adding it to the base global page is correct and from there I would simply call super() if we ever need to override that block. Maybe I am missing something.

ntarocco: We need to review all this IRL, because it is not clear to me why we need it. From a quick look, we always use `THEME_JAVASCRIPT_TEMPLATE`.

jrcastro2 (approved):
Reviewed with @sakshamarora1 LGTU! 🚀

PR #2799, opened 2024-08-21
sidebar: display 'Sizes' properly [+]

fenekku:
- fix to display Sizes (wasn't displaying in v12)
- simplify formats as well

I will backport this to maint-v12.x when merged.

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/side_bar/details.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/side_bar/details.html
@@ -8,8 +8,8 @@
   it under the terms of the MIT License; see LICENSE file for more details.
 #}
 
-{%- from "invenio_app_rdm/records/macros/detail.html" import list_formats, list_languages,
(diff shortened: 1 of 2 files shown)


max-moser (commented):
the `list_formats` seems to have been a basically equivalent re-implementation of `", ".join()`, that looks fine
the `list_string_values` macro is [this one here?](<https://github.com/inveniosoftware/invenio-app-rdm/blob/maint-v12.0.x/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/detail.html#L34>) – how did it work before?
it looks like the macro wants two arguments, but was only supplied one?

max-moser:
the logic changes look simple enough, but i have no idea about what it looked like before – could i ask you for before/after screenshots?

fenekku:
Yes no problem!

**before**
before_sizes

**after**
after_sizes

It's indeed because the list_string_values wasn't given the additional parameter, but in this scenario that whole logic was unnecessary as far as I can tell.

max-moser:
thanks!
looks like it previously was just plain broken :skull: 
much better now!

max-moser (approved):

PR #2806, opened 2024-08-23
tests: use opensearch2

fenekku:
Use opensearch2


fenekku:
just experimental so closing

PR #2807, opened 2024-08-26
Deposit form: Provide props to overridable components

max-moser:
Some of the `<Overridable>` components have been using values that weren't passed as props.
This makes it hard to provide overrides for those components that are functionally the same.

For example, I am writing an override for the `InvenioAppRdm.Deposit.CardDepositStatusBox.container` for [`Invenio-Curations`](https://github.com/tu-graz-library/invenio-curations/) which adds a section for the curation review request.
Without passing the props, this override couldn't access the `props.permissions` which is used by the default component to enable/disable some buttons.

With the props being passed, I can use it as expected:
image

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
@@ -128,7 +128,10 @@ export class RDMDepositForm extends Component {
           />
         </Overridable>
 
-        <Overridable id="InvenioAppRdm.Deposit.CommunityHeader.container">
+        <Overridable
+          id="InvenioAppRdm.Deposit.CommunityHeader.container"
+          record={record}
+        >
           {!this.hide_community_selection && (
             <CommunityHeader
               imagePlaceholderLink="/static/images/square-placeholder.png"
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -641,6 +652,9 @@ export class RDMDepositForm extends Component {
                   <Overridable
                     id="InvenioAppRdm.Deposit.AccessRightField.container"
                     fieldPath="access"
+                    permissions={permissions}
+                    recordRestrictionGracePeriod={recordRestrictionGracePeriod}
+                    allowRecordRestriction={allowRecordRestriction}
```
fenekku, line 657: We'd need `record` here too if one were to override this one.

fenekku (approved):
Just one more props to pass. :+1:

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
max-moser: thanks, i missed that!

max-moser: props to you! :wink:

PR #2809, opened 2024-08-26
config: filter out robots and flag machines

slint:
- Depends on https://github.com/inveniosoftware/invenio-stats/pull/149 and https://github.com/inveniosoftware/invenio-rdm-records/pull/1803


alejandromumo (approved):
CI is failing, will be sorted when `invenio_stats` is released I believe

slint:
Apparently, the `is_machine` field was missing from the mapping for the record view events... I've added it in https://github.com/inveniosoftware/invenio-rdm-records/pull/1803, so when released this will start passing

anikachurilova (approved):

ptamarit (approved):

PR #2814, opened 2024-08-27
Upgrade script: Mint the new concept DOI for each upgraded record

max-moser:
Previously, the script would create a new concept DOI for each record but never actually mint them on DataCite.
The [`TaskOp` part from the PIDs service component](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/components/pids.py#L204) was missing, which called the [`register_or_update_pid` task](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/pids/tasks.py#L17).

### DIFF
--- a/invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
@@ -94,6 +94,12 @@ def execute_upgrade():
                     record.parent, pids
                 )
                 record.parent["pids"] = pids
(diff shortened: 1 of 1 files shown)


fenekku (approved):
That works. These network calls are going to be super slow for an instance like ours with our thousands of records, but I am moving to a tasked out upgrade anyway. At least this will do the right thing.

max-moser:
yeah, this is definitely going the "make it work first, then make it fast" route.

PR #2820, opened 2024-08-30
migration: account for deleted communities and draft concept DOI

fenekku:
- account for deleted Communities . Those don't have `"access"` raising an error.
- **[edit 2024-09-04]** second change : concept DOIs should only be minted for published records (`register_or_update` is called on publishing typically) and in the migration script it would have been called on drafts too. `register_or_update` already covers private/public record though which is great.

### DIFF
--- a/invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
@@ -59,10 +59,12 @@ def execute_upgrade():
     """
 
     def migrate_review_policy(community_record):
-        review_policy = community_record["access"].get(
+        if community_record.is_deleted:
+            return
+
+        community_record["access"].setdefault(
             "review_policy", ReviewPolicyEnum.CLOSED.value
         )
-        community_record["access"]["review_policy"] = review_policy
 
     def update_parent(record):
         """Update parent schema and parent communities for older records."""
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
```diff
@@ -59,10 +59,12 @@ def execute_upgrade():
     """
 
     def migrate_review_policy(community_record):
-        review_policy = community_record["access"].get(
+        if community_record.is_deleted:
+            return
+
+        community_record["access"].setdefault(
```
fenekku, line 65: just some prettification

invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
```diff
@@ -94,12 +96,17 @@ def update_parent(record):
...
                 record.parent["pids"] = pids
-                current_rdm_records.records_service.pids.register_or_update(
-                    id_=record["id"],
-                    identity=system_identity,
-                    scheme="doi",
-                    parent=True,
-                )
+                # Have to commit here otherwise register_or_update won't get
+                # the above data
+                record.parent.commit()
```
fenekku, line 101: In particular without that, records with external DOI will cause register_or_update below to fail since RDMRecord is fetched independently in that method and therefore doesn' provide the right provider(!).

max-moser (approved):
LGTM

Issue #2821, opened 2024-09-02
Bug on adding a record to a community


jrcastro2:
On adding a record to a community, we get an internal server error

PR #2822, opened 2024-09-02
modal: fix record adding to a community

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2821


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCommunitySubmission/RecordCommunitySubmissionModal.js
```diff
@@ -79,7 +79,14 @@ export class RecordCommunitySubmissionModal extends Component {
           ],
         };
       }
-      const response = await http.post(recordCommunityEndpoint, data);
+      // When adding communities to a record, it should simple return the json representation of the response
+      // no nee to process the response with the schema serializer
```
alejandromumo, line 83:
```suggestion
      // no need to process the response with the schema serializer
```

alejandromumo (approved):
Minor comment. 

I checked that the only place where `data` is used is like `data.processed[0].request`. So this change should be OK (no need to serialize the response to anything more complex)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCommunitySubmission/RecordCommunitySubmissionModal.js
```diff
@@ -79,7 +79,14 @@ export class RecordCommunitySubmissionModal extends Component {
...
+      const response = await http.post(recordCommunityEndpoint, data, {
+        headers: {
+          "Accept": "application/json",
```
zzacharo, line 86: Shouldn't this be the other way around and fix the endpoint to accept the `inveniordm.json+vnd`? @slint

jrcastro2: I am jumping in since I might have a bit more context. (However anyone is welcomed to comment!)

The endpoint currently accepts the `inveniordm.json+vnd` content type. However, it is part of a resource that utilizes a serializer with a schema expecting a standard community dictionary object. Since we are passing a custom dictionary structure instead, the schema validation fails when using our custom `Accept` header. As a result, we have decided to disable this header in both the backend and frontend for this endpoint.

PR #2831, opened 2024-09-05
replace view_func with default_view_func to make APP_RDM_USER_DASHBOARD_ROUTES overwriteable

karkraeg:

### Description

This change makes[ `APP_RDM_USER_DASHBOARD_ROUTES`](https://github.com/inveniosoftware/invenio-app-rdm/blob/56893c6a21e6ff58dd6732fa77ae34554c263a7b/invenio_app_rdm/config.py#L766) configurable by using an approach like this in invenio.cfg:

```
from invenio_app_rdm.config import APP_RDM_ROUTES
from invenio_app_rdm.config import APP_RDM_USER_DASHBOARD_ROUTES
from mysite.views import custom_render
APP_RDM_ROUTES["index"] = ("/", custom_render)
APP_RDM_USER_DASHBOARD_ROUTES["uploads"] = ("/me/uploads", custom_render)
```

### DIFF
--- a/invenio_app_rdm/users_ui/views/ui.py
+++ b/invenio_app_rdm/users_ui/views/ui.py
@@ -14,6 +14,7 @@ from flask_login import current_user
 
 from ..searchapp import search_app_context
 from .dashboard import communities, requests, uploads
+from invenio_app_rdm.theme.views import create_url_rule
 
 
 #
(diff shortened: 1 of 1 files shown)


karkraeg:
Hi @alejandromumo do you think that's a good solution, or do you see anything speaking against it from your perspective?

karkraeg:
Hi @alejandromumo do you think that's a good solution, or do you see anything speaking against it from your perspective?

karkraeg:
I'll close this in favor of [https://github.com/inveniosoftware/invenio-app-rdm/pull/2831 b](https://github.com/inveniosoftware/invenio-rdm-records/issues/1846) because there is no need for now to change this. It's better to fix the actual bug in the JS Code than to change how this configuration variable behaves.

PR #2832, opened 2024-09-05
config: make OAI-PMH record index dynamic

slint:


invenio_app_rdm/config.py
```diff
@@ -694,7 +699,9 @@ def files_rest_permission_factory(obj, action):
 # ===============
 # See https://invenio-indexer.readthedocs.io/en/latest/configuration.html
 
-INDEXER_DEFAULT_INDEX = "rdmrecords-records-record-v7.0.0"
+# TODO: Set this to `None` in the default config of Invenio-Indexer
+# We want that indexers are always explicit about the index they are indexing to
+INDEXER_DEFAULT_INDEX = None
```
kpsherva, line 704: Would it make sense to replace the TODO with a pr removing the default in invenio-indexer?

slint: Opened https://github.com/inveniosoftware/invenio-indexer/pull/158 and will replace in the comment.

That means making a major release on `invenio-indexer`, so I pushed it back a bit since we've already gone through a major bump in InvenioRDM recently. In theory, this shouldn't break anyone's workflow I think we've moved quite some time ago away from having `records-record-v1.0.0` being some sort of magic schema/mapping that just exists by default in `invenio-records`.

PR #2835, opened 2024-09-10
fix: the following / prevents rspack from building

utnapischtim:


slint (approved):
I think these are ok to merge regardless of the end purpose :)

zzacharo (approved):

PR #2848, opened 2024-09-17
admin-records: account for system owned records

fenekku:
account_for_system_owner


fenekku:
Check backend for entity resolver in long-term. for now, this is good enough.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
```diff
@@ -73,8 +78,7 @@ class SearchResultItemComponent extends Component {
           collapsing
           className="word-break-all"
         >
-          {/* TODO needs to be expanded */}
-          {result.parent.access.owned_by.user}
+          {(recordOwner && recordOwner.user) || "System"}
```
kpsherva, line 81: please leave a comment with a link to backend fix issue so that we replace it with the long term solution :)

kpsherva (approved):

kpsherva:
agreed to this temporary fix and later on providing the long term backend solution

fenekku:
Added reference. Merging.

PR #2850, opened 2024-09-18
release v12.0.6

fenekku:
- **search-result: provide key to part of community array element**
- **search-result: namespace overridable id for community search results**
- **theme: fix accordion rotation**
- **css: added class for padding removal**
- **migration: account for deleted communities and draft concept DOI**
- **admin-records: account for system owned records**
- **admin-records: add reference to gh issue**
- **release: v12.0.6**


fenekku:
Will merge as soon as CI passes and then release.

PR #2856, opened 2024-09-24
remove flask upper pin

utnapischtim:
context: migrate to flask >= 3.0

- fix: remove DeprecationWarning for sqlalchemy
- setup: remove flask upper pin

* the sqlalchemy commit is here because without upgrading sqlalchemy it is not possible to go to flask >= 2.3

* the remove of invenio-admin is not part of the migration to flask >= 3.0 because the problem with the app.before_first_request has been fixed in invenio-admin. nevertheless the PR for it should be merged!

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -4,7 +4,7 @@
 # Copyright (C) 2019-2021 Northwestern University.
 # Copyright (C)      2021 TU Wien.
 # Copyright (C) 2022 KTH Royal Institute of Technology
-# Copyright (C) 2023 Graz University of Technology.
+# Copyright (C) 2023-2024 Graz University of Technology.
 #
 # Invenio App RDM is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -183,9 +184,10 @@ def custom_fields_search(field, field_value, field_cfg=None):
         if not locale:
             locale = current_app.config.get("BABEL_DEFAULT_LOCALE", "en")
         # example: cern:experiments.title.en
-        namespace_string = "\:".join(namespace_array) + f".{localised_title}.{locale}"
+        namespace_string = ":".join(namespace_array) + f".{localised_title}.{locale}"
```
kpsherva, line 187: this change will break the searches in custom fields
basically if you search for a specific field
`q=custom_fields.cern\:experiments:ATLAS`
the slash prevents the custom field namespace separator `:` to be interpreted as a part of lucene syntax

kpsherva: could you add a comment above, discouraging people from removing this slash ?:)

utnapischtim: this has been moved to a separate PR

PR #2860, opened 2024-09-25
collections: added collection page with minimal UI

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1820

--- 

**Screenshots**

I have the following collections:

- EU (Community)
    - Funding programmes (Tree)
        - Horizon 2020 collection (2 records)
            - Open access collection (1 record)
            - Restricted access  collection (1 record)

Screenshot at Sep 26 11-04-01

Screenshot at Sep 26 11-04-13

Screenshot at Sep 26 11-04-21
e654f0be">

### DIFF
--- a/invenio_app_rdm/communities_ui/templates/semantic-ui/invenio_communities/collections/collection.html
+++ b/invenio_app_rdm/communities_ui/templates/semantic-ui/invenio_communities/collections/collection.html
@@ -1,0 +1,28 @@
+{# -*- coding: utf-8 -*-
+
+This file is part of Invenio.
+Copyright (C) 2024 CERN.
+
+Invenio is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
+#}
+
+{% extends "invenio_communities/details/base.html" %}
+
+{%- block javascript %}
+{{ super() }}
+{{ webpack['invenio-app-rdm-collections-search.js'] }}
+{%- endblock %}
+
+{%- block page_body %}
+{{ super() }}
+<div class="ui container fluid theme-tertiary page-subheader rel p-10 mr-0-mobile ml-0-mobile">
+  <h1 class="ui container">
+    {{ collection.title }}
+  </h1>
+</div>
+<div id="collection-search"
+  data-invenio-search-config='{{search_app_collection_config(app_id="InvenioAppRDM.CollectionsSearch", endpoint=collection.links.search, hidden_params=[["q", collection.query]]) | tojson }}'
+  data-permissions='{{ permissions | tojson }}'>
+</div>
+{%- endblock page_body %}
--- a/invenio_app_rdm/communities_ui/views/communities.py
+++ b/invenio_app_rdm/communities_ui/views/communities.py
@@ -19,7 +19,10 @@ from invenio_communities.views.communities import (
 from invenio_communities.views.decorators import pass_community
 from invenio_pages.proxies import current_pages_service
 from invenio_pages.records.errors import PageNotFoundError
-from invenio_rdm_records.proxies import current_community_records_service
+from invenio_rdm_records.proxies import (
+    current_community_records_service,
+    current_rdm_records,
+)
 from invenio_rdm_records.resources.serializers import UIJSONSerializer
 from invenio_records_resources.services.errors import PermissionDeniedError
 
@@ -143,4 +146,35 @@ def community_static_page(pid_value, community, community_ui, **kwargs):
         page=page,
         community=community_ui,
         permissions=permissions,
+    )
+
+
+@pass_community(serialize=True)
+def community_collection(
+    community, community_ui, pid_value, tree_slug, collection_slug
+):
+    """Render a community collection page."""
+    collections_service = current_rdm_records.collections_service
+    collection = None
+    try:
+        collection = collections_service.read_slug(
+            g.identity,
+            community_id=community.id,
+            slug=collection_slug,
+            tree_slug=tree_slug,
+        )
+    except Exception:
+        abort(404)
+
+    if not collection:
+        abort(404)
+
+    collection_dict = collection.to_dict()
+    return render_community_theme_template(
+        "invenio_communities/collections/collection.html",
+        collection=collection,
+        collection_dict=collection_dict,
+        community=community,
+        permissions=community.has_permissions_to(HEADER_PERMISSIONS),
+        theme=community_ui.get("theme", {}),
     )
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,14 @@
 Changes
 =======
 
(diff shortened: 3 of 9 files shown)


invenio_app_rdm/communities_ui/templates/semantic-ui/invenio_communities/collections/collection.html
```diff
@@ -0,0 +1,31 @@
...
+{% extends "invenio_communities/details/base.html" %}
+
+{%- block javascript %}
+{{ super() }}
+{{ webpack['invenio-app-rdm-collections-search.js'] }}
+{%- endblock %}
+
+{%- block page_body %}
+{{ super() }}
+<div class="ui container fluid theme-tertiary page-subheader rel p-10 mr-0-mobile ml-0-mobile">
```
kpsherva, line 19: do we have any alternative for the collection name header? to me it looks too similar to the "Research and innovation" one

invenio_app_rdm/communities_ui/views/communities.py
```diff
@@ -126,3 +129,34 @@ def community_static_page(pid_value, community, community_ui, **kwargs):
...
+        collection = collections_service.read_slug(
+            g.identity,
+            community_id=community.id,
+            slug=collection_slug,
+            tree_slug=tree_slug,
+        )
+    except Exception:
+        abort(404)
+
+    if not collection:
```
kpsherva, line 151: in which case it is possible to return None for the service response?

invenio_app_rdm/communities_ui/templates/semantic-ui/invenio_communities/collections/collection.html
```diff
@@ -0,0 +1,31 @@
...
+  {%- if collection.logo %}
+  <img class="ui rounded image logo" src="{{ collection.logo }}" alt="" />
+  {%- endif %}
+  <h1 class="ui container">
+    {{ collection.title }}
```
anikachurilova, line 24: This looks a bit too big on the screenshots, can we try other ways?

invenio_app_rdm/communities_ui/templates/semantic-ui/invenio_communities/collections/collection.html
```diff
@@ -0,0 +1,31 @@
...
+{%- block javascript %}
+{{ super() }}
+{{ webpack['invenio-app-rdm-collections-search.js'] }}
+{%- endblock %}
+
+{%- block page_body %}
+{{ super() }}
+<div class="ui container fluid theme-tertiary page-subheader rel p-10 mr-0-mobile ml-0-mobile">
+  {%- if collection.logo %}
+  <img class="ui rounded image logo" src="{{ collection.logo }}" alt="" />
```
anikachurilova, line 21: There is no `logo` in the model, how is this getting resolved?

invenio_app_rdm/communities_ui/views/communities.py
```diff
@@ -126,3 +129,34 @@ def community_static_page(pid_value, community, community_ui, **kwargs):
...
+    collections_service = current_rdm_records.collections_service
+    collection = None
+    try:
+        collection = collections_service.read_slug(
+            g.identity,
+            community_id=community.id,
+            slug=collection_slug,
+            tree_slug=tree_slug,
+        )
+    except Exception:
```
anikachurilova, line 148: Can we catch more specific exceptions here? The way to solve this, is to create a specific exceptions error classes on the backend and then catch them here

invenio_app_rdm/communities_ui/views/communities.py
```diff
@@ -126,3 +129,34 @@ def community_static_page(pid_value, community, community_ui, **kwargs):
...
+        abort(404)
```
anikachurilova, line 149: Will this be improved in the following prs?

invenio_app_rdm/communities_ui/views/communities.py
anikachurilova, line 151: Same here, will this be changed in future? We should rely on exceptions and not on the None values

ntarocco:
I have the feeling that the current implementation is not yet rendering the collection search results as in the screenshot, as I don't see how the final query is built. I guess it will done in the next PR.

invenio_app_rdm/communities_ui/views/communities.py
alejandromumo: Definitely, we need better error handling on the Collection service / API. 

Will add it in the next PR

invenio_app_rdm/communities_ui/views/communities.py
alejandromumo: Yes, alongside adding specific errors to the collections backend.

invenio_app_rdm/communities_ui/views/communities.py
alejandromumo: yes, collections backend will have a specific error for this. `None`should not come out from the service layer

alejandromumo:
> I have the feeling that the current implementation is not yet rendering the collection search results as in the screenshot, as I don't see how the final query is built. I guess it will done in the next PR.

The query is computed on the [collection's API](https://github.com/alejandromumo/invenio-rdm-records/blob/7be1ad0301cd6334d24d32d0f5500a058dca3a60/invenio_rdm_records/collections/api.py#L152-L158) . In the next PR we're adding validation to the query (using `luqum.parse`)

invenio_app_rdm/communities_ui/templates/semantic-ui/invenio_communities/collections/collection.html
alejandromumo: For jinja this resolves to a `None`. The problem would be `collection.logo.somethingelse`.

Nonetheless, it's a leftover so thanks for the heads up!

invenio_app_rdm/communities_ui/templates/semantic-ui/invenio_communities/collections/collection.html
alejandromumo: Indeed, it's based on "Research and Innovation" so it can be improved.

Will have a look in a second PR, right now we want to move collections forward to have real data showing up.

invenio_app_rdm/communities_ui/templates/semantic-ui/invenio_communities/collections/collection.html
alejandromumo: Will have a look in a future PR

invenio_app_rdm/communities_ui/views/communities.py
alejandromumo: service should be throwing an exception in case of `CollectionNotFound`. Will be added in the next PR

PR #2875, opened 2024-10-10
assets: fix item description overflow issue

kpsherva:
* addresses mathjax formulas truncation

Forced overflow (formula simply is hidden):
Screenshot 2024-10-10 at 12 30 59

before forcing the overflow
Screenshot 2024-10-10 at 12 31 12

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides
@@ -95,6 +95,7 @@
 
 .ui.items {
 
+
   &.link > .item:hover .content .header {
     color: @darkTextColor;
   }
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides
```diff
@@ -109,6 +110,13 @@
     }
 
     .content {
+      .description.truncate-lines-2 {
```
ntarocco, line 113: Looks like [not very well supported](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp) by browsers (yet)? Did you manage to test it with the mainstream browsers?

kpsherva: I checked https://caniuse.com/css-line-clamp it does not seem to be aligned with mozilla docs

ntarocco: Screenshot 2024-10-10 at 13 54 29

kpsherva: checked on safari, chrome, firefox, opera: works on all of them
mobile: chrome, firefox, duduckgo, safari: works as well. Looks like the docs are not up to date

PR #2894, opened 2024-10-31
admin: increase title length and fix mobile

### LINKED ISSUE #2898: admin UI fixes

carlinmack:

Close #2898 

### Description

**Before**:
image

**After**:
image

---

**Before**:
image

**After**:
image

### Checklist

- [ ] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [ ] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
@@ -51,29 +51,27 @@ class SearchResultItemComponent extends Component {
           data-label="Title"
           className="word-break-all"
         >
-          <div>
-            <BoolFormatter
-              value={result.access.record === "public"}
-              color="green"
-              icon="lock unlocked"
-            />
-            <BoolFormatter
-              value={
-                result.access.record === "restricted" ||
-                result.access.files === "restricted"
-              }
-              color="red"
-              icon="lock"
-            />
-            <a target="_blank" rel="noreferrer" href={result.links.self_html}>
-              {_truncate(result.metadata.title || i18next.t("Empty draft title"), {
-                length: 100,
-              })}
-            </a>
-            <br />
-            <div className="text-muted">
-              {result.id} | {i18next.t("version")} {result.versions.index}
-            </div>
+          <BoolFormatter
+            value={result.access.record === "public"}
+            color="green"
+            icon="lock unlocked"
+          />
+          <BoolFormatter
+            value={
+              result.access.record === "restricted" ||
+              result.access.files === "restricted"
+            }
+            color="red"
+            icon="lock"
+          />
+          <a target="_blank" rel="noreferrer noopener" href={result.links.self_html}>
+            {_truncate(result.metadata.title || i18next.t("Empty draft title"), {
+              length: 100,
+            })}
+          </a>
+          <br />
+          <div className="text-muted">
+            {result.id} | {i18next.t("version")} {result.versions.index}
           </div>
         </Table.Cell>
         <Table.Cell
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
@@ -51,27 +51,29 @@ class SearchResultItemComponent extends Component {
           data-label="Title"
           className="word-break-all"
         >
-          <BoolFormatter
-            value={result.access.record === "public"}
-            color="green"
-            icon="lock unlocked"
-          />
-          <BoolFormatter
-            value={
-              result.access.record === "restricted" ||
-              result.access.files === "restricted"
-            }
-            color="red"
-            icon="lock"
-          />
-          <a target="_blank" rel="noreferrer" href={result.links.self_html}>
-            {_truncate(result.metadata.title || i18next.t("Empty draft title"), {
-              length: 50,
-            })}
-          </a>
-          <br />
-          <div className="text-muted">
-            {result.id} | {i18next.t("version")} {result.versions.index}
+          <div>
+            <BoolFormatter
+              value={result.access.record === "public"}
+              color="green"
+              icon="lock unlocked"
+            />
+            <BoolFormatter
+              value={
+                result.access.record === "restricted" ||
+                result.access.files === "restricted"
+              }
+              color="red"
+              icon="lock"
+            />
+            <a target="_blank" rel="noreferrer" href={result.links.self_html}>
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
```diff
@@ -51,27 +51,29 @@ class SearchResultItemComponent extends Component {
...
-          />
-          <a target="_blank" rel="noreferrer" href={result.links.self_html}>
-            {_truncate(result.metadata.title || i18next.t("Empty draft title"), {
-              length: 50,
-            })}
-          </a>
-          <br />
-          <div className="text-muted">
-            {result.id} | {i18next.t("version")} {result.versions.index}
+          <div>
```
zzacharo, line 54: Why the extra div needed?

carlinmack: To fix the mobile view

kpsherva: do we understand why adding div is fixing it? it seems like a "hack" not an actua fix

carlinmack: So the reason for this behaviour is that to display the column label in every row we override the normal SemanticUI behaviour 

https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/table.overrides#L111-L112

The highlighted code creates a two column grid where each child is put into alternating columns, and to add the column label as the first child, so the behaviour we are seeing is expected. I think it is reasonable to expect that if you only want one item on the right then you need to pass one child

kpsherva: I believe a long term solution should be found, rather than adding divs every time we have several elements in the table cell

carlinmack: I don't think it's ideal, but at the moment you could do something like this if you wanted (imagining that an instance supports subtitles)

| title<br>subtitle | Survey of Web Development<br>Focusing on InvenioRDM |
| ---------------- | --------------- |
| author           | Carlin MacKenzie          |

We could "solve" the problem by doing `grid-template-rows:` or changing it to `inline-flex` but in my view it is a bug that the parent element is missing here

kpsherva: I think an easier solution would be to set
on the `<td>`
```
display: grid 
``` 

and on the pseudo element before:
```
grid-column:1
```

all the other children of td (by `> *` selector):

```
grid-column:2
```

since in the mobile view we will always have 2 columns

no unexpected behaviours should happen in this way (I think it is a bit similar to what you proposed above)

carlinmack: good call! fixed, tested and updated the screenshots above

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/table.overrides
```diff
@@ -108,19 +108,24 @@
             }
 
             &[data-label]  {
-              display: inline-grid !important;
-              grid-template-columns: [col1-start] 30% [col2-start] 70% [end];
+              display: grid;
+              grid-template-columns: [col1-start] 20% [col2-start] 80% [end];
```
carlinmack, line 112: 30% had too much whitespace and 20% doesn't cause the label to go onto two lines on any admin panels

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
```diff
@@ -64,9 +64,9 @@ class SearchResultItemComponent extends Component {
             color="red"
             icon="lock"
           />
-          <a target="_blank" rel="noreferrer" href={result.links.self_html}>
+          <a target="_blank" rel="noreferrer noopener" href={result.links.self_html}>
             {_truncate(result.metadata.title || i18next.t("Empty draft title"), {
-              length: 50,
+              length: 100,
```
kpsherva, line 69: we can replace the truncate with CSS solution https://github.com/inveniosoftware/invenio-app-rdm/commit/c97c82825e5c178cdee623a0c70bb357a641a846 (you would need a new class `truncate-lines-1`
the problem there was that truncate function disregards closing tags completely - for example when used with latex, which causes the formulas to display in a broken way. CSS handles that case way better (tested on many browsers already)

carlinmack: If we want to truncate it we can't have it being inline as it [displays as a -webkit-box](https://github.com/carlinmack/invenio-theme/blob/remove-unused/invenio_theme/assets/semantic-ui/less/invenio_theme/theme/globals/site.overrides#L419)

image

(using `truncated` instead of `truncate-lines-1` does not work as there is no max width as shown)

image

It is weird and unintuitive. Maybe this is why `truncate` in Jinja was chosen

slint: I think longterm, using CSS might be the better option, since:

- there's no information loss
- no JS necessary
- better handling for "split" formulas/phrases/markup
- weirdly also works when searching for text that is truncated (even though you don't see the highlighted result, but I maybe there's a trick for that)

PR #2911, opened 2024-11-15
semantic-ui: users: header: Use endpoints from config

sakshamarora1:
Fixes: https://github.com/inveniosoftware/invenio-communities/issues/1209

### DIFF
--- a/invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html
+++ b/invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html
@@ -23,14 +23,14 @@
       </div>
     </div>
     {% if active_dashboard_menu_item == "uploads" %}
-      <a class="ui tiny button positive left labeled icon m-0" href="/uploads/new">
+      <a class="ui tiny button positive left labeled icon m-0" href={{ url_for("deposit_create") }}>
         <i class="upload icon" aria-hidden="true"></i>
         {{ _('New upload') }}
       </a>
     {% endif %}
 
     {% if active_dashboard_menu_item == "communities" %}
-      <a class="ui tiny button positive left labeled icon m-0" href="/communities/new">
+      <a class="ui tiny button positive left labeled icon m-0" href={{ url_for("communities_new") }}>
         <i class="plus icon" aria-hidden="true"></i>
         {{ _('New community') }}
       </a>


invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html
```diff
@@ -23,14 +23,14 @@ <h1 class="ui medium header" aria-label="{{ _('Your dashboard') }}">{{ current_u
       </div>
     </div>
     {% if active_dashboard_menu_item == "uploads" %}
-      <a class="ui tiny button positive left labeled icon m-0" href="/uploads/new">
+      <a class="ui tiny button positive left labeled icon m-0" href={{ config.APP_RDM_ROUTES['deposit_create'] }}>
```
slint, line 26:
```suggestion
      <a class="ui tiny button positive left labeled icon m-0" href={{ url_for("deposit_create") }}>
```

minor/shelve: in theory it's the same, since the code paths are technically the same. I think we're missing though passing an explicit `endpoint` parameter when registering the routes, which means that they get auto-generated from the function name (which also caused the issue you mentioned). 

Fixing the explicit endpoint naming is a bit out of scope and a bigger task, so I'll shelve it, but starting to use `url_for` where possible would be good.

zzacharo: Out of curiosity, what is the advantage of that? Normalization on how do we access the routes?

zzacharo (approved):

invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html
slint: Yeah, I would prefer using `url_for` and having explicit endpoint names, since this is a Flask thing vs. an Invenio(RDM) thing (i.e. less to learn for someone new to the codebase).

invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html
```diff
@@ -23,14 +23,14 @@ <h1 class="ui medium header" aria-label="{{ _('Your dashboard') }}">{{ current_u
       </div>
     </div>
     {% if active_dashboard_menu_item == "uploads" %}
-      <a class="ui tiny button positive left labeled icon m-0" href="/uploads/new">
+      <a class="ui tiny button positive left labeled icon m-0" href={{ url_for("deposit_create") }}>
```
slint, line 26: @sakshamarora1 this should be actually `invenio_app_rdm_records.deposit_create` (since the blueprint has [this prefix](https://github.com/inveniosoftware/invenio-app-rdm/blob/a4a7240114825386de6a9ec02d5febdfc92d3014/invenio_app_rdm/records_ui/views/__init__.py#L71)). This is also relevant to my comment about being explicit about the endpoint naming.

utnapischtim: is this working in flask/werkzeug <3.0.0 because with flask,werkzeug >3.0 it fails without the `invenio_app_rdm_records` prefix

PR #2922, opened 2024-11-22
semantic-ui: users: header: Pass entrypoint namespace in url_for

sakshamarora1:
related: https://github.com/inveniosoftware/invenio-app-rdm/pull/2911


ptamarit (approved):
Tested. Fixes the bug as expected.

PR #2933, opened 2024-12-05
doi: handle UI for optional DOI feature

anikachurilova:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/163

### DIFF
--- a/invenio_app_rdm/ext.py
+++ b/invenio_app_rdm/ext.py
@@ -33,6 +33,19 @@ def finalize_app(app):
 
 def init_config(app):
     """Initialize configuration."""
+    record_doi_required = (
+        app.config["RDM_PERSISTENT_IDENTIFIERS"].get("doi", {}).get("required")
+    )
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/ext.py
```diff
@@ -33,6 +33,17 @@ def finalize_app(app):
 
 def init_config(app):
     """Initialize configuration."""
+    record_doi_required = app.config["RDM_PERSISTENT_IDENTIFIERS"]["doi"]["required"]
```
zzacharo, line 36: must not assume that doi is configured

invenio_app_rdm/ext.py
```diff
@@ -33,6 +33,17 @@ def finalize_app(app):
...
+    parent_doi_required = app.config["RDM_PARENT_PERSISTENT_IDENTIFIERS"]["doi"][
```
zzacharo, line 37: must not assume that doi is configured

PR #2938, opened 2024-12-10
views: FAIR signposting level 1 support (HTTP Link headers)

### LINKED ISSUE #2937: FAIR signposting level 1 support (HTTP Link headers & link rel item)
We are missing some HTTP `Link` headers on the:
* Landing page
* Content resource (file download)
* Metadata resource (export formats)

See [level 1 recipes](https://signposting.org/FAIR/#level1) and [level 1 examples](https://signposting.org/FAIR/#examples1).

We could also change `alternate` to `item` in  `<link rel="alternate" type="application/pdf" href="https://example.com/records/1234567/files/filename.pdf">`

ptamarit:

### Description

* Fixes #2937
* Depends on inveniosoftware/invenio-rdm-records#1908
* See description in linked ticket

### Checklist

- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).

**Frontend**

**Reminder**

### DIFF
--- a/invenio_app_rdm/records_ui/views/decorators.py
+++ b/invenio_app_rdm/records_ui/views/decorators.py
@@ -11,7 +11,7 @@
 
 from functools import wraps
 
-from flask import g, make_response, redirect, request, session, url_for
+from flask import current_app, g, make_response, redirect, request, session, url_for
 from flask_login import login_required
 from invenio_communities.communities.resources.serializer import (
     UICommunityJSONSerializer,
@@ -19,6 +19,9 @@ from invenio_communities.communities.resources.serializer import (
 from invenio_communities.proxies import current_communities
 from invenio_pidstore.errors import PIDDoesNotExistError
 from invenio_rdm_records.proxies import current_rdm_records
+from invenio_rdm_records.resources.serializers.signposting import (
+    FAIRSignpostingProfileLvl1Serializer,
+)
 from invenio_records_resources.services.errors import PermissionDeniedError
 from sqlalchemy.orm.exc import NoResultFound
 
@@ -365,8 +368,51 @@ def pass_draft_community(f):
     return view
 
 
-def add_signposting(f):
-    """Add signposting link to view's response headers."""
+def _get_header(rel, value, link_type=None):
+    header = f'<{value}> ; rel="{rel}"'
+    if link_type:
+        header += f' ; type="{link_type}"'
+    return header
+
+
+def _get_signposting_collection(pid_value):
+    ui_url = record_url_for(pid_value=pid_value)
+    return _get_header("collection", ui_url, "text/html")
+
+
+def _get_signposting_describes(pid_value):
+    ui_url = record_url_for(pid_value=pid_value)
+    return _get_header("describes", ui_url, "text/html")
+
+
+def _get_signposting_linkset(pid_value):
+    api_url = record_url_for(_app="api", pid_value=pid_value)
+    return _get_header("linkset", api_url, "application/linkset+json")
+
+
+def add_signposting_landing_page(f):
+    """Add signposting links to the landing page view's response headers."""
+
+    @wraps(f)
+    def view(*args, **kwargs):
+        response = make_response(f(*args, **kwargs))
+
+        # Relies on other decorators having operated before it
+        record = kwargs["record"]
+
+        signposting_headers = FAIRSignpostingProfileLvl1Serializer().serialize_object(
+            record.to_dict()
+        )
+
+        response.headers["Link"] = signposting_headers
+
+        return response
+
+    return view
+
+
+def add_signposting_content_resources(f):
+    """Add signposting links to the content resources view's response headers."""
 
     @wraps(f)
     def view(*args, **kwargs):
@@ -374,11 +420,36 @@ def add_signposting(f):
 
         # Relies on other decorators having operated before it
         pid_value = kwargs["pid_value"]
-        signposting_link = record_url_for(_app="api", pid_value=pid_value)
 
-        response.headers["Link"] = (
-            f'<{signposting_link}> ; rel="linkset" ; type="application/linkset+json"'  # fmt: skip
-        )
+        signposting_headers = [
+            _get_signposting_collection(pid_value),
+            _get_signposting_linkset(pid_value),
+        ]
+
+        response.headers["Link"] = " , ".join(signposting_headers)
+
+        return response
+
+    return view
+
+
+def add_signposting_metadata_resources(f):
+    """Add signposting links to the metadata resources view's response headers."""
+
+    @wraps(f)
+    def view(*args, **kwargs):
+        response = make_response(f(*args, **kwargs))
+
+        # Relies on other decorators having operated before it
+        pid_value = kwargs["pid_value"]
+
+        signposting_headers = [
+            _get_signposting_describes(pid_value),
+            _get_signposting_linkset(pid_value),
+        ]
+
+        response.headers["Link"] = " , ".join(signposting_headers)
+
         return response
 
     return view
--- a/tests/ui/test_signposting_ui.py
+++ b/tests/ui/test_signposting_ui.py
@@ -9,25 +9,67 @@
 
 See https://signposting.org/FAIR/#level2 for more information on Signposting
 """
+import pytest
 
 
-def test_link_in_landing_page_response_headers(running_app, client, record):
-    res = client.head(f"/records/{record.id}")
-
-    assert (
-        res.headers["Link"]
-        == f'<https://127.0.0.1:5000/api/records/{record.id}> ; rel="linkset" ; type="application/linkset+json"'  # noqa
-    )
-
-
-def test_link_in_content_resource_response_headers(
-    running_app, client, record_with_file
+@pytest.mark.parametrize("http_method", ["head", "get"])
+def test_link_in_landing_page_response_headers(
+    running_app, client, record_with_file, http_method
 ):
+    client_http_method = getattr(client, http_method)
+    res = client_http_method(f"/records/{record_with_file.id}")
+
+    # The link headers are already tested in details in `invenio-rdm-records` (see `test_signposting_serializer`).
+    # Here we still want to issue the HTTP call to the URL in order to make sure that the decorator is working properly,
+    # but the assertions are less detailed to avoid having to adapt this test every time we modify the logic in `invenio-rdm-records`.
+    link_headers = res.headers["Link"].split(" , ")
+
+    # The test record does not have:
+    # - an author with an identifier.
+    # - a cite-as since it has no DOI.
+    # - a license.
+
+    # There should be at least 10 export formats supported (e.g. "application/dcat+xml", "application/x-bibtex", etc.).
+    assert sum('; rel="describedby" ;' in header for header in link_headers) >= 10
+
+    # There should be at least one file in the record.
+    assert sum('; rel="item" ;' in header for header in link_headers) >= 1
+
+    # There should be at least one description of the type of the record (e.g. "https://schema.org/Photograph").
+    assert sum('; rel="type"' in header for header in link_headers) >= 1
+
+    # There should be at least one link to a linkset (e.g. "application/linkset" and/or "application/linkset+json")
+    assert sum('; rel="linkset" ;' in header for header in link_headers) >= 1
+
+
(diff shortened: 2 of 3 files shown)


invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -365,20 +367,169 @@ def view(**kwargs):
...
+    # Limit authors to the first 10.
+    for creator in islice(record["metadata"]["creators"], 0, 10):
+        for identifier in creator["person_or_org"].get("identifiers", []):
+            if identifier["scheme"] == "orcid":
+                authors.append(
+                    _get_header(
+                        "author", "https://orcid.org/" + identifier["identifier"]
+                    )
+                )
+    return authors
```
ptamarit, line 414: Lars suggested that we might choose to not include authors at all since the list might be long and the full list can be found in the linkset.

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -365,20 +367,169 @@ def view(**kwargs):
...
+    authors = []
+    # Limit authors to the first 10.
+    for creator in islice(record["metadata"]["creators"], 0, 10):
+        for identifier in creator["person_or_org"].get("identifiers", []):
+            if identifier["scheme"] == "orcid":
+                authors.append(
+                    _get_header(
+                        "author", "https://orcid.org/" + identifier["identifier"]
+                    )
+                )
```
ptamarit, line 413: Should we support other schemes like ROR, etc?

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -365,20 +367,169 @@ def view(**kwargs):
...
+
+
+def _get_signposting_licenses(record):
+    licenses = []
+    for right in record["metadata"].get("rights", []):
+        # First try to get `props.url` from the standard licenses,
+        # then try to get the optional `link` from the custom license.
+        url = right.get("props", {}).get("url") or right.get("link")
+        if url:
+            licenses.append(_get_header("license", url))
```
ptamarit, line 435: The FAIR Signposting docs recommends to use SPDX license identifier (e.g. `https://spdx.org/licenses/CC0-1.0`).
However, [in Zenodo we store URLs like `https://creativecommons.org/publicdomain/zero/1.0/legalcode`](https://github.com/zenodo/zenodo-rdm/blob/a11ff414c1a3c3829c26868a38aae3110b3fd25e/app_data/vocabularies/licenses.csv#L101) and not `spdx.org` URLs.

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -365,20 +367,169 @@ def view(**kwargs):
...
+
+
+def _get_signposting_describes(pid_value):
+    ui_url = record_url_for(pid_value=pid_value)
+    return _get_header("describes", ui_url, "text/html")
+
+
+def _get_signposting_linkset(pid_value):
+    api_url = record_url_for(_app="api", pid_value=pid_value)
+    return _get_header("linkset", api_url, "application/linkset+json")
```
ptamarit, line 462: Note: this is required for level 2 support and was already added in a previous pull request.
Here we only include a link of the type `"application/linkset+json"`, but the docs requires to also include a link of type `"application/linkset"`.

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -365,20 +367,169 @@ def view(**kwargs):
...
+def _get_signposting_types(record):
+    resource_type = record["metadata"]["resource_type"]
+    props = get_vocabulary_props(
+        "resourcetypes",
+        [
+            "props.schema.org",
+        ],
+        resource_type["id"],
+    )
+    url_schema_org = props.get("schema.org")
```
ptamarit, line 396: Not sure if there's a better way to do this lookup.
I followed what's done in [`invenio_rdm_records/resources/serializers/signposting/schema.py`](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/signposting/schema.py#L123-L142).

invenio_app_rdm/records_ui/views/decorators.py
ptamarit, line 1: This is quite a lot of methods added to `decorators.py`, should it be moved to a signposting-specific file?

invenio_app_rdm/records_ui/views/decorators.py
lnielsen: Perhaps just check that it is cached so we don't query db on every landing page request

slint: From what I see these are indeed [cached here](https://github.com/inveniosoftware/invenio-vocabularies/blob/ca6e3ae71035cdfe785630fb6648bfa758d048fe/invenio_vocabularies/services/service.py#L120-L122), which is also mentioned [in `get_vocabulary_props`](https://github.com/inveniosoftware/invenio-rdm-records/blob/7e17fad16868ed40c2b65082246dcac77ad45fb1/invenio_rdm_records/resources/serializers/utils.py#L20-L24).

invenio_app_rdm/records_ui/views/decorators.py
slint: :+1: Agree, I thought the was already some signposting-related directory.

invenio_app_rdm/records_ui/views/decorators.py
slint: It might be that the safer option would be to use something like `idutils.to_url(identifier, scheme)` which will consistently produce a link.

invenio_app_rdm/records_ui/views/decorators.py
slint: Maybe we could apply some sensible limit? E.g. if less than 50 authors, include, otherwise don't include at all and basically have people rely on the explicit authors linkset?

invenio_app_rdm/records_ui/views/decorators.py
slint: If `props["scheme"] == "spdx"` I think we can safely generate the URL like `https://spdx.org/licenses/{right["id"]}`. We might have licenses (or even non-SPDX licenses), in which case just using `url` like here would be ok.

tests/ui/test_signposting_ui.py
```diff
@@ -11,23 +11,58 @@
...
 
 
-def test_link_in_landing_page_response_headers(running_app, client, record):
-    res = client.head(f"/records/{record.id}")
+def test_link_in_landing_page_response_headers(running_app, client, record_with_file):
+    ui_url = f"https://127.0.0.1:5000/records/{record_with_file.id}"
+    api_url = f"https://127.0.0.1:5000/api/records/{record_with_file.id}"
+    filename = "article.txt"
+
+    res = client.head(f"/records/{record_with_file.id}")
```
slint, line 19: question/comment: I think the `HEAD` implementation for Flask/Invenio is that we just treat it as a GET request and skip the body of the response. In that case, we're not saving anything in terms of computation/performance (if that was the original goal of just testing the HEAD response only).

IMHO, it's ok to keep as is, since none of the logic done for generating the header links is that much more complex or adds that big of an overhead compared to the rest of the GET response.

slint (approved):
LGTM, some minor comments only

invenio_app_rdm/records_ui/views/decorators.py
ptamarit: - [ ] Move the code to a signposting-related file or directory.

tests/ui/test_signposting_ui.py
ptamarit: The `Link` header should be included in both `GET` and `HEAD`, as stated in the [FAIR Signposting docs](https://signposting.org/FAIR/#typedlinks) says:
> In addition to being available via HTTP GET requests, the HTTP header that contains Link is accessible via the HTTP HEAD request, which only returns transaction metadata not a resource representation. As such machine agents can obtain a map for their journey by issuing a HTTP HEAD even against resources that have access restrictions. All the while saving bandwidth and hence energy.

- [ ] Modify the tests to assert not only `HEAD`, but also `GET`.

invenio_app_rdm/records_ui/views/decorators.py
slint: Ouch, I tried in the browser and copy-pasting URLs for some reason kept the original case... Ok, this is a bummer, I think we'll have to add the original `spdx` ID with the exact case as a `props.spdx_id` field or similar... 

I think it would be fine to shelve this and just use the `url`, depends on whether we want to spend more time to re-import SPDX and update the existing license vocabulary (funnily, the dump we have is from more than a year ago).

invenio_app_rdm/records_ui/views/decorators.py
ptamarit: I'm now relying on [invenio_rdm_records/resources/serializers/signposting/schema.py's serialize_author](https://github.com/inveniosoftware/invenio-rdm-records/blob/938b7cf588d986c8e15de10d5d81baae17e2fddc/invenio_rdm_records/resources/serializers/signposting/schema.py#L35) which picks the first linkable ID.

invenio_app_rdm/records_ui/views/decorators.py
ptamarit: There is now much less code in `decorators.py` now that I rely on `invenio_rdm_records/resources/serializers/signposting/schema.py`.

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -365,20 +366,86 @@ def view(**kwargs):
...
+        # Relies on other decorators having operated before it
+        pid_value = kwargs["pid_value"]
+
+        signposting_headers = [
+            _get_signposting_describes(pid_value),
+            _get_signposting_linkset(pid_value),
+        ]
+
+        response.headers["Link"] = " , ".join(signposting_headers)
+
```
ptamarit, line 448: Note that, unlike the Landing Page which relies on `invenio_rdm_records.resources.serializers.signposting`, the Content Resources and Metadata Resources are *not* relying on `invenio_rdm_records.resources.serializers.signposting` because:
1. `ContentResourceSchema` and `ContentResourceSchema` expect the record to be passed via `context={"record_dict"}` which makes it more difficult to reuse here.
2. The logic is pretty simple to add only the `collection`, `describes` and `linkset` headers, so re-implementing it here is not that bad.

tests/ui/test_signposting_ui.py
```diff
@@ -9,25 +9,73 @@
...
+        f'<{api_url}> ; rel="describedby" ; type="application/vnd.inveniordm.v1.full+csv"',
+        f'<{api_url}> ; rel="describedby" ; type="application/vnd.inveniordm.v1.simple+csv"',
+        f'<{api_url}> ; rel="describedby" ; type="application/x-bibtex"',
+        f'<{api_url}> ; rel="describedby" ; type="application/x-dc+xml"',
+        f'<{api_url}> ; rel="describedby" ; type="text/x-bibliography"',
+        f'<{ui_url}/files/{filename}> ; rel="item" ; type="text/plain"',
+        # The test record does not have a license.
+        '<https://schema.org/Photograph> ; rel="type"',
+        '<https://schema.org/AboutPage> ; rel="type"',
+        f'<{api_url}> ; rel="linkset" ; type="application/linkset+json"',
```
ptamarit, line 47: The logic for the landing page is implemented in `FAIRSignpostingProfileLvl1Serializer` in `invenio-rdm-records` and is already tested there (see inveniosoftware/invenio-rdm-records#1908).
It stills makes sense to at least issue the HTTP call to the endpoint here, to make sure that the decorator is working properly, but maybe the assertion should be less detailed to avoid having to adapt this test every time we modify the other module?

ptamarit:
Just rebased to trigger the tests now that inveniosoftware/invenio-rdm-records#1908 is merged and that [invenio-rdm-records v17.1.0](https://pypi.org/project/invenio-rdm-records/17.1.0/) is released.

ptamarit:
Also squashed all commits into one, since the first implementation was then replaced by one relying on `FAIRSignpostingProfileLvl1Serializer`.

tests/ui/test_signposting_ui.py
ptamarit: I agreed with myself and changed the test to write more generic assertions like:

> There should be at least 10 export formats supported (e.g. "application/dcat+xml", "application/x-bibtex", etc.).

PR #2947, opened 2024-12-18
ui: fix banners overlapping

anikachurilova:
Closes: https://github.com/CERNDocumentServer/cds-rdm/issues/296

Page banner:
Screenshot 2024-12-18 at 10 25 20

"Newer version" record banner:
Screenshot 2024-12-18 at 10 18 02

Error record banner:
Screenshot 2024-12-18 at 10 18 25

"Preview" record banner
Screenshot 2024-12-18 at 10 18 49

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -104,7 +104,7 @@
 
     <!-- PREVIEW HEADER -->
     {% if is_preview %}
-      <div class="ui info flashed bottom attached manage message">
+      <div class="ui info flashed bottom attached manage message record-banner">
         <div class="ui container">
           <div class="header">
             <i class="eye icon"></i>
@@ -127,7 +127,7 @@
     <!-- /PREVIEW HEADER -->
 
     {% if record.is_published and record.links.latest_html and not record.versions.is_latest %}
-      <div class="ui warning flashed bottom attached manage message">
+      <div class="ui warning flashed bottom attached manage message record-banner">
         <div class="ui container">
           <div class="ui relaxed grid">
             <div class="column">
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/message.overrides
```diff
@@ -5,10 +5,12 @@
 
 .ui.flashed.message,
 .ui.form .flashed.message {
-  z-index: 0;
   &.manage {
     padding: 1em 0;
   }
+  &.record-banner {
```
kpsherva, line 11: why is a new class needed? can't we apply this to the flashed message directly?

kpsherva: also, shouldn't it be in invenio-theme?

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -104,7 +104,7 @@
 
     <!-- PREVIEW HEADER -->
     {% if is_preview %}
-      <div class="ui info flashed bottom attached manage message">
+      <div class="ui info flashed bottom attached manage message record-banner">
```
kpsherva, line 107: since you moved the css to invenio theme, you should also propagate the template changes to the invenio-theme templates, otherwise it will be fixed only here in the record details template of app rdm

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/collections/message.overrides
kpsherva: ok sorry, it took me some time to understand the implications.
if I managed to understand correctly, the fix will work only for the record detail page, what if we have the banner in other pages? f.e. community form has an error?
if that is the case could you propose a solution which is more generic and works for all cases, not only for a record?

PR #2952, opened 2025-01-08
views: optional index page registration

kpsherva:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/204

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/theme.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/theme.js
@@ -134,13 +134,15 @@ const handleAuthButtonClick = () => {
 
 $authButton.on({ click: handleAuthButtonClick });
 
-const invenioConfig = JSON.parse(document.body.dataset.invenioConfig);
-const isMathJaxEnabled = invenioConfig?.isMathJaxEnabled;
-if (window.invenio) {
-  window.invenio.onSearchResultsRendered = () => {
-    if (isMathJaxEnabled) {
-      // Re-render mathematical content on the page using MathJax.
-      return window.MathJax?.typeset();
-    }
-  };
+if (document.body.dataset.invenioConfig) {
+  const invenioConfig = JSON.parse(document.body.dataset.invenioConfig);
+  const isMathJaxEnabled = invenioConfig?.isMathJaxEnabled;
+  if (window.invenio) {
+    window.invenio.onSearchResultsRendered = () => {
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/theme/views.py
```diff
@@ -43,8 +43,10 @@ def create_blueprint(app):
         template_folder="templates",
         static_folder="static",
     )
-
-    blueprint.add_url_rule(**create_url_rule(routes["index"], default_view_func=index))
+    if app.config["RDM_FRONTPAGE"]:
```
utnapischtim, line 46: may i ask what the use case is to make the `/` route optional?
what will be rendered then as default if it is not configured?

kpsherva: I would like to have custom view context, see
https://github.com/CERNDocumentServer/cds-rdm/pull/311/files#diff-1a67a7df3625c50413611bb5279e5be98b8b9b870aff748a0dde81062a0e3ceb

slint: For what it's worth, I've left a comment on the [CDS PR](https://github.com/CERNDocumentServer/cds-rdm/pull/313#discussion_r1914481728), suggesting a different approach for overriding views.

I can add this approach to docs, unless, we want to find a better approach so we're not limiting ourselves in the current 2-tuple format for the route values.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/theme.js
```diff
@@ -134,13 +134,15 @@ const handleAuthButtonClick = () => {
...
-const invenioConfig = JSON.parse(document.body.dataset.invenioConfig);
-const isMathJaxEnabled = invenioConfig?.isMathJaxEnabled;
-if (window.invenio) {
-  window.invenio.onSearchResultsRendered = () => {
-    if (isMathJaxEnabled) {
-      // Re-render mathematical content on the page using MathJax.
-      return window.MathJax?.typeset();
-    }
-  };
+if (document.body.dataset.invenioConfig) {
```
ntarocco, line 137: This should always be there. @anikachurilova worked on this.
What was the issue?

kpsherva: it is not there in the frontpage

ntarocco: Then it should be fixed. @anikachurilova can you please check?

PR #2966, opened 2025-01-22
preview: do not mint parent doi if doi is not reserved and doi is opt…

zzacharo:
When doi is optional, then the parent doi should be minted when previewing a record only when a local managed record doi is reserved.

### Before
Screenshot 2025-01-22 at 11 41 25

### After
Screenshot 2025-01-22 at 11 39 35

### DIFF
--- a/invenio_app_rdm/records_ui/views/records.py
+++ b/invenio_app_rdm/records_ui/views/records.py
@@ -178,7 +178,7 @@ def record_detail(
             )
         except ValidationError:
             abort(404)
-        # inject parent doi format for new drafts so we can show in preview
+        # inject parent doi format for new drafts so we can show in preview if parent doi is required
         if current_app.config["DATACITE_ENABLED"]:
             service = current_rdm_records.records_service
             datacite_provider = [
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/records_ui/views/records.py
```diff
@@ -187,11 +187,27 @@ def record_detail(
...
+                    record_doi = record._record.pids.get("doi", {})
+                    is_doi_reserved = record_doi.get(
+                        "provider", ""
+                    ) == "datacite" and record_doi.get("identifier")
+                    if not is_doi_reserved:
+                        should_mint_parent_doi = False
+
+                if should_mint_parent_doi:
+                    datacite_provider = datacite_provider[0]
+                    parent_doi = datacite_provider.client.generate_doi(
```
jrcastro2, line 207: Question: I guess that the users can still change their mind after generating the parent DOI and remove the DOI, resulting in a "orphan" parent DOI? (jsut asking to confirm if I understand this correctly) If that's the case I find it a bit weird, maybe something to fix in another issue, unless I am missing something.

jrcastro2 (approved):
LGTM!
Is a nice improvement, left a question to confirm I understand correctly the issue.

PR #2970, opened 2025-01-28
ui: fix namespace url filter

egabancho:
* The filter was failing with and IndexError when trying to parse fields with no namespace.

### DIFF
--- a/invenio_app_rdm/records_ui/views/filters.py
+++ b/invenio_app_rdm/records_ui/views/filters.py
@@ -166,6 +166,10 @@ def truncate_number(value, max_value):
 def namespace_url(field):
     """Get custom field namespace url."""
     namespace_array = field.split(":")
+    if len(namespace_array) < 2:
+        # Not a namespaced field
+        return None
+
     namespace = namespace_array[0]
     namespace_value = namespace_array[1]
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -160,12 +160,12 @@ def namespace_url(field):
     """Get custom field namespace url."""
     namespace_array = field.split(":")
     namespace = namespace_array[0]
-    namespace_value = namespace_array[1]
     namespaces = current_app.config.get("RDM_NAMESPACES")
 
     if not namespaces.get(namespace):
         return None
 
+    namespace_value = namespace_array[1]
```
max-moser: If I understand correctly, then this move avoids the array index access for fields that don't *have* a namespace (i.e. no `:` in their name)?

Could it be that a field name happens to collide with a namespace name, still causing this to fail?

E.g. having a namespace `test` (as in the tests below), and a field with the name `test`?
This would satisfy the check `if {"test": ...}.get("test")`, still causing this out of bounds error?

egabancho: Good catch!

max-moser (approved):
LGTM, thanks!

PR #2972, opened 2025-01-31
resultView: fix truncateed description

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/343
image

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides
@@ -169,3 +169,7 @@
     align-self: center;
   }
 }
+
+.item > .content{
+  max-width: 100%
+}


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
```diff
@@ -82,7 +82,7 @@ class RecordsResultsListItem extends Component {
         numOtherVersions={numOtherVersions}
       >
         <Item key={key ?? result.id}>
-          <Item.Content>
+          <Item.Content className="max-width-100">
```
kpsherva, line 85: I would add it to item.overrides directly instead

as in
```
.item > .content{
  max-width: 100%
}

it will fix it for any other component which overrides the ResultListItem. 
In case of how you implemented it it will work only for these 3 cases.

jrcastro2: Ok, makes sense, it will affect all the Item cmp, but I guess it's fine

zzacharo (approved):

PR #2975, opened 2025-02-07
cli: add user deletion

ntarocco:


invenio_app_rdm/cli.py
```diff
@@ -93,3 +94,80 @@ def rebuild_all_indices(order):
...
+        click.secho("Aborted.", fg="red")
+        return
+
+    from invenio_access.utils import get_identity
+
+    idty = get_identity(user)
+
+    from invenio_rdm_records.proxies import current_rdm_records_service
+
+    drafts = current_rdm_records_service.search_drafts(idty)
```
utnapischtim, line 122: why is it not possible to delete a user if drafts are still active? in my opinion it should not matter where a user is registered or part of, if the user wants that their personal data should be deleted it should be deleted. the discussion is how to handle that situation in a way that it doesn't crashes the system

invenio_app_rdm/cli.py
```diff
@@ -93,3 +94,80 @@ def rebuild_all_indices(order):
...
+        LoginInformation,
+        SessionActivity,
+        User,
+        userrole,
+    )
+    from invenio_db import db
+    from invenio_oauthclient.models import RemoteAccount, RemoteToken, UserIdentity
+    from invenio_users_resources.proxies import current_users_service
+
+    with db.session.begin_nested():
```
utnapischtim, line 158: this one should be moved to a dedicated api/model class

invenio_app_rdm/cli.py
```diff
@@ -93,3 +94,80 @@ def rebuild_all_indices(order):
...
+        ra = RemoteAccount.query.filter(RemoteAccount.user_id == user_id).one_or_none()
+        if ra:
+            RemoteToken.query.filter(RemoteToken.id_remote_account == ra.id).delete()
+            ra.delete()
+
+        LoginInformation.query.filter(LoginInformation.user_id == user_id).delete()
+        User.query.filter(User.id == user_id).delete()
+
+    db.session.commit()
+    current_users_service.indexer.delete_by_id(user_id)
```
dfdan, line 172: Trying this on v12, I found I have to perform the index purge *before* the datastore operations - it looks like calling .delete_by_id needs to call the DB(?)

(Not just before the commit, I moved it before the begin_nested)

PR #2986, opened 2025-02-24
views: signposting: files: fix filename encoding issues for downloads

### LINKED ISSUE #2937: FAIR signposting level 1 support (HTTP Link headers & link rel item)

ptamarit:
* Bug fixes #2937
    * Causing `TypeError: http header must be encodable in latin1` (might be a problem since Flask v3 only)
* Related to changes in Flask v3 changes in zenodo/zenodo-rdm#1122 and inveniosoftware/invenio-files-rest@a419808c16617125590381b7dc02f41ca49359b2

### DIFF
--- a/invenio_app_rdm/urls.py
+++ b/invenio_app_rdm/urls.py
@@ -28,6 +28,9 @@ Design decisions:
 - Generated urls are absolute for now
 """
 
+import unicodedata
+from urllib.parse import quote
+
 from flask import current_app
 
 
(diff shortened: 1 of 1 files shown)


slint:
We need an issue or TODO so that we "centralize" this helper somewhere... Either `invenio-base` or `invenio-files-rest`.

ptamarit:
`download_url_for` in `invenio-app-rdm` is actually unused.
It's the method in `invenio-rdm-records` which is used (see inveniosoftware/invenio-rdm-records#1965).

PR #2989, opened 2025-03-03
Skip health checks for files that don't have a uri

max-moser:
These files will fail with certainty because there is no on-disk file on which to calculate the MD5 checksum.


fenekku (approved):
LGTM

slint (approved):

PR #2996, opened 2025-03-16
migrate: fix language problem

utnapischtim:
* it could be that language setting in user profile configuration leads
  to be used in the award title language key, which leads to follow up
  errors on v13 after a reindex.

some links:
https://inveniordm.docs.cern.ch/reference/rest_api_awards/

setting:
- go to user profile -> set preferences local to not en
- create an custom award:

```
 {                                                
     "rights": [                                 
         {                                        
             "id": "cc-by-4.0"             
         }                                        
     ],                                           
     "funding": [                            
         {                                        
             "funder": {                     
                 "id": "05ewg0g09"     
             }                                    
         },                                       
         {                                        
             "award": {                      
                 "title": {                       
                     "de": "tst title in deutsch"
                 }                                
             },                                   
             "funder": {                          
                 "id": "02y1pfw65"           
             }                                    
         }                                        
     ],                                           
     "publisher": "Graz University of Technology",
     "publication_date": "2025-06-17"             
 }
 ```
old solution : https://github.com/utnapischtim/invenio-rdm-records/commit/d8b3c3ce70a807e2f15f2dc9a7d46dcda0bdf5d6
current situation: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/datacite/schema.py#L639  only `en` is taken into account

### DIFF
--- a/invenio_app_rdm/upgrade_scripts/migrate_12_0_to_13_0.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_12_0_to_13_0.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2023-2024 CERN.
-# Copyright (C) 2024 Graz University of Technology.
+# Copyright (C) 2024-2025 Graz University of Technology.
 #
 # Invenio-App-RDM is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by utnapischtim** [comment]
the problem is that the award uses the language which is set in profile 

image

and checks if the controlled vocabulary has set this language, if not v13 seems to through an error, but if the language is there, i think it would use it and i would store into the database. there are then scenarios [like](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/datacite/schema.py#L639) where it only checks about the `en` language key. there is another problem on the search/landing page, which i can't find at the moment

to reproduce this you would have to set at least following to your `invenio.cfg` file
```
I18N_LANGUAGES = [
    ('de', _('German')),
]
```

btw `title` in `AwardRelationSchema` [here](https://github.com/caltechlibrary/invenio-vocabularies/blob/5b37548db3f21532577001bcc7a67819083d0baa/invenio_vocabularies/contrib/awards/schema.py#L78)  is a `i18n_string`

tmorrell:
Ah, I've got it. I've tested it and only `de` no longer breaks the landing or search page. It doesn't show up in the serialized DataCite output, but I would say this is a bug that we should fix it. But it kind of brings up a question on how to internationalize the deposit form. What language do we assume folks will be entering?

- If we assume folks are entering the language of their local for the award title, then the current deposit form behavior is fine and we should fix the serializer (like your code fix). No data migration would be necessary.
- If we assume folks are entering english (which is implicitly done for fields like title), then the deposit form should put the title in `en` and then we would need a data migration.

We're probably going to need to re-think the deposit form and languages in light of this, as well as DataCite 4.6 supporting more language tags on different fields.

PR #2988, opened 2025-02-26
urls: integrate invenio_url_for

fenekku:
Part of link generation feature - see https://github.com/inveniosoftware/invenio-base/pull/183

Depends on:
- https://github.com/inveniosoftware/invenio-base/pull/183 merged and released
- https://github.com/inveniosoftware/invenio-app/pull/98 merged and released

~~Doesn't depend on invenio-records-resources or invenio-rdm-records~~. Well it brings all the versions together, so kind of does.

PR #3018, opened 2025-04-01
integrate invenio sitemap

fenekku:
- **urls: integrate invenio_url_for** (ignore as this work was done on that previous one)
- requires: 
  * https://github.com/inveniosoftware/invenio-app-rdm/pull/2988
  * https://github.com/inveniosoftware/invenio-sitemap/pull/1 merged and released
- **sitemap: integrate invenio-sitemap** (just check https://github.com/inveniosoftware/invenio-app-rdm/pull/3018/commits/19607abd295df0b030ae13642b85ceb1d6c382c4)


invenio_app_rdm/communities_ui/sitemap.py
```diff
@@ -0,0 +1,63 @@
...
+            .sort("-updated")
+            # Using preserve_order is fine.
+            # Using Point in Time is a recommended alternative but not
+            # particularly better than this one for our needs. See
+            # https://opensearch.org/docs/latest/search-plugins/searching-data/point-in-time/
+            .params(preserve_order=True)
+            # In keeping with original: only page is looked for
+            # (curation policy is not most relevant)
+            .source(["slug", "updated", "metadata.page"])
+            .scan()
```
fenekku, line 35: This is mostly porting what was in ZenodoRDM with some tweaks.
For instance, `verified` is not a criteria, but public visibility is and sorting syntax changed. 

*Point 1*
About `.params(preserve_order=True)`: the whole sorting aspect (which was ported from what ZenodoRDM did) is arguably unnecessary in the first place. It could be thought to be beneficial in the sense that most recently changed resources are in first sitemaps so discovered first + changes are concentrated in a few sitemaps, such that a crawling bot that keeps history of lastmod wouldn't have to revisit the sitemaps / resources pages not updated since its last crawl. However, the sorting changes the content of each sitemap (+ sitemap index) files, strictly pushing more recent resources into older sitemap files so they will likely have a different more recent lastmod anyway. So it's doubtful those benefits would materialize. Plus, what the googlebot does is opaque / can't rely on undocumented behavior. 

*Point 2*
Usage of document engine is also arguable: this is done asynchronously (worker task), speed of reading is not crucial since result is cached afterwards (and databases are pretty decent at reading these days).

So overall... this could be less intricate, but maybe I am missing something, so I left it as it was before mostly.

*Point 3*
The SitemapSection's interface split in `iter_entities` and `to_dict` reveals its usefulness (not huge mind you) as it allows an Instance to inherit from this class and override just the `iter_entities` part to re-establish `verified` filter.

fenekku:
The other big question about this integration is how it would play with zenodo-rdm's Sitemap implementation. 

- invenio-sitemap registers routes non-conflicting with those of zenodo-rdm (so both would co-exist if zenodo-rdm or invenio-sitemap doesn't do anything)  
- selection of record/community urls is different than zenodo-rdm

Someone from zenodo-rdm should chime in for their perspective.

invenio_app_rdm/theme/views.py
```diff
@@ -36,6 +37,7 @@ def create_blueprint(app):
         static_folder="static",
     )
 
+    # blueprint.add_url_rule(routes["index"], view_func=index)
```
utnapischtim, line 40: maybe remove

invenio_app_rdm/theme/views.py
```diff
@@ -67,7 +69,11 @@ def index():
 
 def robots():
     """Robots.txt."""
-    return current_app.send_static_file("robots.txt")
+    # return current_app.send_static_file("robots.txt")
```
utnapischtim, line 72: mabe remove

invenio_app_rdm/communities_ui/sitemap.py
slint: > Point 1

Indeed, most of these things were in our efforts to make Google Crawls behave better and index more recently created/updated content first... As you mentioned, what Googlebot does in practice is pretty much a black box and might change at any time.

> Point 2

Past implementations were actually doing DB queries, which worked perfectly fine (see [zenodo-rdm](https://github.com/zenodo/zenodo-rdm/commit/aead000855a8a1123f93059f9ca26d2cc26d2a23), and [legacy zenodo](https://github.com/zenodo/zenodo/blob/482ee72ad501cbbd7f8ce8df9b393c130d1970f7/zenodo/modules/sitemap/generators.py#L48)). The reason for the switch was the `is_verified` field filtering, which would have to be computed. We could have done some in-memory caching, but that would unnecessarily complicate things, and the field was already in the search index. Also, holding DB connections for a long time is something we try to avoid, so that was also a quick win.

> Point 3

(continued from above)... Having the `iter_entities` interface is quite nice, since it means one can easily either modify the search filter or even switch implementations entirely to use a DB query.

slint (approved):
LGTM! Thanks for the nice abstraction, indeed as you mentioned for Zenodo we'll have to override the classes and add some filters. IMHO this is fine, and I gladly take it over some tightly coupled config interface to search DSL queries or similar :)

PR #3025, opened 2025-04-11
deposit: add copyrights field

kpsherva:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/381


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -351,6 +352,7 @@ export class RDMDepositForm extends Component {
                         })}
                       />
                     </Overridable>
+                    <CopyrightsField fieldPath="metadata.copyright" />
```
egabancho, line 355: Can you wrap it inside an `Overridable`? ☺️

PR #3027, opened 2025-04-16
Add codespell support (config, workflow to detect/not fix) and make it fix a few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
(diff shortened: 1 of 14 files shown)


ntarocco:
Thank you for your contribution. We appreciate the effort you put into this integration. Given our current number of repositories, we need to carefully consider the best approach for global integration to minimize disruptions across our existing codebase. We'll be discussing this internally to determine the most effective strategy.
For these reasons, we'll be closing this pull request for now.

PR #3035, opened 2025-04-28
records-ui: add error handler for `NoResultFound` exceptions

max-moser:
This is relevant for the case when sharing access to files for unpublished drafts.

## Example

Given there is a draft `87ra6-2bs54` with the file `TU Wien Research Data - 2025.pdf` and the user has access to the draft, then...

Accessing the URL `https://localhost:5000/records/87ra6-2bs54/files/TU Wien Research Data - 2025.pdf?download=1&preview=1` works fine – the file is downloaded as expected.
However, if the `preview=1` flag is omitted from the URL, then it breaks apart with a `500 internal server error` response. :boom: 

Since at TUW we're sending emails about unhandled exceptions to admins, this spams our inbox with this sort of errors.

### DIFF
--- a/invenio_app_rdm/records_ui/views/__init__.py
+++ b/invenio_app_rdm/records_ui/views/__init__.py
@@ -22,6 +22,7 @@ from invenio_records_resources.services.errors import (
     PermissionDeniedError,
     RecordPermissionDeniedError,
 )
+from sqlalchemy.orm.exc import NoResultFound
 
 from invenio_app_rdm.views import create_url_rule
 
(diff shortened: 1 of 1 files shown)


tmorrell (approved):
Thanks, seems like a good idea to catch that error.

Samk13 (approved):
Thanks for this! Maybe adding a test case wouldn't hurt for this one.

egabancho:
Playing the devil's advocate here 😈 
As a "low-level" exception, could we be masking or hiding a legitimate error that should be reported?

max-moser:
@Samk13 agreed – unfortunately, the `create_app()` fixture currently only initializes the API app, so that will require a bit more work.
i've created a follow-up issue for this though: https://github.com/inveniosoftware/invenio-app-rdm/issues/3036

@egabancho the exception is quite low-level indeed, and perhaps extended error handling (in the service layer?) would be a superior approach
however, that would require a lot more (package-spanning) effort, and i think `NoResultFound` is sufficiently aligned with the semantics of a 404 code to warrant this (UI only) error handling :angel:

PR #3044, opened 2025-05-08
deposit: report actual file quota rather than configured values

max-moser:
Previously, the deposit form would get its information about the draft's quota from the RDM-Records `get_file_quota()` function.
That one seems to be intended more for initialization-only logic though, since it only reports configuration values and not the actual effective values.

In this PR, I introduce a new function that tries to get the effective values from the draft first and only afterwards falls back to the configuration values.
That way, the displayed values are more aligned with reality.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@
 Changes
 =======
 
+Version <next>
+
+- deposit: show actual effective quota for the draft rather than config values, if available
+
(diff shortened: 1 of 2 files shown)


tmorrell (approved):
This makes a lot of sense, and we should get it in before v13.

slint (approved):

tmorrell:
I've merged in https://github.com/inveniosoftware/invenio-app-rdm/pull/3074

PR #3057, opened 2025-05-16
templates: add jinja blocks

jrcastro2:


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -381,6 +398,10 @@ <h4>{{ _("Request access") }}</h4>
               </section>
             {%- endblock record_details -%}
 
+            {%- block after_record_details -%}
```
ntarocco, line 401:
```suggestion
            {%- block after_record_additional_details -%}
```

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -274,11 +275,19 @@
...
+            {%- block after_record_title -%}
+              {# After record title hook #}
+            {%- endblock after_record_title -%}
+
             {# Description #}
             {%- block record_content -%}
               {%- include "invenio_app_rdm/records/details/description.html" %}
             {%- endblock record_content -%}
 
+            {%- block after_record_content -%}
```
ntarocco, line 287:
```suggestion
            {%- block after_record_descriptions -%}
```

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -274,11 +275,19 @@
               </section>
             {%- endblock record_title -%}
 
+            {%- block after_record_title -%}
+              {# After record title hook #}
+            {%- endblock after_record_title -%}
```
slint, line 280: minor/question: wouldn't it also work to just override `record_title` and added after?

```jnja
{% block record_title %}

{# {# Things here are the same as a "before_record_title" block #} #}

{{ super() }}

{# Things here are the same as a "after_record_title" block #}
{% endblock record_title %}
```

My concern is that we already lack a solid "Blocks API" for our templates, and adding more names/IDs might mean a rougher path to standardisation.

PR #3058, opened 2025-05-16
beat: add delete job logs recurrent task

jrcastro2:
NEEDS https://github.com/inveniosoftware/invenio-jobs/pull/80

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -481,6 +481,10 @@ CELERY_BEAT_SCHEDULE = {
         "task": "invenio_rdm_records.requests.access.tasks.clean_expired_request_access_tokens",
         "schedule": crontab(minute=4, hour=0),
     },
+    "delete-job-logs": {
+        "task": "invenio_jobs.logging.tasks.delete_logs",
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/config.py
```diff
@@ -481,6 +481,10 @@ def files_rest_permission_factory(obj, action):
         "task": "invenio_rdm_records.requests.access.tasks.clean_expired_request_access_tokens",
         "schedule": crontab(minute=4, hour=0),
     },
+    "delete-job-logs": {
+        "task": "invenio_jobs.logging.tasks.delete_logs",
+        "schedule": crontab(minute=5, hour=0),
```
kpsherva, line 486: seems a bit too frequent :)

jrcastro2: This runs every day at night, deleting logs older than what specified in the config var `JOBS_LOGGING_RETENTION_DAYS` (90 days by default). Why do you think is too frequent?

kpsherva: ouf my mistake, I looked only on the minutes config

kpsherva (approved):

PR #3062, opened 2025-05-22
Add permissions as conf

jrcastro2:


ntarocco (commented):
I am ok with this change, however @utnapischtim and @fenekku were working on something similar.
It would be nice to have their opinions too (even if merged because needed).

PR #3065, opened 2025-05-23
Add record to requests

jrcastro2:


zzacharo (approved):

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -233,7 +233,8 @@ def record_detail(
     return render_community_theme_template(
         current_app.config.get("APP_RDM_RECORD_LANDING_PAGE_TEMPLATE"),
         theme=theme,
-        record=record_ui,
+        record=record,
```
utnapischtim, line 236: why do you add both?
why not refactoring in a way that only `record` or `record_ui` is necessary

jrcastro2: Mostly because they are different things, on one side we have the record object which can be used f.e. to compute the permissions, which wouldn't be possible with record_ui, which is simply a dict of the record computed with the UIJsonSerializer.

So to me it's a bit clearer to separate both concerns, in one hand to have the object to be used to compute whatever is needed and on the other hand to have the object that is used to display.

Also the alternative to me looks like we would need to combined the 2 objects in one specific to be passed to the view.

PR #3067, opened 2025-06-03
deposit UI: add overridable hook for a separate section

kpsherva:


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -377,6 +377,23 @@ export class RDMDepositForm extends Component {
                     />
                   </AccordionField>
                 </Overridable>
+                <Overridable
+                  id="InvenioAppRdm.Deposit.BasicInformation.afterHook.container"
```
utnapischtim: question: what is the idea behind the naming?

kpsherva: I would like to gather some feedback about the naming :)
The idea behind the whole Overridable id is that we need a placeholder for any custom section (AccordionField) that we want to add to the form

slint: I feel the "hook" wording is not adding anything to the ID. Some proposals:

- Plain: `InvenioAppRdm.Deposit.BasicInformation.{before/after}.container`
- Section: `InvenioAppRdm.Deposit.BasicInformation.{before/after}Section.container`
- Slot: `InvenioAppRdm.Deposit.BasicInformation.{before/after}Slot.container`

PR #3069, opened 2025-06-04
details: pass scheme on related identifiers url generator

zzacharo:
closes https://github.com/CERNDocumentServer/cds-rdm/issues/484

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/detail.html
@@ -162,7 +162,7 @@
       {% endif %}
 
       {% set ident_val = identifier.identifier|default('') %}
-      {% set url = ident_val|pid_url %}
+      {% set url = ident_val|pid_url(scheme=identifier.scheme) %}
       {% if url %}
         <a href="{{ url }}" target="_blank" title="{{ _('Opens in new tab') }}">
           {{ ident_val }}


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/detail.html
jrcastro2, line 1: I think this is also used here: https://github.com/inveniosoftware/invenio-app-rdm/blob/29d22f614fafef78feca471ee572daace2137fd8/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/detail.html#L261 should we also update it?

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/detail.html
```diff
@@ -162,7 +162,7 @@ <h2>
       {% endif %}
 
       {% set ident_val = identifier.identifier|default('') %}
-      {% set url = ident_val|pid_url %}
+      {% set url = ident_val|pid_url(identifier.scheme) %}
```
jrcastro2: nit: Probably clearer:
```suggestion
      {% set url = ident_val|pid_url(scheme=identifier.scheme) %}
```

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/detail.html
slint: I would say no unfortunately, because I'm not sure where you would get the `scheme` value for each custom field... If for your custom field you know it's always going to be a specific `scheme`, maybe we could introduce some `field_cfg.props.identifier_scheme` to use.

slint (approved):

PR #3072, opened 2025-06-11
upgrade_scripts: handle thesis custom field migration

slint:


invenio_app_rdm/upgrade_scripts/migrate_12_0_to_13_0.py
```diff
@@ -43,7 +43,17 @@ def update_record(record):
...
-            # TODO: Add any record datamodel migration code here
+            #
+            # Record datamodel migration
+            #
+
+            # Custom fields
+            custom_fields = record.get("custom_fields", {})
+
+            # Thesis field
+            if university := custom_fields.pop("thesis:university", None):
```
tmorrell: I wonder if we should try to query for records that have thesis:university instead. This is going to iterate over all records, when most people probably haven't used the thesis:university field. It's going to add lots of time to the migration, and in most cases isn't going to do anything.

slint: Agree, and actually for Zenodo doing search queries is roughly how we run this "live". If we don't have any other data migration PRs, I could have a go at converting this recipe into what we run:


```python
from invenio_access.permissions import system_identity
from invenio_rdm_records.proxies import current_rdm_records_service as records_service
from invenio_search.engine import dsl
from invenio_db import db

def migrate_thesis_university(record_or_draft):
    custom_fields = record_or_draft.get("custom_fields", {})
    university = custom_fields.get("thesis:university")
    if university and "thesis:thesis" not in custom_fields:
        custom_fields["thesis:thesis"] = {"university": university}

    record_or_draft.commit()

# Common query filter
has_thesis = dsl.Q("exists", field="custom_fields.thesis:university")

# Handle published records
published_records = records_service.search(system_identity, extra_filter=has_thesis)
for result in published_records.hits:
    record = records_service.record_cls.pid.resolve(result["id"])
    migrate_thesis_university(record)

# Handle draft records
draft_records = records_service.search_drafts(system_identity, extra_filter=has_thesis)
for result in draft_records.hits:
    draft = records_service.draft_cls.pid.resolve(result["id"], registered_only=False)
    migrate_thesis_university(draft)

db.session.commit()
```

utnapischtim: i am against that version. i will not say we did the upgrade script always that way, since hundred and hundred of years, but the argument is that i think it is expect able how the upgrade scripts work. they iterate over all records/drafts and apply the metadata updates and commit afterwards. doing it the way above we would open up a new way where we would have to write for every new metadata upgrade a new query and further code.

tmorrell: Iterating over all the records doesn't work for our repositories currently, it's simply too slow. I know this is how we've done upgrade scripts historically, but Zenodo doesn't do things this way and I think it's not going to be sustainable as we get larger repositories on the platform.

This example highlights the issue of the approach. Everyone is going to be iterating over records to change fields they don't have. It's a waste of time during the v13 upgrade because there aren't metadata field changes that impact every record.

From a practical perspective, the thesis upgrade isn't a big deal for us because we already have our fields structured this way. But I do think we need to plan to change our approach for v14.

utnapischtim: fair point. maybe we should think of to update the approach already for v13.

slint: If the only actual data migration we have is for the thesis field, I would consider it, since it really would make the migration process much quicker.

I think somehow we managed to go through an entire InvenioRDM version without any major datamodel changes (I won't say it tells us something about having a "mature" datamodel 😏).

utnapischtim: i am sorry to somehow crash the party ;) but i have a PR open: https://github.com/inveniosoftware/invenio-app-rdm/pull/2996/files which would add another data migration to the upgrade script. but it is not really a data migration it is more a fixing a bug thing and i am still not sure if it was only a problem of our instance or an overall problem for instances with not only en as language

invenio_app_rdm/upgrade_scripts/migrate_12_0_to_13_0.py
```diff
@@ -20,86 +20,107 @@
...
+    record_error_counter = 0
+    draft_success_counter = 0
+    draft_error_counter = 0
 
-    Please read the disclaimer on this module before thinking about executing
-    this function!
-    THIS MODULE IS WORK IN PROGRESS, UNTIL official v13 release
-    """
+    # Handle published records
+    published_records = records_service.search(system_identity, extra_filter=has)
```
slint:
```suggestion
    published_records = records_service.search(
        system_identity, 
        params={"allversions": True, "include_deleted": True},
        extra_filter=has,
    )
```

- A bit of a weird footgun for the search service calls that it doesn't include all versions by default.
- Regarding deleted records, I think it's nice to always have valid records in the system (e.g. in case they're restored)

invenio_app_rdm/upgrade_scripts/migrate_12_0_to_13_0.py
```diff
@@ -20,86 +20,107 @@
...
+        try:
+            migrate(record)
+            record_success_counter += 1
+        except Exception as error:
+            secho(f"> Error {repr(error)}", fg="red")
+            error = f"Record {record.pid.pid_value} failed to update"
+            record_error_counter += 1
+
+    # Handle draft records
+    draft_records = records_service.search_drafts(system_identity, extra_filter=has)
```
slint: Here only the `allversions: True` filter needs to be applied

PR #3097, opened 2025-07-07
set default parent_doi_required to False

mfenner:

### Description

This fixes an issue in v13 if RDM_PARENT_PERSISTENT_IDENTIFIERS = {}

### Checklist

- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

### DIFF
--- a/invenio_app_rdm/ext.py
+++ b/invenio_app_rdm/ext.py
@@ -37,7 +37,9 @@ def init_config(app):
         app.config["RDM_PERSISTENT_IDENTIFIERS"].get("doi", {}).get("required")
     )
     parent_doi_required = (
-        app.config["RDM_PARENT_PERSISTENT_IDENTIFIERS"].get("doi", {}).get("required")
+        app.config["RDM_PARENT_PERSISTENT_IDENTIFIERS"]
+        .get("doi", {})
+        .get("required", False)
     )
 
     if record_doi_required != parent_doi_required:


invenio_app_rdm/ext.py
```diff
@@ -37,7 +37,9 @@ def init_config(app):
         app.config["RDM_PERSISTENT_IDENTIFIERS"].get("doi", {}).get("required")
```
utnapischtim, line 37: should we use here also a default `False`? it seems strange to not use the same default value

tmorrell: Probably. There is a slight difference in that we recommend in the docs that there can be no parent persistent identifiers. But I don't see why the default shouldn't be `False` for both.

utnapischtim: since the documentation normally sets the `required` to `False` or `True` the default value should also be a `bool` type

mfenner: The documentation about RDM_PERSISTENT_IDENTIFIERS is explicit about the configuration https://inveniordm-dev.docs.cern.ch/operate/customize/dois/#parent-or-concept-dois

I haven't looked into what happens if required is omitted, or other settings are left out. My main goal was to have the v12 configuration as documented work for v13.

fenekku: Not to stir the pot more, but since it's in the needed column... The issue seems to be that we need to have boolean values to compare against each other viably. Casting to `bool` is the most effective and preserves the documented behavior:
 
```suggestion
        bool(app.config["RDM_PERSISTENT_IDENTIFIERS"].get("doi", {}).get("required"))
```

(+ suggested cast below)

Additionally, passing a default `True`/`False` to `get` (e.g., `.get("required", True)`) wouldn't deal with a passed `"required": None`, but this does. (not that we should have to always account for any configuration input type/structure, but in this case it's simple and effective).

If no movement or objection on this, I will push changes by tomorrow.

invenio_app_rdm/ext.py
```diff
@@ -37,7 +37,9 @@ def init_config(app):
...
     )
     parent_doi_required = (
-        app.config["RDM_PARENT_PERSISTENT_IDENTIFIERS"].get("doi", {}).get("required")
+        app.config["RDM_PARENT_PERSISTENT_IDENTIFIERS"]
+        .get("doi", {})
+        .get("required", False)
     )
```
fenekku, line 43:
```suggestion
parent_doi_required = bool(
        app.config["RDM_PARENT_PERSISTENT_IDENTIFIERS"]
        .get("doi", {})
        .get("required")
)
```

fenekku (commented):
Moving this along.

invenio_app_rdm/ext.py
mfenner: I like this approach. It is clearer, as falsy values (None, 0, etc.) can be confusing.

fenekku: I've encountered this earlier: the branch of this PR comes from an organization with closed branch access, so I can't push a commit to that branch. I've made the change as a separate commit here: https://github.com/inveniosoftware/invenio-app-rdm/pull/3102 . We can merge that one, or you can make the change to this branch or I can be given access to this branch. 

(The confusing part has always been that for individuals — just not for orgs —
 pushing to their branch is typically allowed, so in practice it's an uneven terrain when it comes to building on someone else's work)

mfenner: Does it mean I should do PRs from my personal account? What do other people/orgs contributing to InvenioRDM packages do?

fenekku: It's not a problem that was encountered / discussed much, so I think others also use orgs and we get the same situation there too :sweat_smile: . Using org forks do seem like the natural thing to do though, so asking for personal forks seems a bit much to me. The problem really only shows up when someone other than original author tries to contribute to the same branch (should we be less helpful? hahaha!). I will bring this up at next maintainer call if we have time (I can see many different outcome)

mfenner: Thanks! Will keep this in mind with the next PR.

PR #3117, opened 2025-07-21
Fix: enable Overridden components in frontpage and community records search

mfenner:
### Description

React components can't be overridden in the frontpage and community records search because the overriddenComponents import is broken (doesn't import the components). This PR fixes this. The PR also makes the appName in the frontpage consistent with the rest of InvenioRDM (InvenioAppRDM.RecordsList -> InvenioAppRdm.RecordsList).

The issue and fix was discussed in Discord in September 2024, and confirmed to be working for the [Rogue Scholar ](https://rogue-scholar.org/)InvenioRDM instance.

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [ ] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
[...]

PR #3120, opened 2025-07-22
fix: import `overridable-registry.js` in `records_list` macro

palkerecsenyi:
**This is an alternative fix for #3117** (thanks for opening this @mfenner!)

---

### Description

* The `records_list` macro was importing `invenio-app-rdm-frontpage.js` early, before any other component-specific JS was imported. This is necessary, as the macro can be used in a range of places, so we need to make sure its associated JS is imported wherever it is used.

* Since we were importing this JS before `overridable-registry`, the loop to add the overriden components to the singleton store: https://github.com/inveniosoftware/invenio-app-rdm/blob/979d6ab4c84fffaa8c70b2b32a40efac179bfd98/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/overridableRegistry/index.js#L11-L13 was being run _after_ the frontpage JS ran the `getAll` function on the store. This meant user-specified overrides were not included in the object (which is a deep-cloned value rather than a reference), so failed to get rendered correctly on the frontpage.

* This PR adds the `overridable-registry.js` import to the `records_list` macro. Even though this will result in a page including the import multiple times, the import behaves as a singleton so it won't actually be imported more than once. This way, we ensure the import is in the correct location regardless of where the macro is used.

* We also avoid using a pass-by-reference for the override object, which could cause some errors and conflicts if components try to accidentally modify it down the line. At the same time, this should fix the issue mentioned [in the original Discord conversation](https://discord.com/channels/692989811736182844/1224742258415239269/1283002968970362975).

* **I have not included the [ID name change](https://github.com/inveniosoftware/invenio-app-rdm/pull/3117/files#diff-36165362d41c45111f91800c4ccf6659e35d2cefbe890859fe76b35519b41310R15) from the original PR**. While the current ID prefix indeed does not follow the naming convention, changing it would likely introduce a breaking change in case anyone is already relying on this. However I'm still happy to change this if we deem the risk to be sufficiently low :)

### DIFF
--- a/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/macros/records_list.html
+++ b/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/macros/records_list.html
@@ -45,5 +45,6 @@
     {% endfor %}
   </div>
 
-  {{ webpack["invenio-app-rdm-frontpage.js"]}}
+  {{ webpack['overridable-registry.js'] }}
+  {{ webpack['invenio-app-rdm-frontpage.js']}}
 {% endmacro %}


zzacharo (approved):

invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/macros/records_list.html
```diff
@@ -45,5 +45,6 @@ <h2 class="ui header">{{ title }}</h2>
     {% endfor %}
   </div>
 
-  {{ webpack["invenio-app-rdm-frontpage.js"]}}
+  {{ webpack['overridable-registry.js'] }}
```
utnapischtim, line 48: why is this added to the macro and not to one of the `base.html` files we have. it seems strange to me to add it to a macro even if the singleton is called only once

palkerecsenyi: My idea was that by adding it to the macro we effectively declare it as a 'dependency', similarly to how we have already done with `invenio-app-rdm-frontpage.js`.

Right now, we're only using this macro within `frontpage.html`, but an implementor of RDM may choose to add it somewhere else within their templates and this is a simple way we can ensure the associated JS is always loaded in the correct location. Indeed we also import `overridable-registry.js` in [`javascript.html`](https://github.com/inveniosoftware/invenio-app-rdm/blob/979d6ab4c84fffaa8c70b2b32a40efac179bfd98/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/javascript.html#L12), but the macro imports `invenio-app-rdm-frontpage.js` _before_ that's rendered. We must import the frontpage JS file as that file is unique to the macro so it shouldn't be imported elsewhere. But importing it early means we must also import the `overridable-registry.js` just before.

Hope this makes sense!

utnapischtim: i think i don't understand what `overridable-registry.js` does. i thought that this file enables the overridables for the whole page. if that is true it is still strange to me to put it into a macro. i thought since it it is something essential for the whole page it should be more prominent placed.

palkerecsenyi: Indeed `overridable-registry.js` enables overridables for the whole page. Currently, we are loading `javascript.html` always at the [bottom of the body](https://github.com/inveniosoftware/invenio-theme/blob/f0f7aaae389a870737c1930fdb2cf95c478016dd/invenio_theme/templates/semantic-ui/invenio_theme/page.html#L117-L132) (as doing it before the body would mean React fails to render at all). The macro is rendered within the main body content, so before the JS. However, we import the macro-specific `invenio-app-rdm-frontpage.js` inside the macro, meaning the import is placed in the body before the rest of the JS imports. As a result, `invenio-app-rdm-frontpage.js` is loaded before `overridable-registry.js`.

Some other fixes I considered were:

- Making `records_list` add the JS to the pre-declared `javascript` block that would send it to the bottom of the page
  - This does not work, as you cannot access the parent block context from within a macro

- Moving the content of `records_list` to be inline in `frontpage.html`, which is acceptable as the Invenio modules do not use it elsewhere.
  - But it might be depended on by third-party Invenio implementations and this would lose the reproducibility aspect.

- Tell implementors to manually import the associated JS whenever they use the macro, as shown in the example in [my comment](https://github.com/inveniosoftware/invenio-app-rdm/pull/3117#issuecomment-3097266871)
  - But this could be quite counter-intuitive for developers, it requires some additional documentation and removes the tight connection between the macro's HTML and associated JS.

This is just the best solution I could think of for now, but it's entirely possible there's a neater way to do it, I agree this isn't perfect! Let me know if you have any other ideas, I'm happy to experiment with different methods.

utnapischtim: thank you for the detailed explanation. yes i think too that the solution is not the optimal, but since i have no other solution for this problem i am fine with it. i will merge it.

mfenner:
This fix works for me (deploying 13.0.0rc7 in the Rogue Scholar instance). Thank you @palkerecsenyi and @utnapischtim !

PR #3127, opened 2025-07-22
templates: macros: detail: Remove programme link

sakshamarora1:
fixes: https://github.com/CERNDocumentServer/cds-rdm/issues/527

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/detail.html
@@ -214,12 +214,7 @@
 
 {% macro list_vocabulary_values(field, values, field_cfg) %}
   {% if values.title_l10n is defined %}
-    {% set search_url = field | custom_fields_search(values.title_l10n, field_cfg) %}
-    {% if search_url %}
-      <a href="{{ search_url }}">{{ values.title_l10n | safe }}</a>
-    {% else %}
-      {{ values.title_l10n }}
-    {% endif %}
+    {{ values.title_l10n }}
   {% else %}
     {% for value in values %}
       {{ value.title_l10n }}{{ ", " if not loop.last }}


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/detail.html
```diff
@@ -214,12 +214,7 @@ <h2>
 
 {% macro list_vocabulary_values(field, values, field_cfg) %}
   {% if values.title_l10n is defined %}
-    {% set search_url = field | custom_fields_search(values.title_l10n, field_cfg) %}
-    {% if search_url %}
-      <a href="{{ search_url }}">{{ values.title_l10n | safe }}</a>
-    {% else %}
-      {{ values.title_l10n }}
-    {% endif %}
+    {{ values.title_l10n }}
```
zzacharo: Why do we remove the search link from all vocabularies?

sakshamarora1: It is only for the custom_fields, this macro is being called by `show_section_custom_fields` only, maybe better to rename this macro to `list_custom_fields_vocabulary_values`?
The link redirects to a query string search which doesn't work in the records search, [see example](https://repository.cern/search?q=custom_fields.cern%5C%3Aprogrammes%3ACERN%20Doctoral%20Student%20Program&l=list&p=1&s=10&sort=bestmatch)

zzacharo: Discussed IRL: It needs some discussion, to try to find a proper solution instead of removing the feature...

invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -190,9 +190,12 @@ def custom_fields_search(field, field_value, field_cfg=None):
         namespace_string = r"\:".join(namespace_array) + f".{localised_title}.{locale}"
     else:
         namespace_string = r"\:".join(namespace_array)
+        if (field_cfg or {}).get("link_by_id", False):
```
zzacharo: why not something like

```suggestion
        if (field_cfg or {}).get("landing_page_search_attr", False):
            namespace_string += f".{landing_page_search_attr}"
```

naming can be changed ofc....I am thinking more on allowing users to configure the attribute they would like to search on. The default can remain as it was with either the localized title or just the raw value. I am afraid that is too rigid to have a property only for id. @kpsherva @ntarocco wdyt?

ntarocco: I agree, more flexible

invenio_app_rdm/records_ui/views/filters.py
```diff
@@ -190,9 +190,14 @@ def custom_fields_search(field, field_value, field_cfg=None):
         namespace_string = r"\:".join(namespace_array) + f".{localised_title}.{locale}"
     else:
         namespace_string = r"\:".join(namespace_array)
+        # Check if the field config has a landing page search attribute to search by
+        landing_page_search_attr = (field_cfg or {}).get("landing_page_search_attr", False)
+        if landing_page_search_attr:
+            namespace_string += f".{landing_page_search_attr}"
```
zzacharo: this should be 
```suggestion
            namespace_string += r".".format({landing_page_search_attr})
```
no? To avoid the warning?

zzacharo (approved):

PR #3132, opened 2025-07-23
fix(deposits): update doi default option only if managed

sakshamarora1:
fixes: https://github.com/CERNDocumentServer/cds-rdm/issues/523

If the DOI is set as not required in the config and the record is published with 'No, I don't need one' selected, editing of that record will not change the DOI checkbox option. It should only update it to 'No, I need one' when the record already has a DOI.

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -84,10 +84,12 @@ def get_form_pids_config(record=None):
                 optional_doi_transitions["message"] = optional_doi_transitions.get(
                     "message"
                 ).format(sitename=sitename)
-                if set(optional_doi_transitions["allowed_providers"]) - set(
-                    ["external", "not_needed"]
+                if (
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -84,10 +84,12 @@ def get_form_pids_config(record=None):
                 optional_doi_transitions["message"] = optional_doi_transitions.get(
                     "message"
                 ).format(sitename=sitename)
-                if set(optional_doi_transitions["allowed_providers"]) - set(
-                    ["external", "not_needed"]
+                if (
+                    "external" not in optional_doi_transitions["allowed_providers"]
```
zzacharo, line 88:
```suggestion
if all(provider not in optional_doi_transitions["allowed_providers"] for provider in ("external", "not_needed")):
```

zzacharo (approved):

PR #3136, opened 2025-07-25
UI: requests: add template for record deletion

yashlamba:
- Added template for record deletion request type
- Fixed an issue with resolving deleted record requests.

### DIFF
--- a/invenio_app_rdm/requests_ui/templates/semantic-ui/invenio_requests/record-deletion/index.html
+++ b/invenio_app_rdm/requests_ui/templates/semantic-ui/invenio_requests/record-deletion/index.html
@@ -1,0 +1,88 @@
+{# -*- coding: utf-8 -*-
+
+  This file is part of Invenio.
+  Copyright (C) 2016-2025 CERN.
+
+  Invenio is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+#}
+
+{% extends "invenio_requests/details/index.html" %}
+
+{% set active_dashboard_menu_item = 'requests' %}
+{% set active_community_header_menu_item = 'requests' %}
+
+{%- block request_header %}
+    {% set back_button_url = url_for("invenio_app_rdm_users.requests") %}
+  {% from "invenio_requests/macros/request_header.html" import inclusion_request_header %}
+  {{ inclusion_request_header(
+      request=invenio_request,
+      record=record,
+      accepted=request_is_accepted,
+      back_button_url=back_button_url,
+      back_button_text=_("Back to requests")
+    ) }}
+{%- endblock request_header %}
+
+{% block request_timeline %}
+  <div
+    class="ui container rdm-tab-container fluid rel-pt-2 ml-0-mobile mr-0-mobile"
+    id="request-request-deletion-tab-container"
+  >
+    <div
+      class="ui secondary pointing menu rdm-tab-menu"
+      id="request-deletion-request-tab"
+    >
+      <a
+        class="active item"
+        data-tab="conversation"
+        role="tab"
+        aria-selected="true"
+        aria-controls="conversation-tab-panel"
+        id="conversation-tab"
+      >
+        {{ _("Conversation") }}
+      </a>
+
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/requests_ui/views/requests.py
slint, line 1: nit: commit messsage should be `requests: fix deleted record resolution on requests`

invenio_app_rdm/requests_ui/templates/semantic-ui/invenio_requests/record-deletion/index.html
```diff
@@ -0,0 +1,114 @@
...
+          role="tab"
+          class="item"
+          data-tab="checks"
+          aria-selected="false"
+          aria-controls="checks-tab-panel"
+          id="checks-tab"
+        >
+          <i class="{% include 'invenio_checks/requests/overall_severity_level.html' %}"></i> {{ _("Checks") }}
+        </a>
+      {% endif %}
```
slint: minor: I guess this is not needed? I know we discussed about automated checks for support requests, but I would add it once we have the use-case.

invenio_app_rdm/requests_ui/templates/semantic-ui/invenio_requests/record-deletion/index.html
```diff
@@ -0,0 +1,114 @@
...
+      accepted=request_is_accepted,
+      back_button_url=back_button_url,
+      back_button_text=_("Back to requests")
+    ) }}
+{%- endblock request_header %}
+
+{% block request_timeline %}
+  <div
+    class="ui container rdm-tab-container fluid rel-pt-2 ml-0-mobile mr-0-mobile"
+    id="request-community-submission-tab-container"
```
slint: minor: the `id` looks wrong here (I hope we don't have any CSS referencing it like this 😅)

invenio_app_rdm/requests_ui/templates/semantic-ui/invenio_requests/record-deletion/index.html
```diff
@@ -0,0 +1,114 @@
...
+#}
+
+{% extends "invenio_requests/details/index.html" %}
+
+{% set active_dashboard_menu_item = 'requests' %}
+{% set active_community_header_menu_item = 'requests' %}
+
+{%- block request_header %}
+    {% set back_button_url = url_for("invenio_app_rdm_users.requests") %}
+  {% from "invenio_requests/macros/request_header.html" import inclusion_request_header %}
```
slint, line 17: minor/nit: I guess this is already in other places, but the fact that we import and `inclusion_request_header` Jinja macro from `invenio-requests` sounds wrong. We should have a "generic" header macro/template, that just renders the request title and use it here.

invenio_app_rdm/requests_ui/templates/semantic-ui/invenio_requests/record-deletion/index.html
```diff
@@ -0,0 +1,114 @@
...
+        class="ui bottom attached tab segment borderless"
+        data-tab="checks"
+        role="tabpanel"
+        aria-labelledby="checks-tab"
+        id="checks-tab-panel"
+        hidden="hidden"
+      >
+        {% include "invenio_checks/requests/details.html" %}
+      </div>
+    {% endif %}
```
slint: ditto as above regarding `checks`

invenio_app_rdm/requests_ui/templates/semantic-ui/invenio_requests/record-deletion/index.html
yashlamba: Couldn't find a reference, but it's weird that we have just copied and pasted in all templates.

invenio_app_rdm/requests_ui/templates/semantic-ui/invenio_requests/record-deletion/index.html
yashlamba: I kept it for that reason only, have removed it.

PR #3141, opened 2025-07-29
UI record injection

ntarocco:


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCitationField.js
```diff
@@ -155,7 +156,7 @@ export class RecordCitationField extends Component {
 
 RecordCitationField.propTypes = {
   styles: PropTypes.array.isRequired,
-  record: PropTypes.object.isRequired,
+  recordLinks: PropTypes.object.isRequired,
```
slint, line 159: minor/comment: do we actually need all the links or wouldn't it be enough to just pass `recordLinks.self`? I know in CDS you also have the option for getting the citations for the parent DOI, so if that's the reason ignore this :)

ntarocco: `self` is enough. That was my first implementation, I was wondering if it was maybe too restrictive and "weird" to only inject the absolutely minimal things. My worry would be that in other case, we then start passing tens of specific params, which is good and bad at the same time.
I can change it to `self` if you prefer ☺️

PR #3082, opened 2025-06-20
config: add Celery beat schedule for updating collections size

slint:

PR #3147, opened 2025-08-05
fix(deposits): pass community_ui to the themed deposit form

max-moser:
The themed deposit form expected `community_ui` to be passed, but it was only passed as `community`

### DIFF
--- a/invenio_app_rdm/records_ui/views/deposits.py
+++ b/invenio_app_rdm/records_ui/views/deposits.py
@@ -15,6 +15,9 @@ from copy import deepcopy
 
 from flask import current_app, g, redirect
 from flask_login import login_required
+from invenio_communities.communities.resources.serializer import (
+    UICommunityJSONSerializer,
+)
 from invenio_communities.errors import CommunityDeletedError
 from invenio_communities.proxies import current_communities
 from invenio_communities.views.communities import render_community_theme_template
@@ -446,7 +449,7 @@ def new_record():
 @login_required
 @no_cache_response
 @pass_draft_community
-def deposit_create(community=None):
+def deposit_create(community=None, community_ui=None):
     """Create a new deposit."""
     can_create = current_rdm_records.records_service.check_permission(
         g.identity, "create"
@@ -479,6 +482,7 @@ def deposit_create(community=None):
         searchbar_config=dict(searchUrl=get_search_url()),
         record=new_record(),
         community=community,
+        community_ui=community_ui,
         community_use_jinja_header=community_use_jinja_header,
         files=dict(default_preview=None, entries=[], links={}),
         preselectedCommunity=community,
@@ -531,6 +535,7 @@ def deposit_edit(pid_value, draft=None, draft_files=None, files_locked=True):
                 id_=community["id"], identity=g.identity
             )
             community_theme = community.to_dict().get("theme", {})
+            community_ui = UICommunityJSONSerializer().dump_obj(community.to_dict())
         except CommunityDeletedError:
             pass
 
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -479,6 +479,7 @@ def deposit_create(community=None):
         searchbar_config=dict(searchUrl=get_search_url()),
         record=new_record(),
         community=community,
+        community_ui=community,
```
kpsherva: this is a little bit confusing comparing with the line above, could you try to get community_ui representation and assign it here instead of the community object? I think it was done in other community related views (for example check requests_ui)

max-moser: the UI serialization already happens in the decorator: https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/views/decorators.py#L349-L367

kpsherva: ouch, sorry (you see the confusion). Can we do the opposite then and in community assign the actual community, not UI serialized one? @jrcastro2 already made the attempt to normalise these around, I think we missed this one

max-moser: if i'm not mistaken, the UI-serialized stuff is just a dictionary?
as such, i'll edit the decorator to give `community` as ~API object~ result item, and `community_ui` as UI-serialized?

slint: FYI, I applied a similar fix in https://github.com/inveniosoftware/invenio-app-rdm/pull/3082 and missed the deposit view I guess 🫥

These views were using [the `@pass_community` decorator](https://github.com/inveniosoftware/invenio-communities/blob/72cbf688f8025c50cab3933b85c046c242af6528/invenio_communities/views/decorators.py#L22) which sets to `community` the service result and `community_ui` the UI serialization, so we should be consistent here as well.

max-moser: if i understand correctly, this is what i've added now.
or do you mean we should only use the `@pass_community` decorator and toss out the `@pass_draft_community`?

kpsherva: I mean when we pass variables to jinja context, we should avoid the naming confusion
so when we pass 
```
community_ui=community
community=community
```
from the naming I would expect to have two different community objects in the template context, while I only get the UI one.

so either we should remove `community=community` entirely (potentially with some consequences) or do
```
community_ui=community_ui # ui serialized object
community=community # community service result item object
```
pinging @jrcastro2 for reference on the same changes on record and record_ui

max-moser: to be clear, i *have* updated this PR to have `community=<result-item>`, and `community_ui=<ui-serialized-item>`
if i happened to miss one, could you please point it out in the lines?

jrcastro2: Recently we agreed to not mix the object with the UI dict, in the past we were calling it record and it could be the record_ui, which is confusing because sometimes you need the actual object and not the dict. So we fixed it in for the record in this PR: https://github.com/inveniosoftware/invenio-app-rdm/commit/3b8cbc4a06db2a32dcd5e30309cc251a8613cc5b#diff-2fb602eed821cbd7d07f7c61560b3e8ed50e165259da085c113fff47274b8496R236-R237

Now with the latest changes it looks good to me 👍

kpsherva (approved):

invenio_app_rdm/records_ui/views/deposits.py
slint: > if i understand correctly, this is what i've added now. or do you mean we should only use the `@pass_community` decorator and toss out the `@pass_draft_community`?

No, it's more about what type of object goes in `community` and `community_ui`.  The `@pass_draft_community` is needed since it gets the community from the querystring.

PR #3148, opened 2025-08-05
Provide migration to confirm affiliations entries to new shape

fenekku:
Needed for upgrade to v13. I will backport when merged.

I also then used `pipenv run invenio vocabularies update --vocabulary affiliations --origin ror-http` instead of `import`. I will make a separate PR for that in docs tomorrow.

### DIFF
--- a/invenio_app_rdm/upgrade_scripts/migrate_12_0_to_13_0.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_12_0_to_13_0.py
@@ -2,14 +2,15 @@
 #
 # Copyright (C) 2023-2024 CERN.
 # Copyright (C) 2024-2025 Graz University of Technology.
+# Copyright (C) 2025 Northwestern University.
 #
 # Invenio-App-RDM is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


slint:
I have a feeling that this change (i.e. moving PIDs to the `pid` column for affiliations, funders, and names) was actually introduced in v12 already, via [`invenio-vocabularies v2.0.0`](https://github.com/inveniosoftware/invenio-vocabularies/commit/40e6d91ed2ccda72429a638c888636d4ac76d0b9).

max-moser (approved):
LGTM!
i just tried out the affiliations part, and it certainly has not hurt the vocabularies:
```ipython
In [8]: from invenio_records_resources.proxies import current_service_registry as reg

In [9]: as_ = reg.get('affiliations')

In [10]: affs = [as_.record_cls(a.data, model=a) for a in as_.record_cls.model_cls.query.all()]

In [11]: from invenio_access.permissions import system_identity

In [12]: for a in affs: as_.read(system_identity, a.pid.pid_value).to_dict()
```

for the names, i've even had to run some more `data_for_name.pop(...)` operations because i've had wonky entries in my DB due to some nonsense i did; but the general approach also helped me.

PR #3149, opened 2025-08-06
ci: switch to `uv` for Python tests

slint:


slint:
> Ah, I guess the same approach could be taken to gradually introduce coverage to Python 3.13? I was fearing adding that python version to the main reusable workflows and breaking a bunch of CIs all in one go. This testing of waters seem like the better to go indeed.

Agree for testing 3.13 this way, though we can anyways override [the `python-version` input](https://github.com/inveniosoftware/workflows/blob/master/.github/workflows/tests-python.yml#L10) in the module's workflow config. Starting with the more complex/exotic modules also gives more confidence that this will not just break everything upstream.

slint:
Closed since this passed tests, which means we could enable this globally in the workflows repo :)

PR #3153, opened 2025-08-08
community value fixes

slint:
- **fixtures: allow specifying `template_name` in page fixtures**
  

- **pages: add community base template as a valid option**
  

- **fix(views): use correct community value**
  * After the change in 33f1b0d2, values passed to the UI or used for
    accessing UI-related attributes (e.g. the "theme"), must come from
    the community UI-serialized value instead of the service result item.


slint:
Yes, especially the views fixes some broken deposit flows via communities.

slint:
Backport PR at https://github.com/inveniosoftware/invenio-app-rdm/pull/3155

PR #3162, opened 2025-08-19
config: compat with new VCS integration

palkerecsenyi:
Closes https://github.com/inveniosoftware/invenio-vcs/issues/2

---

* We are updating `invenio-github` to be a generic package supporting many VCS providers, not just GitHub. As such, it will be renamed (currently to `invenio-vcs`).

* The `RDMGitHubRelease` class has been renamed and restructured, so this must be reflected in `config.py` for continued functionality. No other changes are needed to this repository at the moment.

* See https://github.com/inveniosoftware/invenio-rdm-records/pull/2284 and https://github.com/inveniosoftware/invenio-rdm-records/pull/2283

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2019-2025 CERN.
+# Copyright (C) 2019-2026 CERN.
 # Copyright (C) 2019-2020 Northwestern University.
 # Copyright (C) 2021-2025 Graz University of Technology.
 # Copyright (C) 2022-2025 KTH Royal Institute of Technology.
@@ -80,7 +80,6 @@ from invenio_rdm_records.services.errors import (
     InvalidAccessRestrictions,
     InvalidCommunityVisibility,
 )
-from invenio_rdm_records.services.github.release import RDMGithubRelease
 from invenio_rdm_records.services.permissions import RDMRequestsPermissionPolicy
 from invenio_rdm_records.services.stats import permissions_policy_lookup_factory
 from invenio_rdm_records.services.tasks import StatsRDMReindexTask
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/config.py
```diff
@@ -1435,6 +1436,9 @@ def github_link_render(record):
     # Record deletion
     RecordDeletionAcceptNotificationBuilder.type: RecordDeletionAcceptNotificationBuilder,
     RecordDeletionDeclineNotificationBuilder.type: RecordDeletionDeclineNotificationBuilder,
+    # VCS repository releases
```
zzacharo: These ones will need to be either config based added here or probably in the instance level....

palkerecsenyi: Yes agreed, I added them here temporarily for testing.

CHANGES.rst
```diff
@@ -2,27 +2,23 @@
...
+    Copyright (C) 2021-2025 Graz University of Technology.
 
     Invenio App RDM is free software; you can redistribute it and/or modify
     it under the terms of the MIT License; see LICENSE file for more details.
 
 Changes
 =======
 
-Version v14.0.0b5.dev0 (released 2026-02-03)
+Version v14.0.0b4.dev7 (released 2026-02-05)
```
carlinmack: nit: I feel like the release commit should be the last one, rather than before, as these changes aren't in this PR

utnapischtim: yes, it looks strange that on master we have `v14.0.0b5.dev0` already and afterwards it comes `v14.0.0b4.dev7`

Samk13: I’ve pushed an empty tag `v14.0.0b5.dev1` on upstream already, and released `v14.0.0b5.dev2`, so you can change this one to `v14.0.0b5.dev3`

PR #3171, opened 2025-08-25
feat: request/immediate record deletion

yashlamba:


invenio_app_rdm/administration/moderation/requests.py
```diff
@@ -0,0 +1,161 @@
...
+
+    display_search = True
+    display_delete = False
+    display_create = False
+    display_edit = False
+    display_read = False
+
+    search_config_name = "REQUESTS_MODERATION_REQUEST_SEARCH"
+    search_facets_config_name = "REQUESTS_MODERATION_REQUEST_FACETS"
+    search_sort_config_name = "REQUESTS_MODERATION_REQUEST_SORT_OPTIONS"
```
slint: We shouldn't be relying on config for these (for now at least). They should be defined here, since this admin view is pretty specialized already.

invenio_app_rdm/administration/moderation/requests.py
```diff
@@ -0,0 +1,161 @@
...
+    @classmethod
+    def get_service_schema(cls):
+        """Get marshmallow schema of the assigned service."""
+        request_type = cls.resource.service.request_type_registry.lookup(
+            "record-deletion"
+        )
+        # handle dynamic schema class declaration for requests
+        schema_cls = request_type.marshmallow_schema()
+        schema = schema_cls()
+        return schema
```
slint, line 126: This probably can be removed since we're not auto-generating the details view page.

invenio_app_rdm/administration/moderation/requests.py
```diff
@@ -0,0 +1,161 @@
...
+    extension_name = "invenio-requests"
+    api_endpoint = "/requests"
+    name = "request_details"
+    resource_config = "requests_resource"
+    title = _("Request details")
+    display_delete = False
+    display_edit = False
+
+    pid_path = "id"
+    request_headers = {"Accept": "application/json"}
```
slint: Can also be removed, since we're rendering Jinja templates with the request content

utnapischtim:
Since this includes a new UI feature, would it be possible to add screenshots to the PR? this would help to collect new features for the next major release of app-rdm.

invenio_app_rdm/administration/moderation/requests.py
yashlamba: renamed for the config to be in app rdm

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordDeletion/DeletionModal.js
```diff
@@ -0,0 +1,356 @@
...
+        return c;
+      }
+    });
+    this.setState({ checkboxState: nextCheckboxState });
+  };
+
+  deletionRequestSchema = (immediateDeletionAllowed) => {
+    if (immediateDeletionAllowed) {
+      return Yup.object({
+        reason: Yup.string().required("Required"),
```
Samk13, line 93:
```suggestion
        reason: Yup.string().required(i18next.t("Required")),
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordDeletion/DeletionModal.js
```diff
@@ -0,0 +1,356 @@
...
+      });
+    } else {
+      return Yup.object({
+        reason: Yup.string().required("Required"),
+        comment: Yup.string()
+          .min(25, "Please write at least 25 characters")
+          .required("Required"),
```
Samk13, line 100:
```js
reason: Yup.string()..required(i18next.t("Required")),
comment: Yup.string()
    .min(25, i18next.t("Please write at least 25 characters"))
    .required(i18next.t("Required")),
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordDeletion/DeletionModal.js
```diff
@@ -0,0 +1,356 @@
...
+                            state={checklistState}
+                            onStateChange={this.handleRadioUpdate}
+                          />
+                        ))}
+                      </TableBody>
+                    </Table>
+                    {messages.length > 0 && (
+                      <Message info>
+                        {messages.length === 1 ? (
+                          <p dangerouslySetInnerHTML={{ __html: messages[0] }} />
```
Samk13, line 280: Would it be possible to avoid using `dangerouslySetInnerHTML`?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/requests/RequestsSearchResultItem.js
```diff
@@ -0,0 +1,81 @@
...
+      <Table.Row>
+        <Table.Cell
+          key={`${columns[0][1]["text"]}-${columns[0][1]["order"]}`}
+          data-label={columns[0][1]["text"]}
+          className="word-break-all"
+        >
+          <a href={AdminUIRoutes.detailsView(listUIEndpoint, result, idKeyPath)}>
+            {/* TODO we need a better way to get the (translatable) label of a request type
+            https://github.com/inveniosoftware/invenio-requests/issues/414 */}
+            {result.type === "record-deletion" && "Record deletion"}
```
Samk13, line 35:
```suggestion
            {result.type === "record-deletion" && i18next.t("Record deletion")}
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordDeletion/DeletionModal.js
```diff
@@ -0,0 +1,356 @@
...
+      checklistState.some((x) => x === true || x === undefined);
+
+    return (
+      <Modal
+        open={open}
+        closeIcon
+        onClose={this.handleClose}
+        role="dialog"
+        aria-modal="true"
+        tab-index="-1"
```
Samk13, line 171:
```suggestion
        tabIndex="-1"
```
see: https://react.dev/reference/react-dom/components#all-html-components

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordDeletion/DeletionModal.js
carlinmack: It's because the message can contain links, and these messages are defined in config so aren't user generated content

Samk13: That makes sense, thanks for clarifying. Maybe something to consider long-term: instead of dangerouslySetInnerHTML, we could render the config messages as React elements (e.g., only parsing <a> tags). That would still allow links while avoiding raw HTML injection. Not urgent, just an idea for future cleanup.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordDeletion/DeletionModal.js
carlinmack: also ariaModal?

Samk13: Found this In React docs: 
all DOM properties and attributes (including event handlers) should be camelCased. For example, the HTML attribute tabindex corresponds to the attribute tabIndex in React. The exception is aria-* and data-* attributes, which should be lowercased. For example, you can keep aria-label as aria-label.
see: 
https://legacy.reactjs.org/docs/dom-elements.html#:~:text=In%20React%2C%20all,aria%2Dlabel.

carlinmack:
> Should be discussed in IRL regarding the duplicated code of the comments in the UI of admin panel.

We can re-use the same template, but we have a choice in how to fix it, either:

1. Rename the main content block in the admin panel so that it is the same as the one in the non-admin base template
      * We would then need to rename this block in all admin templates
2. We add the main content block either wrapping, beside or inside the admin main content block
      * No changing of templates needed but we need to decide where the block should go and be happy about having two blocks where you can define the main content

PR #3175, opened 2025-08-28
deposit: pass to the form config the published record if we edit one

zzacharo:


invenio_app_rdm/records_ui/views/deposits.py
```diff
@@ -521,6 +521,12 @@ def deposit_edit(pid_value, draft=None, draft_files=None, files_locked=True):
     ui_serializer = UIJSONSerializer()
     record = ui_serializer.dump_obj(draft.to_dict())
 
+    published_record = {}
```
kpsherva:
```suggestion
    published_record = None
```
not sure how this would play out but maybe safer if we pass something unmutable

Issue #3137, opened 2025-07-25
ui: No formatting for large numbers in CompactStats


palkerecsenyi:
When showing large numbers in the [`CompactStats`](https://github.com/inveniosoftware/invenio-app-rdm/blob/be47314058c1143cdb488bb9c699eb2aee97202e/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/CompactStats.js#L12) no formatting is applied which means large numbers do not look very readable currently:

<img width="180" height="58" alt="Image" src="https://github.com/user-attachments/assets/2412e69a-e8c6-4c00-b38f-1fde520707df" />

We could use [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) to implement an internationalised formatter for the numbers quite easily (e.g. "1,234" in en-GB but "1.234" in many other locales)

Sandy-07-coder:
Hi @palkerecsenyi! 👋

I’d love to work on this issue — this would be my first good first issue contribution. I’ve read about `Intl.NumberFormat` on MDN and also looked into the `compact` number formatting style, so I feel ready to try implementing this.

Could you please assign this to me if it’s still available?

Thanks a lot for your help! 🚀

palkerecsenyi:
Hi @Sandy-07-coder! Thanks for the comment and for contributing to InvenioRDM! I've assigned the issue to you.

Please pay attention to the [contribution guide](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/CONTRIBUTING.rst) and the [copyright policy](https://inveniordm.docs.cern.ch/community/copyright-policy/). You might also find it useful to [run a local InvenioRDM instance](https://inveniordm.docs.cern.ch/install/) to test your changes, as well as to [install the `invenio-app-rdm` package](https://inveniordm.docs.cern.ch/community/code/package-development) into your local instance.

Please let me know if you would like any more information. Thanks again!

Sandy-07-coder:
Hi @palkerecsenyi  
### I made a pull request about this issue, verify it .
#### Could we connect on LinkedIn? my id : www.linkedin.com/in/santhosh2673

PR #3184, opened 2025-09-12
ui: format numbers in CompactStats

palkerecsenyi:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/3137

---

* Adds locale-based formatting for large numbers using the HTML API `Intl.NumberFormat`

* The locale is taken from `i18next.language`. The view count and the download count are both formatted.

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/utils.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/utils.js
@@ -119,3 +119,12 @@ export function SearchItemCreators({ creators, className, othersLink }) {
  */
 export const timestampToRelativeTime = (timestamp) =>
   DateTime.fromISO(timestamp).setLocale(i18next.language).toRelative();
+
+/**
+ * Returns a string-formatted number in the user's locale (e.g. 1,000 in en-GB but 1.000 in de-DE)
+ *
+ * @param {Number} number
+ * @returns string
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/utils.js
```diff
@@ -119,3 +119,12 @@ export function SearchItemCreators({ creators, className, othersLink }) {
...
 export const timestampToRelativeTime = (timestamp) =>
   DateTime.fromISO(timestamp).setLocale(i18next.language).toRelative();
+
+/**
+ * Returns a string-formatted number in the user's locale (e.g. 1,000 in en-GB but 1.000 in de-DE)
+ *
+ * @param {Number} number
+ * @returns string
+ */
+export const localizedFormatNumber = (number) =>
```
kpsherva, line 129: have you conidered moving it to react-invenio-forms? we will need this util in more places, maybe even makes sense to add to react-searchkit @zzacharo

palkerecsenyi: Do we have access to `i18next` in react-invenio-forms?

kpsherva: unfortunately no, we don't

PR #3186, opened 2025-09-17
upgrade_scripts: Add v13 to v14 migration script

sakshamarora1:
related: https://github.com/inveniosoftware/invenio-rdm-records/pull/2154
closes: https://github.com/CERNDocumentServer/cds-rdm/issues/522

### DIFF
--- a/invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
@@ -1,0 +1,173 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2025 CERN.
+#
+# Invenio-App-RDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Record migration script from InvenioRDM 13.0 to 14.0.
+
+Disclaimer: This script is intended to be executed *only once*, namely when
+upgrading from InvenioRDM 13.0 to 14.0!
+If this script is executed at any other time, probably the best case scenario
+is that nothing happens!
+
+
+This script has been tested with following data:
+
+- record
+    - internal_notes
+"""
+
+from click import secho
+from invenio_access.permissions import system_identity
+from invenio_rdm_records.proxies import current_rdm_records_service as records_service
+from invenio_search.engine import dsl
+
+
+def run_upgrade(has, migrate_record, migrate_draft):
+    """Run upgrade."""
+    # Handle published records
+    published_records = records_service.search(
+        system_identity,
+        params={"allversions": True, "include_deleted": True},
+        extra_filter=has,
+    )
+    for result in published_records.hits:
+        try:
+            migrate_record(result)
+        except Exception as error:
+            secho(f"> Error {repr(error)}", fg="red")
+            error = f"Record {result['id']} failed to update"
+
+    # Handle draft records
+    draft_records = records_service.search_drafts(
+        system_identity,
+        params={"allversions": True},
+        extra_filter=has,
+    )
+    for result in draft_records.hits:
+        try:
+            migrate_draft(result)
+        except Exception as error:
+            secho(f"> Error {repr(error)}", fg="red")
+            error = f"Draft {result['id']} failed to update"
+
+
+def run_update_for_resource_type():
+    """Run update for resource type."""
+
+    def migrate_resource_type_in_record(hit_result):
+        """
+        Update resource type from event to publication-dissertation.
+        We go through the service layer to automatically trigger the DOI update and re-indexing.
+        """
+        print(f"Updating resource type for {hit_result['id']}")
+        record = records_service.read_latest(system_identity, hit_result["id"])
+        if record.data["metadata"]["resource_type"]["id"] != "event":
+            secho(
+                f"Skipping record <{record.id}> because it doesn't have resource-type 'event'!",
+                fg="yellow",
+            )
+            return
+
+        draft = records_service.edit(system_identity, record.id)
+        if draft.data["metadata"] != record.data["metadata"]:
+            secho(
+                f"Record <{record.id}> has an existing draft <{draft.id}> with different metadata! Updating record with record's metadata and re-creating draft with new metadata.",
+                fg="yellow",
+            )
+            # Store the existing draft metadata
+            new_draft_metadata = draft.data["metadata"]
+            # Delete the existing draft
+            is_draft_deleted = records_service.delete_draft(system_identity, draft.id)
+            if not is_draft_deleted:
+                secho(f"Failed to delete draft <{draft.id}>! Skipping...", fg="red")
+                return
+            # Create a temporary draft with the record's metadata and update the resource type
+            temp_draft = records_service.edit(system_identity, record.id)
+            temp_draft.data["metadata"]["resource_type"][
+                "id"
+            ] = "publication-dissertation"
+            updated_draft = records_service.update_draft(
+                system_identity, temp_draft.id, temp_draft.data
+            )
+            updated_record = records_service.publish(system_identity, updated_draft.id)
+
+            # Create a new draft with the existing draft's metadata and update the resource type again
+            new_draft = records_service.edit(system_identity, record.id)
+            new_draft.data["metadata"] = new_draft_metadata
+            new_draft.data["metadata"]["resource_type"][
+                "id"
+            ] = "publication-dissertation"
+            updated_draft = records_service.update_draft(
+                system_identity, new_draft.id, new_draft.data
+            )
+        else:
+            # If the draft has the same metadata as the record, just update the resource type
+            draft.data["metadata"]["resource_type"]["id"] = "publication-dissertation"
+            updated_draft = records_service.update_draft(
+                system_identity, draft.id, draft.data
+            )
+            updated_record = records_service.publish(system_identity, updated_draft.id)
+        secho(
+            f"Record <{updated_record.id}> has been updated successfully.", fg="green"
+        )
+
+    def migrate_resource_type_in_draft(hit_result):
+        """
+        Update resource type from event to publication-dissertation.
+        We go through the service layer to automatically trigger the DOI update and re-indexing.
+        """
+        print(f"Updating resource type for {hit_result['id']}")
+        draft = records_service.edit(system_identity, hit_result["id"])
+        if draft.data["metadata"]["resource_type"]["id"] != "event":
+            secho(
+                f"Skipping draft <{draft.id}> because it doesn't have resource-type 'event'!",
+                fg="yellow",
+            )
+            return
+
+        draft.data["metadata"]["resource_type"]["id"] = "publication-dissertation"
+        updated_draft = records_service.update_draft(
+            system_identity, draft.id, draft.data
+        )
+        secho(f"Draft <{updated_draft.id}> has been updated successfully.", fg="green")
+
+    # Common query filter
+    has_resource_type = dsl.Q("query_string", query="metadata.resource_type.id:event")
+
+    secho("Resource type update has started.", fg="green")
+
+    run_upgrade(
+        has_resource_type,
+        migrate_resource_type_in_record,
+        migrate_resource_type_in_draft,
+    )
+
+    secho("Resource type update has finished.", fg="green")
+
+
+def execute_upgrade():
+    """Execute the upgrade from InvenioRDM 13.0 to 14.0.
+
+    Please read the disclaimer on this module before thinking about executing
+    this function!
+    THIS MODULE IS WORK IN PROGRESS, UNTIL official v14 release
+
+    NOTE:
+    since the data upgrade steps are more selective now, the approach how to do
+    it has been changed. now the records/drafts which should be updated are
+    searched by a filter and then the updates are applied to those
+    records/drafts explicitly. this should improve speed and should make it
+    easier to upgrade large instances
+
+    """
+    secho("Starting data migration...", fg="green")
+
+    run_update_for_resource_type()
+
+
+# if the script is executed on its own, perform the upgrade
+if __name__ == "__main__":
+    execute_upgrade()


invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,173 @@
...
+from click import secho
+from invenio_access.permissions import system_identity
+from invenio_rdm_records.proxies import current_rdm_records_service as records_service
+from invenio_search.engine import dsl
+
+
+def run_upgrade(has, migrate_record, migrate_draft):
+    """Run upgrade."""
+    # Handle published records
+    published_records = records_service.search(
```
kpsherva: IMO you should use scan here

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,173 @@
...
+        system_identity,
+        params={"allversions": True, "include_deleted": True},
+        extra_filter=has,
+    )
+    for result in published_records.hits:
```
kpsherva, line 47: how many hits will you get here? will it be only 15?

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,173 @@
...
+            secho(f"> Error {repr(error)}", fg="red")
+            error = f"Record {result['id']} failed to update"
+
+    # Handle draft records
+    draft_records = records_service.search_drafts(
+        system_identity,
+        params={"allversions": True},
+        extra_filter=has,
+    )
+    for result in draft_records.hits:
```
kpsherva: same here

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,173 @@
...
+        print(f"Updating resource type for {hit_result['id']}")
+        record = records_service.read_latest(system_identity, hit_result["id"])
+        if record.data["metadata"]["resource_type"]["id"] != "event":
+            secho(
+                f"Skipping record <{record.id}> because it doesn't have resource-type 'event'!",
+                fg="yellow",
+            )
+            return
+
+        draft = records_service.edit(system_identity, record.id)
```
kpsherva: @ntarocco @zzacharo I have a bit of doubt if we shouldn't update on lower level than using the service, because we create another revision for the record for this change. Unless that is what we want to keep it in the history. WDYT?

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+
+    def migrate_resource_type_in_draft(hit_result):
+        """
+        Update resource type from event to publication-dissertation.
+        We go through the service layer to automatically trigger the DOI update and re-indexing.
+        """
+        print(f"Updating resource type for {hit_result['id']}")
+        draft = records_service.edit(system_identity, hit_result["id"])
+        if draft.data["metadata"]["resource_type"]["id"] != "event":
```
sakshamarora1: Oops, should be `publication-thesis`

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,173 @@
...
+            draft.data["metadata"]["resource_type"]["id"] = "publication-dissertation"
+            updated_draft = records_service.update_draft(
+                system_identity, draft.id, draft.data
+            )
+            updated_record = records_service.publish(system_identity, updated_draft.id)
+        secho(
+            f"Record <{updated_record.id}> has been updated successfully.", fg="green"
+        )
+
+    def migrate_resource_type_in_draft(hit_result):
```
kpsherva, line 132: does it work also for new version draft?

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
ntarocco: Yeah, you need to fetch all of them

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,173 @@
...
+    """Run update for resource type."""
+
+    def migrate_resource_type_in_record(hit_result):
+        """
+        Update resource type from event to publication-dissertation.
+        We go through the service layer to automatically trigger the DOI update and re-indexing.
+        """
+        print(f"Updating resource type for {hit_result['id']}")
+        record = records_service.read_latest(system_identity, hit_result["id"])
+        if record.data["metadata"]["resource_type"]["id"] != "event":
```
ntarocco: Why `event`?

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
ntarocco: I am ok to create a new revision, why not? This also ensure that a bunch of operations are done consistently on publish, e.g. DOI update

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,173 @@
...
+            secho(
+                f"Skipping record <{record.id}> because it doesn't have resource-type 'event'!",
+                fg="yellow",
+            )
+            return
+
+        draft = records_service.edit(system_identity, record.id)
+        if draft.data["metadata"] != record.data["metadata"]:
```
ntarocco: I am not very sure that this comparison will work fine. We should instead implement the logic that:
1. if published, but no draft, then edit -> publish
2. if published, with a draft, can we update and save both, without triggering a version conflict? (to be fully tested)
3. if only draft, update

ntarocco (commented):
All the `event` should be fixed/replaced. The logic of what to change for records and drafts should be reviewed too.

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
sakshamarora1: Thanks! I will update the logic to work for if the draft exists instead.

> can we update and save both, without triggering a version conflict?

Do you mean via the service layer? I couldn't find another way to directly update the record.

https://github.com/inveniosoftware/invenio-drafts-resources/blob/07888445c5212ad969d82c7cebfffe405cecc48a/invenio_drafts_resources/services/records/service.py#L110-L112

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
sakshamarora1: Yes, as it's a normal draft

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
sakshamarora1: forgot to change after testing, fixed now!

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
slint: In general, going through the service is the best approach here (so that all the components, etc. run as well, like e.g. DataCite DOI metadata updates). Also, see the [data model](https://inveniordm.docs.cern.ch/maintenance/architecture/records/#data-model) for the overal lifecycle.

For 2. though, it's going to be rough... There's no easy way to make the update without triggering a revision bump, so the best approach is probably to make the change, and then update the `RDMDraftMetadata.fork_version_id` with the new value from `RDMRecordMetadata.version_id` (cc @utnapischtim I think we should have been doing this in the previous migrations as well 😬).

Because this is an offline migration, doing it this way is a bit "safer" (since there are fewer moving parts). But still, making changes to the record without going through the draft is a general recipe for disaster...

---

An alternative would be to capture the "changeset" of the current draft, discard the draft, go through `1.` and then edit again and "apply" the captured changeset. But this sounds complicated.

ntarocco: Indeed, the latter was shared by @sakshamarora1 too, but I think that it can be tricky, in particular if a change in the DOI was involved in the draft. Basically, when it is not only about metadata.
I am not sure what's best here. Messing with the `version_id` should work if done and tested correctly.

sakshamarora1: I have added the line `draft._record.fork_revision_id = record._record.revision_id`, and tested it in my local instance, the changes are propagating correctly!

ntarocco: > I have added the line `draft._record.fork_revision_id = record._record.revision_id`, and tested it in my local instance, the changes are propagating correctly!

Did you manage to test that you can successfully publish the draft after the change to the revisions?

sakshamarora1: Yes, it works.

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,179 @@
...
+from invenio_rdm_records.proxies import current_rdm_records_service as records_service
+from invenio_search.engine import dsl
+
+
+def run_upgrade(has, migrate_record, migrate_draft):
+    """Run upgrade."""
+    # Handle published records
+    published_records = records_service.scan(
+        system_identity,
+        params={"allversions": True, "include_deleted": True},
```
ntarocco: I am not 100% sure that the service layer edit/publish will work for deleted records.

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+    for result in published_records.hits:
+        try:
+            migrate_record(result)
+        except Exception as error:
+            secho(f"> Error {repr(error)}", fg="red")
+            error = f"Record {result['id']} failed to update"
+
+    # Handle draft records
+    draft_records = records_service._search(
```
ntarocco, line 55: using `search` here will have the same issue: you only get the first 10/15 results.

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,179 @@
...
+            record._record.commit()
+            db.session.commit()
+            records_service.indexer.index(record._record)
+            # Update DOI metadata if record has DOI
+            if record.pids.get("doi", None):
+                records_service.pids.register_or_update(
+                    system_identity, record.id, "doi", parent=False
+                )
+            # Step 2: Update the resource type in the draft
+            draft = records_service.edit(system_identity, draft.id)
```
ntarocco: you already have the `draft` from above

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,179 @@
...
+            draft.data["metadata"]["resource_type"]["id"] = "publication-dissertation"
+            draft._record.fork_version_id = record._record.revision_id
```
ntarocco, line 116: can you add a comment here to explain?

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,179 @@
...
+Disclaimer: This script is intended to be executed *only once*, namely when
+upgrading from InvenioRDM 13.0 to 14.0!
+If this script is executed at any other time, probably the best case scenario
+is that nothing happens!
+
+
+This script has been tested with following data:
+
+- record
+    - internal_notes
```
ntarocco: can you update the docstring here?

ntarocco (commented):
Thanks a lot! It would be really really really good to add tests for this procedure, testing the 3 cases of records/drafts and making sure DOI is updated and draft can be published after messing with the revisions.

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
sakshamarora1: This is the internal `_search()` method, to get the search object, line 54 has `.scan()`

tests/test_migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,309 @@
...
+
+        # Verify all thesis records were migrated
+        thesis_records = records_service.search(
+            system_identity,
+            params={"allversions": True},
+            extra_filter=dsl.Q(
+                "term", **{"metadata.resource_type.id": "publication-thesis"}
+            ),
+        )
+        assert thesis_records.total == 0
```
sakshamarora1, line 248: When a draft is published as record, the draft is soft deleted right? But it comes up in the search here. Is that expected? The test is failing here.

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
ntarocco: This question was not answered.

tests/test_migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,308 @@
...
```
ntarocco, line 261: Tests not passing?

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
sakshamarora1: It will work, but now I've removed it as we don't want to re-publish a deleted record.

tests/test_migrate_13_0_to_14_0.py
sakshamarora1: Added a delay for the indexing. Now they are passing.

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,187 @@
...
+    # Handle published records
+    published_records = records_service.scan(
+        identity=system_identity,
+        params={"allversions": True},
+        extra_filter=has,
+    )
+    for result in published_records.hits:
+        try:
+            migrate_record(result)
+            time.sleep(0.5)  # Small delay to allow index refresh
```
kpsherva: please replace with refresh_index in the test (not in the script)

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,185 @@
...
+from click import secho
+from invenio_access.permissions import system_identity
+from invenio_db import db
+from invenio_drafts_resources.resources.records.errors import DraftNotCreatedError
+from invenio_rdm_records.proxies import current_rdm_records_service as records_service
+from invenio_search.engine import dsl
+
+
+def run_upgrade(has, migrate_record, migrate_draft):
+    """Run upgrade."""
```
tmorrell:
```suggestion
    """Run upgrade on selected records and drafts."""
```

invenio_app_rdm/upgrade_scripts/migrate_13_0_to_14_0.py
```diff
@@ -0,0 +1,185 @@
...
+            )
+            return
+
+        draft.data["metadata"]["resource_type"]["id"] = "publication-dissertation"
+        updated_draft = records_service.update_draft(
+            system_identity, draft.id, draft.data
+        )
+        secho(f"Draft <{updated_draft.id}> has been updated successfully.", fg="green")
+
+    # Common query filter
```
tmorrell:
```suggestion
    # Query records/drafts with resource type publication-thesis
```

tmorrell (approved):
Looks good. I just added some small comments

PR #3198, opened 2025-10-10
feat(form): rename creators to authors and help text

### LINKED ISSUE #3197: Creator/Contributor: make the difference clear
It has been reported by multiple users that the difference between Creator and Contributor is not clear.
Moreover, some users confuse "Creator" with who is creating the record, instead of the publication's author, and as a consequence, use Contributor for the authors of the publication.

While the difference between author and contributor is explained [here](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/creator/#creator) and [here](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/contributor/), I think that the most actionable explanation is to clarify that contributors are not added to the citation string in InvenioRDM.

1. Change the Creator label to Author
2. Add a help text to the Contributor to clarify the citation part, so it is easy for users to decide how/where to add authors.

ntarocco:
* rename the Creators field to Authors
* add help text under Contributors to explain the difference
* closes #3197

**Requires**: https://github.com/inveniosoftware/invenio-rdm-records/pull/2176

Screenshot 2025-10-10 at 10 08 29

Screenshot 2025-10-10 at 10 08 50

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
@@ -262,13 +262,13 @@
             {# Title #}
             {%- block record_title -%}
               <section id="record-title-section"
-                       aria-label="{{ _('Record title and creators') }}">
+                       aria-label="{{ _('Record title, authors and contributors') }}">
                 <h1 id="record-title"
                     class="wrap-overflowing-text">{{ metadata.title }}</h1>
 
                 {% if record_ui["ui"]["creators"] or record_ui["ui"]["contributors"] %}
                   <section id="creatibutors"
-                           aria-label="{{ _('Creators and contributors') }}">
+                           aria-label="{{ _('Authors and contributors') }}">
                     {%- include "invenio_app_rdm/records/details/creatibutors.html" %}
                   </section>
                 {% endif %}
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
@@ -303,8 +303,11 @@ export class RDMDepositForm extends Component {
                       fieldPath="metadata.creators"
                     >
                       <CreatibutorsField
-                        label={i18next.t("Creators")}
+                        label={i18next.t("Authors/Creators")}
                         labelIcon="user"
+                        addButtonHelpText={i18next.t(
+                          "Use the Authors/Creators field for names that should appear in the citation. Use the Contributors field below for other names."
+                        )}
                         fieldPath="metadata.creators"
                         roleOptions={this.vocabularies.metadata.creators.role}
                         schema="creators"
@@ -423,6 +426,9 @@ export class RDMDepositForm extends Component {
                     >
                       <CreatibutorsField
                         addButtonLabel={i18next.t("Add contributor")}
+                        addButtonHelpText={i18next.t(
+                          "Contributors are not included in the citation, but are shown on the record page."
+                        )}
                         label={i18next.t("Contributors")}
                         labelIcon="user plus"
                         fieldPath="metadata.contributors"
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/creatibutors.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/creatibutors.html
@@ -13,7 +13,7 @@
   {% if record_ui["ui"]["creators"] and record_ui["ui"]["creators"]["creators"] %}
     <div class="row ui accordion affiliations">
       <div class="sixteen wide mobile twelve wide tablet thirteen wide computer column">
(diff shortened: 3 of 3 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
```diff
@@ -268,7 +268,7 @@
 
                 {% if record_ui["ui"]["creators"] or record_ui["ui"]["contributors"] %}
                   <section id="creatibutors"
-                           aria-label="{{ _('Creators and contributors') }}">
+                           aria-label="{{ _('Authors and contributors') }}">
                     {%- include "invenio_app_rdm/records/details/creatibutors.html" %}
```
palkerecsenyi, line 272: In this file we maybe also need to change the screenreader-only heading:

https://github.com/inveniosoftware/invenio-app-rdm/blob/894676a39fb804922fd7ef18aa9f9f9d4c5a8ec7/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/creatibutors.html#L16

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html
palkerecsenyi:
```suggestion
                       aria-label="{{ _('Record title, authors, and contributors') }}">
```

Or would that be too clunky?

palkerecsenyi (commented):
Just some small things, otherwise looks good from a technical perspective :)

ntarocco:
New version adding both `Authors` and `Creators`. However, I did not want to change the btn and add both, or simply put `Add`, looks ugly. I kept it `Add authors`.
Added extra help text on both:

Screenshot 2025-10-24 at 18 14 09

Screenshot 2025-10-24 at 18 14 19

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -303,8 +303,11 @@ export class RDMDepositForm extends Component {
                       fieldPath="metadata.creators"
                     >
                       <CreatibutorsField
-                        label={i18next.t("Creators")}
+                        label={i18next.t("Authors/Creators")}
                         labelIcon="user"
+                        addButtonHelpText={i18next.t(
+                          "Use the Authors/Creators field for names that should appear in citations. Use Contributors below for others."
```
tmorrell: I think citation should be singular, so something like 

```suggestion
                          "Use the Authors/Creators field for names that should appear in the citation. Use the Contributors field below for other names."
```

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -423,6 +426,9 @@ export class RDMDepositForm extends Component {
                     >
                       <CreatibutorsField
                         addButtonLabel={i18next.t("Add contributor")}
+                        addButtonHelpText={i18next.t(
+                          "Contributors (compared to authors) are not included in citations, but are shown on the record page."
```
tmorrell:
```suggestion
                          "Contributors are not included in the citation, but are shown on the record page."
```
I'm not sure the compared to authors bit is really necessary

mfenner: @tmorrell encouraged me to also say something controversial, not about the specific changes, but the broader topic of creators and contributors. Today I worked on better implementing contributor roles in my InvenioRDM instance based on the needs of the ropensi community blog (https://doi.org/10.59350/510pg-zzf58). As I said before, the terminology author/creator/contributor is messy and for example different between DataCite/Crossref/ORCID/DublinCore/Schema.org and the distinction based on who is listed in a citation is also not always clear. In addition, and more important for InvenioRDM, there is an increasing need to document broader contributions (in the case of ropensci blog for example editor and translator), but also for example reviewer or thesis supervisor plus the various roles in the CrediT taxonomy.

My suggestion is to deploy the proposed form changes (with the details TBD as currently discussed), and then look into bigger changes in the coming months - as a feature for the v14.0 release. Basically consolidating everything around the concept "contributor" with a list of possible roles for each contributor. For DOI registration with DataCite (and Crossref) this must be mapped to their schemata, but this was straightforward when I did this in the commonmeta library.

What we would gain is not only a simpler submission form, and better alignment with other metadata schemata but more flexibility of what contributions we want to include in InvenioRDM metadata.

mfenner: The various roles defined in the citation-style-language (and therefore appearing in at least some citations) are defined here: https://docs.citationstyles.org/en/stable/specification.html?highlight=translator#name-variables Some are not relevant for scholarly content, but many of them could be useful in InvenioRDM record metadata.

ntarocco: I agree, I propose to bring up the topic as a [GitHub Discussion](https://github.com/inveniosoftware/product-rdm/discussions) and present it in a telecon, to collect feedback from everyone.

PR #3199, opened 2025-10-10
profiler: Add blueprint and template

sakshamarora1:


invenio_app_rdm/profiler.py
```diff
@@ -0,0 +1,303 @@
...
+            self.refresh_active_session()
+            return response
+
+    def init_config(self, app):
+        """Initialize configuration."""
+        app.config.setdefault("PROFILER_STORAGE", Path(app.instance_path) / "profiler")
+        app.config.setdefault("PROFILER_ACTIVE_SESSION_LIFETIME", timedelta(minutes=60))
+        app.config.setdefault("PROFILER_ACTIVE_SESSION_REFRESH", timedelta(minutes=30))
+        app.config.setdefault("PROFILER_IGNORED_ENDPOINTS", ["static", r"profiler\..+"])
+        app.config.setdefault("PROFILER_PERMISSION", lambda: True)
```
ntarocco: Please restrict it to super admins only! I recommend comparing the config in Zenodo.

setup.cfg
```diff
@@ -71,6 +71,8 @@ install_requires =
     invenio-pages>=7.0.0,<8.0.0
     invenio-audit-logs>=0.3.0,<1.0.0
     invenio-sitemap>=0.1.0,<2.0.0
+    pyinstrument>=5.0.0,<6
```
ntarocco: Shall we make this optional instead? If yes, the imports in the profiler should take that into account.
I would propose to have a feature flag for this, so by default it is not actived.

ntarocco (commented):
@slint anything against in adding this to InvenioRDM?

invenio_app_rdm/profiler.py
```diff
@@ -0,0 +1,316 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# ZenodoRDM is free software; you can redistribute it and/or modify it
```
ptamarit: Change the header to InvenioRDM/Invenio-App-RDM

invenio_app_rdm/profiler.py
```diff
@@ -0,0 +1,316 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+"""Profiler module."""
+
+import re
+from datetime import datetime, timedelta
+from pathlib import Path
+
+import pyinstrument
+import sqlalchemy as sa
+import sqltap
```
ptamarit: Wrap the import for `pyinstrument` and `sqltap` in a try/except for these optional dependencies (OK, since we are not testing them)

ptamarit (approved):
This will need a rebase and some documentation.
(For the record, I never used the Zenodo profiler.)

sakshamarora1:
Closed in favour of #3423

PR #3200, opened 2025-10-10
fix: align selector with semantic-ui

carlinmack:
:heart: Thank you for your contribution!

### Description

> Please describe briefly your pull request.

For styling, semantic ui targets the title of an accordion via `.ui.accordion .title:not(.ui)` so that you can escape the styling by adding the ui class.  This is important as the selector is for any `.title`s that are children, not just direct children.

image

This is relevant for my upcoming pull requests where I want to use an accordion field inside a section which looks like 
image
but with this PR instead looks like this:
image

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [ ] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [ ] I've created [logical separate commits](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#commit-message).
- [ ] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/community/translations/i18n/).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/community/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/community/code/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/community/code/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/community/code/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

PR #3201, opened 2025-10-10
chore: extract evaluate_record_deletion into utils

carlinmack:

### Description

### Checklist

- [ ] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).

**Frontend**

**Reminder**

### DIFF
--- a/invenio_app_rdm/records_ui/utils.py
+++ b/invenio_app_rdm/records_ui/utils.py
@@ -13,7 +13,9 @@ from itertools import chain
 
 from flask import current_app
 from invenio_access.permissions import system_identity
+from invenio_rdm_records.records.api import RDMRecord
 from invenio_rdm_records.requests.record_deletion import RecordDeletion
+from invenio_rdm_records.services.config import RDMRecordDeletionPolicy
 from invenio_records.dictutils import dict_set
 from invenio_records.errors import MissingModelError
 from invenio_records_files.api import FileObject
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/records_ui/utils.py
```diff
@@ -109,3 +110,47 @@ def get_existing_deletion_request(record_id):
...
+            "allowed": rd_allowed,
+            "recordDeletion": rec_del,
+            "checklist": (
+                current_app.config["RDM_IMMEDIATE_RECORD_DELETION_CHECKLIST"]
+                if immediate.allowed
+                else current_app.config["RDM_REQUEST_RECORD_DELETION_CHECKLIST"]
+            ),
+            "context": {
+                "files": draft._record.files.count,
+                "internalDoi": draft._record.pids["doi"]["provider"] != "external",
```
slint:
```suggestion
                "files": record.files.count,
                "internalDoi": record.pids["doi"]["provider"] != "external",
```

1. We should be using the published record only in this function (since this is the one that has the "locked/published" files and PIDs)
2. We should not be using `._record`. See the other comment in the function signature.

invenio_app_rdm/records_ui/utils.py
```diff
@@ -109,3 +110,47 @@ def get_existing_deletion_request(record_id):
     )
     if existing_requests.total > 0:
         return list(existing_requests)[0]
+
+
+def evaluate_record_deletion(record, draft=None):
```
slint:
```suggestion
def evaluate_record_deletion(record: api.RDMRecord):
```

minor: I think this function could work in a way where we only pass the API `record`. I.e. there's nothing we need/use the draft for. We should adapt the callers to make sure that they pass the correct value.

(No need for the type-hint, I just did to make it clear that we don't expect a service result object, but rather an API record.)

carlinmack: ++ for the type hint as I have no idea the different python `type`s of records we have ahah

invenio_app_rdm/records_ui/utils.py
carlinmack: on line 118 we have `rec_del = RDMRecordDeletionPolicy().evaluate(g.identity, record._record)` 👀

carlinmack: image

🦭 😢

slint: - `RecordItem` is the service result, and doesn't have e.g. systemfields. It's the "projection" that we serialize when we want to pass this record back to an end-user (taking into account permissions).
- `ReocrdItem._record` is the `invenio_rdm_records.records.api.RDMRecord` instance, which has all the systemfields and low-level fields. This should be the one used for evaluating the record deletion since we need access to things ike `record.files.count` or `record.parent.communities`, etc.

Maybe it would be better for UI purposes to use the `RecordItem`, but then we would need to access the attributes as `record["files"]["count"]` and `record["pids"]["doi"]["provider"]` and have a separate function (or pass the `RecordItem` instead).

carlinmack: ~~tbh we have no need for internalDOI anymore.~~ I would be in favour of switching to `record["files"]["count"]` as it's more direct

invenio_app_rdm/records_ui/utils.py
```diff
@@ -109,3 +111,45 @@ def get_existing_deletion_request(record_id):
     )
     if existing_requests.total > 0:
         return list(existing_requests)[0]
+
+
+def evaluate_record_deletion(record: RDMRecord):
+    """Evaluate whether a given record can be deleted by an identity."""
+    rec_del = RDMRecordDeletionPolicy().evaluate(g.identity, record._record)
```
ntarocco: same here for `g.identity`

invenio_app_rdm/records_ui/utils.py
```diff
@@ -109,3 +111,45 @@ def get_existing_deletion_request(record_id):
     )
     if existing_requests.total > 0:
         return list(existing_requests)[0]
+
+
+def evaluate_record_deletion(record: RDMRecord):
```
ntarocco: Is this the same changes as in the other PR (it might include the same commit)?

carlinmack: yes sorry the other PR includes this PR and https://github.com/inveniosoftware/invenio-app-rdm/pull/3200 as I presume they could be merged first. sorry for the confusion

slint (approved):

PR #3203, opened 2025-10-13
Guest access request UI: Extend handling of possible expiration date values

### LINKED ISSUE #3202: Guest access requests: setting expiration date to "" breaks request
Package version (if known): 13.0.3

## Describe the bug
In the guest access request timeline, the record owner can select a 

## Steps to Reproduce
1. Create a guest access request
2. Open the request timeline as record owner
3. Clear the value of the expiration date
4. Click "save"
5. Now the request has `payload.secret_link_expiration: "NaN"` and the request is broken

## Expected behavior

The payload gets set to a valid value and the request isn't broken

## Screenshots (if applicable)

Image

---

Image

max-moser:
Closes #3202 

Some context for the reasoning is in the commit messages

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/requests/timeline/AccessRequestTimelineEdit.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/requests/timeline/AccessRequestTimelineEdit.js
@@ -76,17 +76,17 @@ export class AccessRequestTimelineEdit extends Component {
   initFormValues = () => {
     const {
       request: {
-        payload: { secret_link_expiration },
+        payload: { secret_link_expiration: secretLinkExpiration },
       },
     } = this.props;
 
-    if (parseInt(secret_link_expiration) === 0) {
+    if (parseInt(secretLinkExpiration) === 0 || secretLinkExpiration === "NaN") {
       return {
         secret_link_expiration: "",
       };
     }
     const dateFromDays = DateTime.now().plus({
-      days: parseInt(secret_link_expiration),
+      days: parseInt(secretLinkExpiration),
     });
     return {
       secret_link_expiration: dateFromDays.toISODate(),


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/requests/timeline/AccessRequestTimelineRead.js
```diff
@@ -20,7 +20,7 @@ export class AccessRequestTimelineRead extends Component {
     } = this.props;
 
     const expirationDate =
-      secretLinkExpiration === 0
+      parseInt(secretLinkExpiration) === 0
```
max-moser, line 23: this is backwards compatible since `parseInt(0) == parseInt("0")`

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/requests/timeline/AccessRequestTimelineEdit.js
```diff
@@ -76,17 +76,17 @@ export class AccessRequestTimelineEdit extends Component {
   initFormValues = () => {
     const {
       request: {
-        payload: { secret_link_expiration },
+        payload: { secret_link_expiration: secretLinkExpiration },
       },
     } = this.props;
 
-    if (parseInt(secret_link_expiration) === 0) {
+    if (parseInt(secretLinkExpiration) === 0 || secretLinkExpiration === "NaN") {
```
Samk13, line 83:
```suggestion
    if (isNaN(parseInt(secretLinkExpiration)) || parseInt(secretLinkExpiration) === 0) {
```
String comparison to `"NaN"` is incorrect since `parseInt("NaN")` returns `NaN`, not the string `"NaN"` no?

max-moser: the field has the actual string `"NaN"` stored, as visible in the screenshots: https://github.com/inveniosoftware/invenio-app-rdm/issues/3202
checking `parseInt(...) == NaN` might catch other string values as well that mark special values, not sure here

Samk13: Oh, if "NaN" is coming through as a literal string, that’s a backend/serialization bug and should be fixed upstream instead. 
I'm just marking this as resolved.

max-moser: agreed, then we can kick this bad boy out again after

PR #3214, opened 2025-10-16
fix(admin): use configured base template instead of hard-coded value

max-moser:
Tested with a fresh v14 `my-site`, and our own setup, looks the same in both cases (but fixes Flask-DebugToolbar in our case)

### DIFF
--- a/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/administration_page.html
+++ b/invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/administration_page.html
@@ -7,7 +7,11 @@
   under the terms of the MIT License; see LICENSE file for more details.
 #}
 
-{%- extends "invenio_theme/page.html" -%}
+{% if base_template|default(false) %}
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/templates/semantic-ui/invenio_app_rdm/administration_page.html
```diff
@@ -7,9 +7,9 @@
   under the terms of the MIT License; see LICENSE file for more details.
 #}
 
-{%- extends "invenio_theme/page.html" -%}
+{%- extends config.BASE_TEMPLATE -%}
```
zzacharo: nit: I am not sure if we want to be more flexible like https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/templates/semantic-ui/invenio_requests/details/index.html#L10C1-L14C12

max-moser: done!

PR #3226, opened 2025-11-03
setup: bump oauthclient major version

palkerecsenyi:
* Bumping the major version of invenio-oauthclient to 6.0.0

* This includes https://github.com/inveniosoftware/invenio-oauthclient/pull/360 which is not necessarily breaking but requires a manual DB migration for very large instances, as documented in the module's upgrade guide.

* Further changes are yet come in v6 before RDM v14, such as https://github.com/inveniosoftware/invenio-oauthclient/pull/328.

**Note: CI is currently failing due to a version mismatch of oauthclient on transitive dependencies. This will be resolved once everything is merged.** See [the full list](https://github.com/inveniosoftware/invenio-oauthclient/pull/360#issuecomment-3485029407) of PRs that need to be merged.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,7 @@
 Changes
 =======
 
-Version v14.0.0b3.dev0 (released 2025-XX-XX)
+Version v14.0.0b3.dev0 (released 2025-11-12)
 
 - setup: bump major version of `invenio-oauthclient`
 


CHANGES.rst
```diff
@@ -10,6 +10,11 @@
 Changes
 =======
 
+Version v14.0.0b2.dev3 (released 2025-XX-XX)
+
+- setup: bump major version of `invenio-oauthclient`. This contains some DB migrations but is not a breaking change,
```
utnapischtim: nit: in my opinion the line should be shorter in the release notes in CHANGES.rst so for me `setup: bump major version of invenio-oauthclient` would be enough.

palkerecsenyi: Happy to do this, since we are including the details in the PR description anyway

PR #3232, opened 2025-11-07
fix(deletion-request): removal reasons in landing page data attribute

### LINKED ISSUE #3211: records-ui: premature AJAX requests to fetch deletion request removal reasons
Package version (if known): 14.0.0b1.dev5

## Describe the bug

When navigating to a record landing page of a record you own, there are two AJAX requests being sent to fetch the valid deletion request removal reasons, even though the record deletion modal is not opened.

Image

## Steps to Reproduce
1. Go to a record landing page of a record you own
2. See two AJAX requests to `/api/vocabularies/removalreasons?q=tags:deletion-request` in "Network" tab, even though the "Manage > Delete record" modal wasn't opened

## Expected behavior

We should either:

- Check what we do on the deposit page, where we have the same modal, and it doesn't have this issue.
- Only **after clicking the modal**, make **one** request to fetch the removal reasons
- Preload (and cache?) and inject via Jinja the removal reasons as props or as part of the `record_deletion` variable.

ptamarit:
* Fixes #3211 
* The deposit form was already getting the removal reasons from the DOM, [via `deposits-config`](https://github.com/inveniosoftware/invenio-app-rdm/blob/e2e5423e84b02a6b01b00cf24ac188d7688681fb/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/index.js#L23), then [passed as an option to `RecordDeletion`](https://github.com/inveniosoftware/invenio-app-rdm/blob/e2e5423e84b02a6b01b00cf24ac188d7688681fb/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js#L783)
* The landing page does not have this config containing the vocabularies, so in this PR I am adding the removal reasons as to a new `data-record-deletion-options`, and passing it to the `RecordManagement` component via the `recordDeletionOptions` prop

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordDeletion.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordDeletion.js
@@ -33,7 +33,6 @@ export const RecordDeletion = ({
                 fluid
                 as="a"
                 href={recordDeletion.existing_request}
-                disabled={disabled}
                 className="basic"
                 color="red"
                 aria-haspopup="dialog"


ptamarit:
> Looks good, just 2 things:
> 
>     1. Can you also confirm that the mobile view still works fine and we are only making a single request now?

* :heavy_check_mark: Tested, the mobile view still works fine.
* :heavy_check_mark: We are now making 0 API requests (the removal reasons are inserted in the DOM in the same fashion as it's done in the deposit form).

>     2. The linting tests

:heavy_check_mark: Sorry about this, fixed.

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -158,6 +162,7 @@ def record_detail(
     record_ui = UIJSONSerializer().dump_obj(record.to_dict())
 
     record_deletion = evaluate_record_deletion(record, g.identity)
+    record_deletion_options = VocabulariesOptions().removal_reasons()
```
slint:
```suggestion
    if record_deletion:
        record_deletion_options = VocabulariesOptions().removal_reasons(extra_filter=dsl.Q("term", tags="record-deletion"))
```

- minor: we should wrap this, or else we fetch the remoal reasons on all record pages, regardless of the record deletion feature being enabled or not
- nit: we can actually add a filter parameter or similar to only fetch the removal reasons with `tags: record-deletion` (see also how it's done already in `depositable_resource_types`).
- shelve: I don't think these are cached, which is a bit of shame. I recall the facets vocabulary resolutions though being cached, but it's not critical as well if we are a bit more defensive on when we fetch

slint (commented):
Some comments on being a bit more defensive with fetching (and a nit + shelvable issue)

invenio_app_rdm/records_ui/views/records.py
ptamarit: > * minor: we should wrap this, or else we fetch the remoal reasons on all record pages, regardless of the record deletion feature being enabled or not

:white_check_mark: Done, I'm now doing:

```python
    # The deletion options are only needed if the user is allowed to delete the given record.
    record_deletion_options = (
        VocabulariesOptions().deletion_request_removal_reasons()
        if record_deletion["allowed"]
        else []
    )
```

> * nit: we can actually add a filter parameter or similar to only fetch the removal reasons with `tags: record-deletion` (see also how it's done already in `depositable_resource_types`).

:white_check_mark: We were actually already filtering `removal_reasons` to only include the ones with the tag `record-deletion`. I renamed the method and vocabulary key from `removal_reasons` to `deletion_request_removal_reasons` to make this more clear.

> * shelve: I don't think these are cached, which is a bit of shame. I recall the facets vocabulary resolutions though being cached, but it's not critical as well if we are a bit more defensive on when we fetch

:information_source: Agree that it should be fixed at some point.

invenio_app_rdm/records_ui/utils.py
```diff
@@ -124,7 +124,11 @@ def evaluate_record_deletion(record: RDMRecord, identity):
         or rec_del["request_deletion"].valid_user
     )
     rd_allowed = immediate.allowed or request.allowed
-    existing_request = get_existing_deletion_request(record.id)
+
+    # We show existing requests to valid users (even if they are not allowed to delete a record anymore).
+    existing_request = (
+        get_existing_deletion_request(record.id) if rd_valid_user else None
+    )
```
slint: nit: maybe directly assign below in `record_deletion["existing_request"] = ...` instead of doing before the `if ...` so the logic is close together?

slint (approved):
LGTM, tiny code organization nit

invenio_app_rdm/records_ui/utils.py
ptamarit: Done, I've also modified `get_existing_deletion_request` to directly return `["links"]["self_html"]`, so that then I can simply do this in the assignment:

```python
    record_deletion["existing_request"] = (
        # We show existing requests to valid users (even if they are not allowed to delete a record anymore).
        get_existing_deletion_request(record.id)
        if rd_valid_user
        else None
    )
```

(The method `get_existing_deletion_request` is not used anywhere else.)

PR #3252, opened 2025-11-24
Fix UX: Inactivate Optional Accordions When No Data Is Present

### LINKED ISSUE #1414: Remove fields in plain 'new upload' that look as they are required, but are not

When creating a new record there are fields marked as 'required' but a record can be published without filling these fields.

The 'not' needed fields are 'Date', 'Alternate identifier(s)', 'Related works'
image
image

Remove these fields as you can do manually:

image

## Additional context
1. There are two kinds of 'required' fields.

- If the star is on the heading it is necessary to create a record
image
- If the star is on a field within a heading it is only needed if you fill out the fields on the same line
image
  Note: the star is not on the heading, but only on two fields that either both are empty or both filled. I haven't seen the difference until @fenekku pointed it out to me.

2. The removal would might also affect this issue: https://github.com/inveniosoftware/invenio-app-rdm/issues/1407
- Case A: "Also, it will complain again about the seemingly required list fields (dates, alternate identifiers, related works)."

Samk13:

### Description

Hopefully closes: https://github.com/inveniosoftware/invenio-app-rdm/issues/1414 🤞

This PR cleans up several deposit form fields and accordion behaviors to reduce confusion for new users. 
In demos, users often try to fill optional fields (e.g., Alternate Identifier, Related works, ...) because the red star suggests they are required. 
By removing showEmptyValue and keeping accordions closed when no data is present, the form shows fewer empty optional fields and avoids misleading cues.

### Key points:

* Prevents users from thinking optional fields are required.
* Accordion sections stay inactive when empty.
* Provides a cleaner and clearer initial form experience.

After:
Screenshot 2025-11-24 at 23 19 17

Before:

Screenshot 2025-11-24 at 23 20 52

### Checklist

**Frontend**

**Reminder**

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
@@ -2,7 +2,7 @@
  * SPDX-FileCopyrightText: 2020-2025 CERN.
  * SPDX-FileCopyrightText: 2020-2022 Northwestern University.
  * SPDX-FileCopyrightText: 2021-2022 Graz University of Technology.
- * SPDX-FileCopyrightText: 2022-2024 KTH Royal Institute of Technology.
+ * SPDX-FileCopyrightText: 2022-2025 KTH Royal Institute of Technology.
  * SPDX-License-Identifier: MIT
  */
 
@@ -538,7 +538,7 @@ export class RDMDepositForm extends Component {
                   <AccordionField
                     includesPaths={this.sectionsConfig["funding-section"]}
                     severityChecks={this.severityChecks}
-                    active
+                    active={!!_get(record, "metadata.funding.length")}
                     label={i18next.t("Funding")}
                     ui={this.accordionStyle}
                     id="funding-section"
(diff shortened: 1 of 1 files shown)


tmorrell:
I'm 100% supportive of removing the blank fields which cause the confusing star to show up.

I think showing/hiding fields needs more community discussion. My opinion is funding and related works should be open, alternate identifiers and references closed.

I also think this is likely a v14 change.

RosaLo:
I agree with tmorrell that it would be good to have more community discussion on this. We reasoned that it would be logically consistent to have all sections that are non-obligatory "folded-in"/inactive accordions so that users shouldn't be confused with having non-obligatory fields marked with a * that we observed gave them a hint that they must fill in something in all of the fields with asterisk also in the non-obligatory sections. But there are surely several alternative ways to make the deposit form easier  to understand for first-time end-users without losing options to add optional metadata.

monotasker:
In general, I think this is a really good idea. Maybe it could be configurable per field? I suspect that different instances might want different sections to remain open by default, to encourage people to fill them in. Since people are more likely to complete an open section (if they're not overwhelmed) than to open an accordion themselves. In other words, the choice of open sections is a lever for steering user behaviour. So a config variable would allow an instance to steer user focus to the optional fields they want to highlight.

This doesn't solve the issue of the confusing required asterisks for those fields, of course. But some might want to make that trade-off for one or two of the optional sections.

Of course, that configurability would make this a bigger change.

Samk13:
That's a good point @monotasker. My initial thought is that we might not need an additional configuration layer for this, since instances that want to keep specific sections open can already customize this through the existing overrides.

I'd lean towards keeping the default behavior simple and predictable: empty optional sections are collapsed by default, while sections containing values are open.

But this is exactly the kind of feedback I was hoping to get from the community. 
If others have use cases where the existing override mechanism isn't sufficient, or where having this explicitly configurable would be valuable, it would be great to hear them.

wgresshoff:
This is definitely a good enhancement! But I also have a slight preference of making the optional sections collapsible  by configuration, not by default.

PR #3263, opened 2025-12-09
feat: add remove_community permission check

jrcastro2:
Add permission control for removing records from communities:
- Add remove_community to permissions list in record detail view
- Conditionally render RemoveFromCommunityAction based on can_remove_community permission
- closes https://github.com/CERNDocumentServer/cds-rdm/issues/630

### DIFF
--- a/invenio_app_rdm/records_ui/views/records.py
+++ b/invenio_app_rdm/records_ui/views/records.py
@@ -302,6 +302,7 @@ def record_detail(
                 "moderate",
                 "request_deletion",
                 "immediately_delete",
+                "remove_community_from_record",
             ]
         ),
         custom_fields_ui=custom_fields["ui"],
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/records_ui/views/records.py
```diff
@@ -302,6 +302,7 @@ def record_detail(
                 "moderate",
                 "request_deletion",
                 "immediately_delete",
+                "remove_community",
```
kpsherva: I have the same naming comment as in the other PR, I think we should be explicit that it is about removing community from the record (or record from the community)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RemoveFromCommunity/RemoveFromCommunityAction.js
```diff
@@ -185,4 +199,5 @@ RemoveFromCommunityAction.propTypes = {
   result: PropTypes.object.isRequired,
   recordCommunityEndpoint: PropTypes.object.isRequired,
   successCallback: PropTypes.func.isRequired,
+  canRemoveCommunity: PropTypes.bool.isRequired,
```
kpsherva: for backwards compatibility, I think the default should be set to True - I think if I merge this without releasing rdm-records first, then canRemoveCommunity will be evaluated to False
Also if you set the prop type bool to be required, if the value is false then it gets improperly evaluated as "missing", leading to console errors

PR #3284, opened 2026-01-07
fix(config): remove value for the deprecated WSGI_PROXIES variable

max-moser:
The configuration option `WSGI_PROXIES` has been deprecated in `Invenio-Base` a while ago, in favor of the more granular `PROXYFIX_CONFIG`.
The latter now gets used by default in `docker-services.yml` in the PR https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/323

Since the value is provided in the cookiecutter already and because providing an incorrect value as fallback doesn't really provide any benefits, I've simply removed the old configuration value without replacement here.


fenekku (approved):
Thanks. Add a quic line in https://github.com/inveniosoftware/docs-invenio-rdm v14 release notes about the breaking change.

PR #3285, opened 2026-01-09
feat: add max-width to tombstone container

OliverGeneser:
### Description

Sets max-width limit on the tombstone container, so it looks better on larger screens.

#### Before
image

#### After
image

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/container.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/container.overrides
@@ -82,6 +82,7 @@
       width: 100% !important; // Overwriting semantic-ui's !important. Please don't remove.
     }
   }
+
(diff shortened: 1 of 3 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/container.overrides
```diff
@@ -93,6 +94,10 @@
       margin-right: 0 !important;
     }
   }
+
+  &.max-width-3\/5 {
```
slint:
```suggestion
  &.max-w-3\/5 {
```

nit: to make the TailwindCSS transition smooth as 🧈

Joke aside, I prefer the shorter utility classes.

OliverGeneser: Changed, I do too :laughing:

PR #3288, opened 2026-01-13
feat(records_ui): Add preview_file arg to record landing page

sakshamarora1:
closes: #3289

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/theme.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/theme.js
@@ -16,22 +16,25 @@ $("#record-conceptdoi-badge").on("click", function () {
   $("#conceptdoi-modal").modal("show");
 });
 
-$(".preview-link").on("click", function (event) {
-  const fileKey = event.target.dataset.fileKey;
-  $("#preview-file-title").html(fileKey);
+$("#file-list-table")
+  .find(".preview-link")
+  .on("click", function (event) {
+    const fileKey = event.target.dataset.fileKey;
+    $("#preview-file-title").text(fileKey);
 
-  event.preventDefault(); // Prevent default link navigation for back button to navigate the record and not the iframe
+    event.preventDefault(); // Prevent default link navigation for back button to navigate the record and not the iframe
 
-  const newUrl = window.location.pathname + "?preview_file=" + fileKey;
-  window.history.replaceState(null, "", newUrl);
+    const newUrl = new URL(window.location);
+    newUrl.searchParams.set("preview_file", fileKey); // .set method automatically encodes the value
+    window.history.replaceState(null, "", newUrl);
 
-  // Update iframe with the updated URL to the preview file without adding to browser history
-  const previewUrl = $(this).attr("href");
-  const iframe = document.getElementById("preview-iframe");
-  if (iframe && iframe.contentWindow) {
-    iframe.contentWindow.location.replace(previewUrl);
-  }
-});
+    // Update iframe with the updated URL to the preview file without adding to browser history
+    const previewUrl = $(this).attr("href");
+    const iframe = document.getElementById("preview-iframe");
+    if (iframe && iframe.contentWindow) {
+      iframe.contentWindow.location.replace(previewUrl);
+    }
+  });
 
 // Export dropdown on landing page
 $(".dropdown.export").dropdown({
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
@@ -90,7 +90,7 @@ To access its content, please download it and open it locally.') }}
     permissions=None,
     display_name=None
 ) %}
-  <table class="ui striped table files fluid {{record.ui.access_status.id}}">
+  <table id="file-list-table" class="ui striped table files fluid {{record.ui.access_status.id}}">
     <thead>
       <tr>
         <th>{{_('Name')}}</th>


invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -188,6 +185,33 @@ def view(**kwargs):
                 except NoResultFound:
                     # If the parent pid is being used we can get the id of the latest record and redirect
                     latest_version = service().read_latest(**read_kwargs)
+                    # First check if a filename is present in the request args
+                    if filename:
+                        if filename in latest_version["files"]["entries"].keys():
```
OliverGeneser: I'm still new in the data structure :sweat_smile: But if latest_version/files don't contain files/entries this throws an KeyError... Maybe .get() should be used?

Also .keys() is not need, but used here and not in line 205 :thinking:

sakshamarora1: Thanks, you're right! After some discussion, I have removed from here as this snippet belongs to the overlay.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/theme.js
```diff
@@ -17,7 +17,20 @@ $("#record-conceptdoi-badge").on("click", function () {
 });
 
 $(".preview-link").on("click", function (event) {
-  $("#preview-file-title").html(event.target.dataset.fileKey);
+  const fileKey = event.target.dataset.fileKey;
+  $("#preview-file-title").html(fileKey);
+
+  event.preventDefault(); // Prevent default link navigation for back button to navigate the record and not the iframe
+
+  const newUrl = window.location.pathname + "?preview_file=" + fileKey;
```
OliverGeneser: The fileKey should probably be encoded with encodeURIComponent, so we mitigate any URL reflection risks

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/theme.js
```diff
@@ -17,7 +17,20 @@ $("#record-conceptdoi-badge").on("click", function () {
 });
 
 $(".preview-link").on("click", function (event) {
-  $("#preview-file-title").html(event.target.dataset.fileKey);
+  const fileKey = event.target.dataset.fileKey;
+  $("#preview-file-title").html(fileKey);
```
OliverGeneser: fileKey shouldn't be trusted with .html() as that opens op to XSS attacks.

Example
```<a class="preview-link" data-file-key="">PreviewLink</a>```

File: invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/theme.js [hunk 4]
```diff
@@ -17,7 +17,20 @@ $("#record-conceptdoi-badge").on("click", function () {
 });
 
 $(".preview-link").on("click", function (event) {
```
ntarocco: is there any ID to use here first, to have better performance? Otherwise jQuery will look for all the classes `preview-link` in all the DOM.
E.g. `$("#myid").find(".preview-link")...`

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -111,6 +111,17 @@ def view(**kwargs):
     return view
 
 
+def pass_preview_file(f):
+    """Decorate a view to pass the preview file."""
+
+    @wraps(f)
+    def view(**kwargs):
+        kwargs["preview_file"] = request.args.get("preview_file")
```
ntarocco, line 132: this should be sanitized

kpsherva (changes requested):
Blocking this from merging, due to unsanitized input

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/theme.js
sakshamarora1: Added `"#file-list-table"` to the parent div.

invenio_app_rdm/records_ui/views/decorators.py
sakshamarora1: I have added `secure_filename` for sanitization.
However we make sure that we never pass unknowns. If the filename is not found, then we go for the default file here in this filter:
https://github.com/inveniosoftware/invenio-app-rdm/blob/59c87b288cb08be2127d310d5f695fb20a27da4f/invenio_app_rdm/records_ui/views/filters.py#L38-L48

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/theme.js
sakshamarora1: Thanks! Added

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -26,6 +26,7 @@
 from invenio_rdm_records.services.errors import RecordDeletedException
 from invenio_records_resources.services.errors import PermissionDeniedError
 from sqlalchemy.orm.exc import NoResultFound
+from werkzeug.utils import secure_filename
```
palkerecsenyi: [This function](https://tedboy.github.io/flask/generated/werkzeug.secure_filename.html) might be too restrictive, it removes _all_ non-ascii characters from the file. For example, the following characters (that are valid in UNIX file names) are removed:

- Non-Latin alphabets (cyrillic, arabic, etc)
- All types of spaces
- Most special characters (e.g. `#`)

Files with these characters are valid and accepted elsewhere on InvenioRDM, so it would make it impossible to get a preview to a correctly named file.

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -202,6 +210,9 @@ def record_detail(
 
     record_ui = UIJSONSerializer().dump_obj(record.to_dict())
 
+    if preview_file:
+        record_ui["files"]["default_preview"] = preview_file
```
palkerecsenyi, line 218: nit: maybe would be neater to have this as a separate template variable since it isn't really part of the record itself

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/theme.js
```diff
@@ -16,9 +16,25 @@ $("#record-conceptdoi-badge").on("click", function () {
...
+$("#file-list-table")
+  .find(".preview-link")
+  .on("click", function (event) {
+    const fileKey = event.target.dataset.fileKey;
+    $("#preview-file-title").html(fileKey);
+
+    event.preventDefault(); // Prevent default link navigation for back button to navigate the record and not the iframe
+
+    const newUrl =
+      window.location.pathname + "?preview_file=" + encodeURIComponent(fileKey);
```
palkerecsenyi: what if the current URL already has a query param or a fragment/anchor at the end? This would remove it. Not a huge problem since I don't think we're using any such params or anchors on the record page at the moment, but might be nicer to do it like this to avoid surprises in the future if we add anything else:


```suggestion
const newUrl = new URL(window.location);
newUrl.searchParams.set("preview_file", fileKey);
window.history.replaceState(null, "", newUrl);
```

invenio_app_rdm/records_ui/views/decorators.py
ntarocco: You should check what is done in InvenioRDM when uploading a file, and use the same sanitization.

palkerecsenyi (approved):

invenio_app_rdm/records_ui/views/records.py
sakshamarora1: This was done to not require changes in templates overriding this template downstream.

OliverGeneser (approved):

invenio_app_rdm/records_ui/views/decorators.py
kpsherva: @sakshamarora1 where did you add the secure_filename?

sakshamarora1: Removed the function since it is not accommodating of other possible filenames, see Pal's comment: https://github.com/inveniosoftware/invenio-app-rdm/pull/3288#discussion_r2693867324

PR #3290, opened 2026-01-13
feat(previewer): show message for outdated records

palkerecsenyi:
Closes https://github.com/inveniosoftware/invenio-previewer/issues/246

---

* Added a message shown in the file preview template when the file's record is outdated and a newer one is available. The message aims to avoid implying that there is a new version of that specific file available, since the file itself might not have changed or might have been deleted. The message has almost the same style as the one on the record detail page.

* The message is only shown in the full-screen preview, since users often share links to that directly. It is not shown in the record detail page iframe, since there's already a message there about the record being outdated. The widely-supported Sec-Fetch-Dest header is used to detect whether the preview is being requested inside an iframe.

* This is an RDM-specific feature so it makes sense to include it in this repository as an override to the templates.

image

### DIFF
--- a/invenio_app_rdm/records_ui/views/decorators.py
+++ b/invenio_app_rdm/records_ui/views/decorators.py
@@ -111,6 +111,19 @@ def pass_is_preview(f):
     return view
 
 
+def pass_is_iframe(f):
+    """Decorate a view to check if it's being requested from inside an iframe."""
+
+    @wraps(f)
+    def view(**kwargs):
+        # See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Dest
+        header_value = request.headers.get("Sec-Fetch-Dest")
+        kwargs["is_iframe"] = header_value == "iframe"
+        return f(**kwargs)
+
+    return view
+
+
 # TODO to be improved as a request arg schema (following the REST views)
 def pass_include_deleted(f):
     """Decorate a view to check if it's a include deleted."""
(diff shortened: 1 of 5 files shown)


invenio_app_rdm/theme/templates/semantic-ui/invenio_previewer/rdm_abstract_previewer.html
```diff
@@ -0,0 +1,30 @@
...
+    {{ webpack[css_bundle] }}
+  {% endfor %}
+  {{ webpack['theme.css'] }}
+{%- endblock %}
+
+{%- block page_body %}
+  {%- if request.args.get('embed') %}
+    {%- include 'invenio_previewer/top.html' %}
+  {%- endif %}
+  {%- if not file.is_iframe %}
```
ntarocco, line 23: if `is_iframe` is true, does it show the preview? ☺️

palkerecsenyi: Yes, the preview is shown in the `panel` block which is overridden by each of the individual file type specific templates in invenio-previewer. This `if` just shows/hides `latest_record.html` which is just the message banner.

invenio_app_rdm/records_ui/views/decorators.py
```diff
@@ -111,6 +111,19 @@ def view(**kwargs):
     return view
 
 
+def pass_is_iframe(f):
+    """Decorate a view to check if it's being requested from inside an iframe."""
+
+    @wraps(f)
+    def view(**kwargs):
+        # See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Dest
+        header_value = request.headers.get("Sec-Fetch-Dest")
```
kpsherva, line 120: this change possible prevent from the banner to be visible while embedding in other websites (if intance's policy allows embedding) @ntarocco are we OK with that for RDM?

PR #3291, opened 2026-01-13
style: collapsible messages style

OliverGeneser:
Partially closes inveniosoftware/invenio-requests#492

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/feed.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/feed.overrides
@@ -156,6 +156,50 @@
       margin-top: .5rem;
     }
 
+    .collapsible-comment {
+      display: block;
+      position: relative;
+
+      .collapsible-comment-inner {
+        display: block;
+        overflow: hidden;
+        transition: max-height 300ms ease;
+        max-height: @collapsibleHeight;
+      }
+
+      &.expanded .collapsible-comment-inner {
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/feed.overrides
```diff
@@ -184,6 +228,14 @@
       }
     }
 
+    .requests-event-container {
+      --collapsible-color: #f9f9f9;
```
kpsherva, line 232: nit (not a blocker): I would recommend using the color variables from the theme to keep color consistency

invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/feed.overrides
```diff
@@ -156,6 +156,50 @@
...
 
+    .collapsible-comment {
+      display: block;
+      position: relative;
+
+      .collapsible-comment-inner {
+        display: block;
+        overflow: hidden;
+        transition: max-height 300ms ease;
+        max-height: 200px;
```
zzacharo: @OliverGeneser can we extract this to `feed.variables` e.g `@collapsibleHeight` and use it also below in [line182](https://github.com/inveniosoftware/invenio-app-rdm/pull/3291/changes#diff-da26d0ee16fc7010bb995c28f6e56addd3272b4ecd19140ae81157e356aa5dbcR182) ?

PR #3293, opened 2026-01-14
Feature: overridable version box list items

sakshamarora1:
related: https://github.com/CERNDocumentServer/cds-rdm/issues/671

### DIFF
--- a/invenio_app_rdm/administration/audit_logs/audit_logs.py
+++ b/invenio_app_rdm/administration/audit_logs/audit_logs.py
@@ -6,6 +6,7 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """Invenio administration view module for audit logs."""
+
 from flask import current_app
 from invenio_administration.views.base import AdminResourceListView
 from invenio_i18n import lazy_gettext as _
--- a/invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py
@@ -38,6 +38,7 @@ This script has been tested with following data:
   - drafts visible
   - records visible
 """
+
 import sys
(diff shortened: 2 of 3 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/CopyButton.js
```diff
@@ -26,18 +26,19 @@ class SimpleCopyButton extends React.Component {
...
-    const { hoverState, size } = this.props;
+    const { hoverState, className, uiProps } = this.props;
 
     return (
       <Button
-        className="copy"
+        className={`copy ${className || ""}`}
         basic
-        size={size}
+        size="medium"
```
kpsherva: you can still leave the defaultProp, it shouldn't conflict with `{...uiProps}`normally. If you pass size in the uiProps, the customization it should take over.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/CopyButton.js
```diff
@@ -118,7 +121,7 @@ export class CopyButton extends Component {
               onCopy={this.onCopy}
               url={url}
               hoverState={this.hoverStateHandler}
-              size={size}
+              uiProps={uiProps}
```
kpsherva, line 123: same comment as above

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/CopyButton.js
sakshamarora1: Discussed IRL: size by default is medium, so removed it.

kpsherva (approved):

PR #3301, opened 2026-01-26
Style for disabled comment submit button

OliverGeneser:
Closes https://github.com/inveniosoftware/invenio-requests/issues/512

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/button.overrides
+++ b/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/button.overrides
@@ -111,3 +111,14 @@
     transform: none;
   }
 }
+
+/* accessible disabled buttons */
+button[aria-disabled=true] {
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/feed.overrides
```diff
@@ -59,6 +59,14 @@
     }
   }
 
+  button[aria-disabled=true] {
```
zzacharo: shouldnt that be in [button.overrides](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/elements/button.overrides) ? ping @kpsherva

kpsherva: yes, it should, please move it to the button.overrides since it is global button element change

PR #3304, opened 2026-01-27
feat (deposit): allow overriding deposit serializer

egabancho:
This is how you can use it
```js
import { getInputFromDOM } from "@js/invenio_rdm_records";
import { DepositFormApp } from "@js/invenio_rdm_records/src/deposit/api/DepositFormApp";
import { AwesomeSerializer} from "@js/site/records/deposit/DepositRecordSerializer"

const depositConfig = getInputFromDOM("deposits-config");
const parametrizedRDMDepositForm = parametrize(DepositFormApp, {
  recordSerializer: new AwesomeSerializer(
    depositConfig.default_locale,
    depositConfig.custom_fields.vocabularies,
  ),
});

export const overriddenComponents = {
  // Deposit
  "InvenioAppRdm.Deposit.RDMDepositForm.layout": parametrizedRDMDepositForm,
...
}
```

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
@@ -144,6 +144,7 @@ export class RDMDepositForm extends Component {
         groupsEnabled={groupsEnabled}
         allowEmptyFiles={allowEmptyFiles}
         customFieldsUI={customFieldsUI}
+        recordSerializer={null}
       >
         <DepositFormApp
           config={this.config}
@@ -152,6 +153,7 @@ export class RDMDepositForm extends Component {
           files={files}
           permissions={permissions}
           errors={record.errors}
+          recordSerializer={null}
         >
           <Overridable
             id="InvenioAppRdm.Deposit.FormFeedback.container"


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js
```diff
@@ -144,6 +144,7 @@ export class RDMDepositForm extends Component {
         groupsEnabled={groupsEnabled}
         allowEmptyFiles={allowEmptyFiles}
         customFieldsUI={customFieldsUI}
+        recordSerializer={null}
```
kpsherva, line 147: could you explain how setting the prop to null helps to override it, because I am not sure I understand? 
https://github.com/inveniosoftware/invenio-rdm-records/pull/2243/changes#diff-6dc2b45d8268fce47e87bb05cff1c6cd6b968ed03e0b6cc91b5675bcf485b641R157 here the default prop is already undefined, isn't it enough ?

egabancho: I believe you need to add it to the overridable context to be able to `parametrize` it. I simply set it to `null` to preserve the same behaviour, but we could also set it to `RDMDepositRecordSerializer`, and the result will be the same (or `undefined` for that matter).

Issue #3320, opened 2026-02-05
[object Object] in the error banner


OliverGeneser:
## Additional context

Check PR's
https://github.com/inveniosoftware/invenio-app-rdm/pull/3321
https://github.com/inveniosoftware/react-invenio-forms/pull/327
https://github.com/inveniosoftware/invenio-rdm-records/pull/2250

PR #3321, opened 2026-02-05
fix: form accordion label

OliverGeneser:
Closes: https://github.com/inveniosoftware/invenio-app-rdm/issues/3320


OliverGeneser:
> if the label is already a string, why do we need to duplicate it into the data-label?

@carlinmack It was because of a check in RDM records, but I refactored that, so it's not needed anymore :smile:

PR #3322, opened 2026-02-06
fix: using app locale for citations

Ducica:
* app was using navigator.language which is always
* browser setting. The change enables using localization
* based on app settings and using translated strigns that exist in
* citeproc library

### Description

Currently, the citation is always generated based on browser selected locale. This change uses language selected in the app. Normally, in invenio you only have access to language code, so if the language setting in the browser starts with this language code it sends the entire locale i.e. en_GB, if not, it only sends en. The citeproc library has good behavior, even in case only language code is provided. 

Before:
image

After:

image

Noter that benefit depens on the citation style i.e. some styles that contain more text, other than record data benefit more. 

Please let us know what you think.

**Reminder**

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCitationField.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCitationField.js
@@ -58,7 +58,13 @@ export class RecordCitationField extends Component {
 
   fetchCitation = async (recordLinks, style, includeDeleted) => {
     const includeDeletedParam = includeDeleted === true ? "&include_deleted=1" : "";
-    const url = `${recordLinks.self}?locale=${navigator.language}&style=${style}${includeDeletedParam}`;
+    const navLang = (navigator.language || "").toLowerCase();
+    const uiLang = (i18next.language || "").toLowerCase();
+    const citationLocale =
+      uiLang && navLang.startsWith(uiLang)
+        ? navigator.language
+        : i18next.language || navigator.language;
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordCitationField.js
```diff
@@ -58,7 +58,10 @@ export class RecordCitationField extends Component {
 
   fetchCitation = async (recordLinks, style, includeDeleted) => {
     const includeDeletedParam = includeDeleted === true ? "&include_deleted=1" : "";
-    const url = `${recordLinks.self}?locale=${navigator.language}&style=${style}${includeDeletedParam}`;
+    const citationLocale = navigator.language.startsWith(i18next.language)
+      ? navigator.language
+      : i18next.language;
```
Samk13:
```suggestion
  const navLang = (navigator.language || "").toLowerCase();
  const uiLang = (i18next.language || "").toLowerCase();
  
  const citationLocale =
    uiLang && navLang.startsWith(uiLang)
      ? navigator.language
      : i18next.language || navigator.language;
```

Would it make sense to normalize casing and add a fallback here to avoid `en-US` vs `en-us` mismatches and prevent `locale=undefined` if `i18next.language` isn’t set yet, while keeping the same behavior otherwise?

Ducica: @Samk13 Thanks for the feedback, generally, I've never had issue with missing values in context of i18next or navigator, but your suggestion is more defensive, so I've modified it like you proposed. Regards

PR #3324, opened 2026-02-08
fix: add record files restriction option to enable cache-control header

tmorrell:

### Description

InvenioRDM doesn't currently add cache-control headers to file links, which can result in weird behavior if a cache is put in front of the repo. This is because the cache-control headers are only added by default for non-restricted files (which makes sense) https://github.com/inveniosoftware/invenio-files-rest/blob/616dc34c38006433653211feb5a9e523b0d399e7/invenio_files_rest/helpers.py#L182. But because InvenioRDM doesn't pass this option, the cache-control headers are never added.

This should be reviewed carefully to make sure it doesn't have security implications.

### Checklist

**Reminder**

### DIFF
--- a/invenio_app_rdm/records_ui/views/records.py
+++ b/invenio_app_rdm/records_ui/views/records.py
@@ -411,7 +411,12 @@ def record_file_download(pid_value, file_item=None, is_preview=False, **kwargs):
         obj = file_item._file.object_version
         emitter(current_app, record=file_item._record, obj=obj, via_api=False)
 
-    return file_item.send_file(as_attachment=download)
+    if file_item._record.access.protection.files == "public":
+        restricted = False
+    else:
+        restricted = True
+
+    return file_item.send_file(as_attachment=download, restricted=restricted)
 
 
 @pass_record_or_draft(expand=False)


invenio_app_rdm/records_ui/views/records.py
```diff
@@ -411,7 +411,7 @@ def record_file_download(pid_value, file_item=None, is_preview=False, **kwargs):
         obj = file_item._file.object_version
         emitter(current_app, record=file_item._record, obj=obj, via_api=False)
 
-    return file_item.send_file(as_attachment=download)
+    return file_item.send_file(as_attachment=download, restricted=False)
```
zzacharo: shouldt that be assigned to the record resrtiction?

egabancho: 🤔 By setting `restricted=False`, we are making `send_stream` to set the `Cache-Control` also to `public` [here](https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/helpers.py#L183).

According to [MDM](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control) 
> You should add the private directive for user-personalized content, especially for responses received after login and for sessions managed via cookies.

Maybe a good alternative would be to set `restricted` to the actual record/file restrictions, and then update `send_stream`/`redirect_stream` to set `Cache-Control` to `public` or `private`, respectively.

What do y'all think?

PR #3325, opened 2026-02-09
feat: make resource type badge clickable

### LINKED ISSUE #1200: UI: show full resource type term in "resource type badge" instead of truncated and make it clickable
I was confused by a record that was shown as type "other", but did not show up when I searched and filtered for that resource type.

After inspecting the JSON, I figured out that this record is "publication-other", but the badge is the same.

I think that to avoid confusion it would be better that the badges would show the full resource type name, even though its more verbose. 

Furthermore, I think the badge should be clickable and forward to a search with that type pre-selected as a filter.

OliverGeneser:
Closes: #1200 

### Description
image

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
@@ -48,6 +48,15 @@ class RecordsResultsListItem extends Component {
       "ui.resource_type.title_l10n",
       i18next.t("No resource type")
     );
+    const resourceTypeId = _get(result, "ui.resource_type.id", "");
+
+    const [mainType] = resourceTypeId.split("-");
+    const filterValue = resourceTypeId.includes("-")
+      ? `resource_type:${mainType}+inner:${resourceTypeId}`
+      : `resource_type:${resourceTypeId}`;
+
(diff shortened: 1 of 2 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/RecordsResultsListItem.js
```diff
@@ -47,6 +47,10 @@ class RecordsResultsListItem extends Component {
       "ui.resource_type.title_l10n",
       i18next.t("No resource type")
     );
+    const resourceTypeId = _get(result, "ui.resource_type.id", "");
+    const resourceTypeFilter = resourceTypeId.includes("-")
+      ? `resource_type%3A${resourceTypeId.split("-")[0]}%2Binner%3A${resourceTypeId}`
+      : `resource_type%3A${resourceTypeId}`;
```
carlinmack: Maybe you could [encodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) instead of hardcoding the escape sequences which could be brittle/error prone

carlinmack (commented):
Just one comment

carlinmack (approved):
reviewing the [latest force push](https://github.com/inveniosoftware/invenio-app-rdm/compare/5175aaffed9803688b9c9420dc3e5834879c02ba..7ca24a7174d23d3344044934ed1a7fa23c67dd22), looks good to me

PR #3332, opened 2026-02-19
fix(administration-user-access): user permission based ACL

zzacharo:


invenio_app_rdm/administration/users/permissions.py
```diff
@@ -19,10 +20,9 @@ def can_access_user_administration(identity=None):
...
-    provides = identity.provides
     admin_need = RoleNeed("administration")
     moderator_need = RoleNeed("administration-moderation")
 
-    return (superuser_access in provides) or (
-        admin_need in provides and moderator_need in provides
-    )
+    permission = Permission(admin_need, moderator_need)
+    # superuser-access is always added as an explicit need
+    return permission.allows(identity)
```
Samk13: It still means `administration` OR `administration-moderation` OR `superuser`, while we need `superuser` OR `(administration AND administration-moderation)`
Maybe we should do it as:
```py
if Permission(superuser_access).allows(identity):
return True

provides = identity.provides
admin_need = RoleNeed("administration")
moderator_need = RoleNeed("administration-moderation")
return admin_need in provides and moderator_need in provides
```

PR #3335, opened 2026-02-23
fix(db): use UTC for Postgres

palkerecsenyi:
This PR overrides config that's originally specified in `invenio-db`. This change is therefore duplicated there: https://github.com/inveniosoftware/invenio-db/pull/196

* Right now, we don't have any restriction/requirement on the database's time zone. This can lead to unexpected behaviour when reading/writing timestamps with time zones, since they get shifted to the database's zone which might not be UTC.

* The easiest way to ensure the DB is using UTC is by specifying it on connection via SQLAlchemy. This overrides any default value specified in e.g. postgresql.conf.

* Instances can still easily override this if they should need to, but this is not recommended.

### DIFF
--- a/invenio_app_rdm/config.py
+++ b/invenio_app_rdm/config.py
@@ -528,6 +528,10 @@ SQLALCHEMY_ENGINE_OPTIONS = {
     "pool_recycle": 3600,
     # set a more agressive timeout to ensure http requests don't wait for long
     "pool_timeout": 10,
+    # Ensure the database is using the UTC timezone for interpreting timestamps (Postgres only).
+    # This overrides any default setting (e.g. in postgresql.conf). Invenio expects the DB to receive
+    # and provide UTC timestamps in all cases, so it's important that this doesn't get changed.
+    "connect_args": {"options": "-c timezone=UTC"},
 }
 """SQLAlchemy engine options.
 


invenio_app_rdm/config.py
```diff
@@ -528,6 +528,10 @@ def files_rest_permission_factory(obj, action):
     "pool_recycle": 3600,
     # set a more agressive timeout to ensure http requests don't wait for long
     "pool_timeout": 10,
+    # Ensure the database is using the UTC timezone for interpreting timestamps (Postgres only).
+    # This overrides any default setting (e.g. in postgresql.conf). Invenio expects the DB to receive
+    # and provide UTC timestamps in all cases, so it's important that this doesn't get changed.
+    "connect_args": {"options": "-c timezone=UTC"},
```
slint, line 534: comment: I'm curious where/how/on-what these CLI-like `-c ...` option is passed 👀 

It looks like the `psql` CLI tool flag, so I guess there's some sort of API in `psycopg2`, that allows this kind of usage?

```
❯ psql --help
psql is the PostgreSQL interactive terminal.

Usage:
  psql [OPTION]... [DBNAME [USERNAME]]

General options:
  -c, --command=COMMAND    run only single command (SQL or internal) and exit
  ...
```

palkerecsenyi: This syntax is from the `libpq` C library that's used by `psycopg2`: https://www.postgresql.org/docs/18/libpq-connect.html#LIBPQ-CONNECT-OPTIONS. This is separate from the `psql` CLI.

PR #3354, opened 2026-03-05
feat(administration): add admin user role management UI

Samk13:

### Description

* add user roles modal/action in users admin views
* add user edit page role section with current roles
* wire users groups API routes and pagination in role selector
* improve role update error messages and edit-page initialization

depends on: https://github.com/inveniosoftware/invenio-users-resources/pull/197

> Note: This PR was initially generated with AI assistance and then refactored and cleaned up. The implementation has been reviewed, but additional feedback is more than welcome.

https://github.com/user-attachments/assets/0b8df6dd-1bfb-4122-a404-2b054cb8590a

### Checklist

**Frontend**

**Reminder**

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ImpersonateUser.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ImpersonateUser.js
@@ -75,7 +75,7 @@ export class ImpersonateUser extends Component {
           <Icon name="spy" />
           {i18next.t("Impersonate")}
         </Dropdown.Item>
-        <ActionModal modalOpen={modalOpen} user={user}>
+        <ActionModal modalOpen={modalOpen} resource={user}>
           {modalHeader && <Modal.Header>{modalHeader}</Modal.Header>}
           {!_isEmpty(modalBody) && modalBody}
         </ActionModal>
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRoles.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRoles.js
@@ -1,0 +1,74 @@
+/*
+ * This file is part of Invenio-App-Rdm
+ * Copyright (C) 2023-2024 CERN.
+ * Copyright (C) 2026 KTH Royal Institute of Technology.
+ *
+ * Invenio-App-Rdm is free software; you can redistribute it and/or modify it
+ * under the terms of the MIT License; see LICENSE file for more details.
+ */
+
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { Dropdown, Modal } from "semantic-ui-react";
+import { ActionModal } from "@js/invenio_administration";
+import { i18next } from "@translations/invenio_app_rdm/i18next";
+import { ManageUserRolesForm } from "./ManageUserRolesForm";
+
+const canManageRoles = (user) => {
+  return Boolean(user.links?.groups);
+};
+
+export class ManageUserRoles extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { modalOpen: false };
+  }
+
+  openModal = () => {
+    this.setState({ modalOpen: true });
+  };
+
+  closeModal = () => {
+    this.setState({ modalOpen: false });
+  };
+
+  handleSuccess = () => {
+    const { successCallback } = this.props;
+    this.closeModal();
+    successCallback();
+  };
+
+  render() {
+    const { user } = this.props;
+    const { modalOpen } = this.state;
+    if (!canManageRoles(user)) {
+      return null;
+    }
+
+    return (
+      <>
+        <Dropdown.Item
+          key="manage-user-roles"
+          onClick={this.openModal}
+          icon="id badge"
+          text={i18next.t("Manage roles")}
+        />
+        <ActionModal modalOpen={modalOpen} resource={user}>
+          <Modal.Header>{i18next.t("Manage roles")}</Modal.Header>
+          {modalOpen && (
+            <ManageUserRolesForm
+              actionSuccessCallback={this.handleSuccess}
+              actionCancelCallback={this.closeModal}
+              user={user}
+            />
+          )}
+        </ActionModal>
+      </>
+    );
+  }
+}
+
+ManageUserRoles.propTypes = {
+  user: PropTypes.object.isRequired,
+  successCallback: PropTypes.func.isRequired,
+};
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
@@ -1,0 +1,214 @@
+/*
+ * This file is part of Invenio-App-Rdm
+ * Copyright (C) 2023-2024 CERN.
+ * Copyright (C) 2026 KTH Royal Institute of Technology.
+ *
+ * Invenio-App-Rdm is free software; you can redistribute it and/or modify it
+ * under the terms of the MIT License; see LICENSE file for more details.
+ */
+
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { i18next } from "@translations/invenio_app_rdm/i18next";
+import { withCancel, ErrorMessage } from "react-invenio-forms";
+import { Form, Button, Modal, Message } from "semantic-ui-react";
+import { NotificationContext } from "@js/invenio_administration";
+import { UserModerationApi } from "../users/api";
+
+const selectionHasChanged = (initialRoleIds, selectedRoleIds) => {
+  if (initialRoleIds.length !== selectedRoleIds.length) return true;
+  const selectedRoleIdSet = new Set(selectedRoleIds);
+  return initialRoleIds.some((roleId) => !selectedRoleIdSet.has(roleId));
+};
+
+export class ManageUserRolesForm extends Component {
+  state = {
+    loading: true,
+    error: undefined,
+    showUnmanagedRoles: false,
+    confirmChanges: false,
+    initialRoleIds: [],
+    selectedRoleIds: [],
+    roles: [],
+  };
+
+  componentDidMount() {
+    this.fetchRoles();
+  }
+
+  componentWillUnmount() {
+    this.cancellableAction && this.cancellableAction.cancel();
+  }
+
+  static contextType = NotificationContext;
+
+  fetchRoles = async () => {
+    const { user } = this.props;
+
+    try {
+      const { data = {} } = await UserModerationApi.userGroups(user);
+      const assignedRoles = data.groups || [];
+      const initialRoleIds = assignedRoles.map((role) => String(role.id));
+      const roles = Array.from(
+        new Map(
+          [...assignedRoles, ...(data.available_groups || [])].map(
+            ({ id, name, is_managed: isManaged }) => [
+              String(id),
+              {
+                id: String(id),
+                name: String(name || id),
+                isManaged: Boolean(isManaged),
+              },
+            ]
+          )
+        ).values()
+      );
+
+      this.setState({
+        initialRoleIds,
+        selectedRoleIds: initialRoleIds,
+        roles,
+        loading: false,
+      });
+    } catch (error) {
+      this.setState({
+        error: error?.response?.data?.message || error?.message,
+        loading: false,
+      });
+    }
+  };
+
+  handleSubmit = async () => {
+    const { confirmChanges, initialRoleIds, selectedRoleIds } = this.state;
+    if (!selectionHasChanged(initialRoleIds, selectedRoleIds)) return;
+    if (!confirmChanges) {
+      this.setState({ confirmChanges: true });
+      return;
+    }
+
+    const { addNotification } = this.context;
+    const { actionSuccessCallback, user } = this.props;
+    this.setState({ loading: true, error: undefined });
+
+    try {
+      this.cancellableAction = withCancel(
+        UserModerationApi.setUserGroups(user, { groups: selectedRoleIds })
+      );
+      await this.cancellableAction.promise;
+
+      addNotification({
+        title: i18next.t("Success"),
+        content: i18next.t("User roles were updated."),
+        type: "success",
+      });
+      this.setState({ loading: false });
+      actionSuccessCallback();
+    } catch (error) {
+      if (error === "UNMOUNTED") return;
+      this.setState({
+        error: error?.response?.data?.message || error?.message,
+        loading: false,
+      });
+    }
+  };
+
+  render() {
+    const {
+      error,
+      loading,
+      showUnmanagedRoles,
+      confirmChanges,
+      initialRoleIds,
+      selectedRoleIds,
+      roles,
+    } = this.state;
+    const { actionCancelCallback } = this.props;
+    const selectedRoleIdSet = new Set(selectedRoleIds);
+    const roleOptions = roles
+      .filter(
+        (role) =>
+          selectedRoleIdSet.has(role.id) ||
+          (showUnmanagedRoles ? !role.isManaged : role.isManaged)
+      )
+      .map((role) => ({ key: role.id, text: role.name, value: role.id }))
+      .sort((left, right) => left.text.localeCompare(right.text));
+
+    return (
+      <>
+        <Modal.Content>
+          {error && (
+            <ErrorMessage
+              header={i18next.t("Unable to manage user roles")}
+              content={error}
+              icon="exclamation"
+              className="text-align-left"
+              negative
+            />
+          )}
+          <Message
+            info={confirmChanges}
+            warning={!confirmChanges}
+            icon={confirmChanges ? "question circle" : "warning sign"}
+            content={
+              confirmChanges
+                ? i18next.t("Are you sure you want to update this user's roles?")
+                : i18next.t("Role changes may affect the user's permissions.")
+            }
+          />
+          <Form>
+            <Form.Checkbox
+              toggle
+              checked={showUnmanagedRoles}
+              label={i18next.t("Show unmanaged roles only")}
+              onChange={(_, data) =>
+                this.setState({ showUnmanagedRoles: data.checked })
+              }
+            />
+            <Form.Dropdown
+              id="selected-role-ids"
+              multiple
+              clearable
+              fluid
+              search
+              selection
+              loading={loading}
+              options={roleOptions}
+              value={selectedRoleIds}
+              onChange={(_, data) =>
+                this.setState({
+                  selectedRoleIds: data.value || [],
+                  confirmChanges: false,
+                })
+              }
+            />
+          </Form>
+        </Modal.Content>
+        <Modal.Actions>
+          <Button onClick={actionCancelCallback} floated="left">
+            {i18next.t("Close")}
+          </Button>
+          <Button
+            size="small"
+            labelPosition="left"
+            icon="check"
+            color="green"
+            content={
+              confirmChanges
+                ? i18next.t("Confirm role update")
+                : i18next.t("Update roles")
+            }
+            onClick={this.handleSubmit}
+            loading={loading}
+            disabled={loading || !selectionHasChanged(initialRoleIds, selectedRoleIds)}
+          />
+        </Modal.Actions>
+      </>
+    );
+  }
+}
+
+ManageUserRolesForm.propTypes = {
+  user: PropTypes.object.isRequired,
+  actionCancelCallback: PropTypes.func.isRequired,
+  actionSuccessCallback: PropTypes.func.isRequired,
+};
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
@@ -27,11 +27,11 @@ export default class UserBlockForm extends Component {
     });
   }
 
-  static contextType = NotificationContext;
-
   componentWillUnmount() {
     this.cancellableAction && this.cancellableAction.cancel();
   }
+
+  static contextType = NotificationContext;
 
   handleSubmit = async (values) => {
     const { user } = this.props;
--- a/invenio_app_rdm/administration/users/users.py
+++ b/invenio_app_rdm/administration/users/users.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2023-2024 CERN.
-# Copyright (C) 2024 KTH Royal Institute of Technology.
+# Copyright (C) 2024-2026 KTH Royal Institute of Technology.
 # Copyright (C) 2024 Ubiquity Press.
 #
 # Invenio App RDM is free software; you can redistribute it and/or modify it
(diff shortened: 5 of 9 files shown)


invenio_app_rdm/administration/templates/semantic-ui/invenio_app_rdm/administration/users_edit.html
```diff
@@ -0,0 +1,48 @@
+{#
+Copyright (C) 2026 CERN.
+# Copyright (C) 2026 KTH Royal Institute of Technology.
```
max-moser, line 3:
```suggestion
  Copyright (C) 2026 CERN.
  Copyright (C) 2026 KTH Royal Institute of Technology.
```
Just for a bit more consistency with other files for now (though that should become irrelevant anyway once I finally get around to do that SPDX license info update)

invenio_app_rdm/administration/templates/semantic-ui/invenio_app_rdm/administration/users_edit.html
```diff
@@ -0,0 +1,48 @@
...
+      {{ _("Back") }}
+    </a>
+  </div>
+  <h1>{{ title or name }}</h1>
+  <div class="ui divider" aria-hidden="true"></div>
+{% endblock page_title %}
+
+{% block admin_page_content %}
+  <div class="ui segment">
+    <h3 class="ui header">{{ _("User information") }}</h3>
```
max-moser, line 24: Not a blocker; just because I couldn't quickly find a `<h2>` tag here (and that's generally sometimes the case in other parts of the InvenioRDM codebase as well):
For accessibility reasons, heading levels should not be skipped: https://www.w3.org/WAI/tutorials/page-structure/headings/

At some point in the future when we have a bit more time on our hands, we should probably make sure that this is the case.

Ping @slint @ntarocco @kpsherva

Edit: Found the InvenioRDM guidelines here too: https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/#increase-heading-levels-by-one

invenio_app_rdm/administration/templates/semantic-ui/invenio_app_rdm/administration/users_edit.html
```diff
@@ -0,0 +1,48 @@
...
+      <div class="item">
+        <div class="header">{{ _("Email") }}</div>
+        <div id="invenio-user-email" class="description"></div>
+      </div>
+    </div>
+  </div>
+  <div class="ui segment">
+    <h3 class="ui header">{{ _("Role management") }}</h3>
+    <h4 class="ui header">{{ _("Current user roles") }}</h4>
+    <div id="invenio-user-roles" class="ui bulleted list"></div>
```
max-moser, line 39: Also not a blocker, just a question: `.ui.bulleted.list` sounds a lot like `<ul>` would be a natural fit; is there a benefit of using the more generic `<div>` tag?

I just found our guidelines for this matter: https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/#use-meaningful-markup

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/index.js
```diff
@@ -9,22 +10,33 @@ import { createSearchAppInit } from "@js/invenio_search_ui";
...
 
-const defaultComponents = initDefaultSearchComponents(domContainer);
+  const defaultComponents = initDefaultSearchComponents(domContainer);
 
-const overridenComponents = {
-  ...defaultComponents,
-  "InvenioAdministration.SearchResultItem.layout": SearchResultItemLayout,
-  "SearchApp.layout": UserSearchLayout,
-  "SearchApp.facets": SearchFacets,
+  const overridenComponents = {
```
max-moser, line 21:
```suggestion
  const overriddenComponents = {
```

max-moser (commented):
I haven't had time to look through the JS code yet, here's just a few remarks so far

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
```diff
@@ -0,0 +1,295 @@
...
+        .map((roleId) => String(roleId));
+
+      const assignedById = new Map(
+        assignedRoles.filter((role) => role?.id).map((role) => [String(role.id), role])
+      );
+      const allRolesById = new Map(
+        allRoles.filter((role) => role?.id).map((role) => [String(role.id), role])
+      );
+      // Keep assigned roles searchable even if groups search is paginated/truncated.
+      assignedById.forEach((role, roleId) => {
```
max-moser: Question: From what I understand from MDN, the second parameter of the callback function in `forEach` would be the index of the element: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

It seems to me that the logic here would look up roles by their ID rather than index though?

Samk13: Good question! In this case, `assignedById` is a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map), not an array.
For Map.prototype.forEach, the callback signature is (value, key), so `roleId` here is actually the map key is the `roleID`, not an index.

Since the map was created with `[String(role.id), role]`, this iterating by role ID.
Please correct me if I am wrong.

ScreenShot:
Screenshot 2026-04-17 at 14 55 53
Note the id == name as I seeded them with a custom [script](https://github.com/Samk13/invenio-dev-latest/blob/master/scripts/dev/seed-roles.py) rather then the API endpoint

max-moser: Right!
Thanks for the clarification, I missed that difference 😅 

(My JS-fu is not that great, learned something new!)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
```diff
@@ -0,0 +1,295 @@
...
+                    search
+                    loading={loadingOptions}
+                    fieldPath="selectedOptions"
+                    options={roleOptions}
+                  />
+                  <ErrorLabel fieldPath="selectedOptions" />
+                </Form.Field>
+              </Form>
+              <Message visible>
+                <h4>{i18next.t("Current user roles")}</h4>
```
max-moser: Not a blocker: Skipped heading?

max-moser (approved):
Thanks a lot for the work on this feature!
The implementation looks good to me.

I've added a few remarks regarding our frontend guidelines with respect to accessibility, though those are not blockers.
If you'd like to address them now, that'd be great; otherwise we can deal with them later.
I think there's a *lot* of other spots in the code base that would need a cleanup regarding these things too.

Hopefully we get around to do a cleanup/QA sprint at some point 😄

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
Samk13: I appreciate that!
Please keep pointing things out, saves me from myself 😅

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/index.js
Samk13: Probably, we should keep it as is for now: 
`overridenComponents`: 
https://github.com/search?q=repo%3Ainveniosoftware%2Finvenio-app-rdm%20overridenComponents&type=code
and `overriddenComponents`:
https://github.com/search?q=repo%3Ainveniosoftware%2Finvenio-app-rdm+overriddenComponents&type=code

Maybe we can fix it in a separate PR?

invenio_app_rdm/administration/templates/semantic-ui/invenio_app_rdm/administration/users_edit.html
Samk13: Updated. While I was there, I also added whether the role is managed or not 
both icon-coded and written.
Screenshot 2026-04-17 at 17 23 01

Samk13:
@max-moser I noticed that I wasn’t filtering out the unmanaged roles in the modal.
I’ve now changed it so that unmanaged roles appear disabled, and the list of selectable roles only shows managed roles.

https://github.com/user-attachments/assets/f71697db-da69-4620-ab6f-eda9d2329348

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ImpersonateUser.js
```diff
@@ -75,9 +75,11 @@ export class ImpersonateUser extends Component {
           <Icon name="spy" />
           {i18next.t("Impersonate")}
         </Dropdown.Item>
-        <ActionModal modalOpen={modalOpen} user={user}>
-          {modalHeader && <Modal.Header>{modalHeader}</Modal.Header>}
-          {!_isEmpty(modalBody) && modalBody}
+        <ActionModal modalOpen={modalOpen} resource={user}>
```
Samk13, line 78: Not related to this feature, but this fixes the console error as there is no prop `user`in [ActionModal](https://github.com/inveniosoftware/invenio-administration/blob/58b9d35e5647fcf8ea31821d3d3e12128f36f950/invenio_administration/assets/semantic-ui/js/invenio_administration/src/actions/ActionModal.js#L22)

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/edit.js
```diff
@@ -0,0 +1,142 @@
...
+const fetchUserInfo = async (userId) => {
+  try {
+    const response = await http.get(`/api/users/${userId}`);
+    renderUserInfo(response.data);
+  } catch (error) {
+    renderUserInfo({});
+  }
+};
+
+const createRoleItem = (role) => {
```
utnapischtim: question: why do you don't use react's jsx syntax here?

utnapischtim: why is this not implemented as a react component?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/edit.js
```diff
@@ -0,0 +1,142 @@
...
+    content.appendChild(description);
+  }
+
+  item.appendChild(content);
+  return item;
+};
+
+const renderUserRoles = (roles) => {
+  const rolesEl = document.getElementById("invenio-user-roles");
+  if (!rolesEl) return;
```
utnapischtim: opinion: personally i don't like to format it like that. i don't like one-liners in that context. to me they hide something.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/edit.js
```diff
@@ -0,0 +1,142 @@
...
+    description.className = "description text-muted";
+    description.textContent = role.description;
+    content.appendChild(description);
+  }
+
+  item.appendChild(content);
+  return item;
+};
+
+const renderUserRoles = (roles) => {
```
utnapischtim: question: again, why is this not implemented as a react component?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
```diff
@@ -0,0 +1,328 @@
...
+    this.fetchRoleOptions();
+  }
+
+  componentWillUnmount() {
+    this.cancellableAction && this.cancellableAction.cancel();
+  }
+
+  static contextType = NotificationContext;
+
+  fetchAllGroups = async () => {
```
utnapischtim: question: do we have groups and roles handled in this PR? where are the groups visible on the UI, i couldn't find them.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
```diff
@@ -0,0 +1,328 @@
...
+    if (!rolesById.has(roleId)) {
+      rolesById.set(roleId, role);
+    }
+  });
+
+  return rolesById;
+};
+
+const _prepareRolesState = (assignedRoles, allRoles) => {
+  const rolesById = _mergeRolesById(assignedRoles, allRoles);
```
utnapischtim: question: do we have a redux store around for this sort of code?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/edit.js
kpsherva: same question here - it seems the code here is inconsistent with the rest of the administration panel implementation and should be standardised to follow the same pattern

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/edit.js
```diff
@@ -0,0 +1,142 @@
...
+  } catch (error) {
+    renderUserInfo({});
+  }
+};
+
+const createRoleItem = (role) => {
+  const item = document.createElement("li");
+  item.className = "flex align-items-center rel-mb-1";
+
+  const icon = document.createElement("i");
```
kpsherva: I am struggling in terms of readability of the code when the classes are assigned to elements

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
Samk13: I checked the code around, and there's no Redux used for such modal/action flows. Similar components keep form state locally and use callbacks on success.

I refactored the logic to make this clearer. Adding Redux here would introduce unnecessary complexity without shared consumers.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/edit.js
Samk13: Thanks, agreed. I refactored the code to use React components/JSX instead of manually creating DOM elements and assigning classes imperatively.

The role list is now rendered through `UserRoleItem` / `UserRolesList`, so the structure and class names are visible in JSX and should be easier to read and maintain.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
Samk13: Yes, they are the same thing. The API calls them `groups`, while the UI calls them `roles` because that is what admins manage. We are using these 2 terms interchangeably.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/edit.js
Samk13: Thanks for pointing this out. I did a full rewrite now.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/edit.js
Samk13: I did a full rewrite of the code now. Please let me know how we could improve it further.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ImpersonateUser.js
```diff
@@ -75,9 +75,11 @@ export class ImpersonateUser extends Component {
...
+          <>
```
utnapischtim: question: why are the <></> necessary here?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRoles.js
```diff
@@ -0,0 +1,97 @@
...
+            labelPosition="left"
+            primary
+            size="small"
+          >
+            <Icon name="id badge" />
+            {i18next.t("Manage roles")}
+          </Button>
+        )}
+        <ActionModal modalOpen={modalOpen} resource={user}>
+          <>
```
utnapischtim: nit: this <></> is also not necessary or?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRoles.js
```diff
@@ -0,0 +1,97 @@
...
+    successCallback();
+  };
+
+  render() {
+    const { user, asDropdownItem } = this.props;
+    const { modalOpen } = this.state;
+    return (
+      <>
+        {asDropdownItem ? (
+          <Dropdown.Item
```
utnapischtim: nit: i understand why it is sometimes a `Dropdown.Item` and sometimes a `Button`, but this structure looks strange. is this a pattern we use on other places in inveniordm too?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
```diff
@@ -0,0 +1,251 @@
...
+import { Form, Button, Modal, Message, Icon } from "semantic-ui-react";
+import { NotificationContext } from "@js/invenio_administration";
+import * as Yup from "yup";
+import {
+  emptyUserRolesState,
+  fetchUserRoleManagementState,
+  UserModerationApi,
+} from "../users/api";
+
+const _extractBackendError = (error) => {
```
utnapischtim: nit: this function looks unnecessary complicated. can the whole error handling be simplified? are there precedence's in InvenioRDM where it is handled like that? Normally i would say the frontend is the presentation layer and should therefore not need to parse or change something to show it.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
```diff
@@ -0,0 +1,251 @@
+/*
```
utnapischtim, line 1: minor: this component looks like as it mixes logic and presentation. so the fetch part should go into a store.at some point we have the `../users/api` but some of the fetch logic is implemented in this component. in my opinion the whole fetch logic should go into a store. if we don't have yet one, we should set up one for the administration page

Samk13: Agree with the direction, but I would keep it out of this PR. There does not seem to be an existing administration store pattern to plug into, and the API/data aggregation part is already in `users/api/roles.js` the component mostly keeps modal-local state such as loading/error and selected values. 
Setting up an administration store would be a broader refactor affecting more admin actions/forms, so I think it is better tracked as a follow-up instead of introducing that architecture here.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
```diff
@@ -0,0 +1,235 @@
...
+    actionCloseCallback();
+  };
+
+  render() {
+    const { error, loading, loadingRoles, rolesState } = this.state;
+    const { assignedRoles, initialManagedRoleIds, managedRoles } = rolesState;
+    const roleOptions = _toSortedRoleOptions(managedRoles);
+
+    return (
+      <Formik
```
kpsherva: was there a blocker in using the automatically generated form administration class view? Does it need to be reimplemented - I am asking because it brings potentially more maintenance effort for the future

this also applies to the actions available to be done in the admin - it might be a bit underdeveloped still for some use cases, but I think it is worth exploring if we can shave off the maintenance of javascript code

Samk13: I checked this again. The generated user edit form currently only covers normal editable user fields like username/email. Role assignment is handled through separate /users/<id>/groups endpoints and needs data from both the assigned roles and the available roles list. 
We also need to expose only managed roles for editing while preserving unmanaged assigned roles on submit.

So I don't think the generated admin edit form fits this case without adding custom fetch/filter/submit logic anyway. 
I agree with the maintenance concern, but this seems closer to a custom relationship action than a standard resource edit form.
If you have a better approach, please share, and I will fix it.

Samk13:
Following the Telecon discussion on 2026/05/05, we agreed on: 
The admin UI should list both managed and unmanaged roles and allow admins to assign or remove them from users where permitted.

This supports institutions syncing roles from identity providers or pre-creating roles in Invenio. 
Long-term, we may introduce more granular permissions so that only specific admins can manage unmanaged roles.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/components/ManageUserRolesForm.js
kpsherva: the admin actions and views can be configured to handle separate api endpoints, including the marshmallow schema which is controlling the input on the backend (can be assigned to the action).
The managed/unmanaged roles assignment should be also handled by the backend, otherwise you can bypass the UI "red tape" by simply using REST API directly. I think it would be worth trying to use the actions/views of admin panel implementation, since it was done for cases like this, to avoid maintenance cost of custom JS

Samk13: Thanks for the feedback. We simplified the PR based on the review and telecon.

 The backend remains the source of truth for permissions and managed/unmanaged role enforcement. The UI filter is only a convenience, and the action is shown only when the backend exposes the required links.

 We removed the custom edit/details page approach. Role management is now only a users-list dropdown action.

 A small custom modal body remains because the form needs to fetch current assignments and available roles from the backend, then render them as a searchable multi-select submitting groups: List(Str). The generic ActionForm cannot currently provide that dynamic role-selector
 UX.

Here is how it looks now after removing the custom edit user page:

https://github.com/user-attachments/assets/ae11ea25-bf2a-46d8-8368-75f7d99c4720

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
```diff
@@ -27,12 +27,12 @@ export default class UserBlockForm extends Component {
...
   }
 
-  static contextType = NotificationContext;
-
   componentWillUnmount() {
     this.cancellableAction && this.cancellableAction.cancel();
   }
 
+  static contextType = NotificationContext;
+
```
Samk13, line 35: fix linting warning

PR #3368, opened 2026-03-13
admin: deleted records link with include_deleted parameter

ptamarit:
When reviewing deleted records, administrators usually want to see the deleted records content and not the tombstones.
It probably does not make sense to expose the "include_deleted" links in the API response, so we built the URL client-side (only needed for the admin panel).

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
@@ -44,6 +44,13 @@ class SearchResultItemComponent extends Component {
     // See  https://github.com/inveniosoftware/invenio-app-rdm/issues/2849
     const recordOwner = result?.parent?.access?.owned_by;
 
+    let selfLink = result.links.self_html;
+    if (result?.deletion_status?.is_deleted) {
+      const url = new URL(selfLink);
+      url.searchParams.set("include_deleted", "1");
+      selfLink = url.toString();
+    }
+
     return (
       <Table.Row>
         <Table.Cell
@@ -64,7 +71,7 @@ class SearchResultItemComponent extends Component {
             color="red"
             icon="lock"
           />
-          <a target="_blank" rel="noreferrer noopener" href={result.links.self_html}>
+          <a target="_blank" rel="noreferrer noopener" href={selfLink}>
             {_truncate(result.metadata.title || i18next.t("Empty draft title"), {
               length: 100,
             })}


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
```diff
@@ -44,6 +44,11 @@ class SearchResultItemComponent extends Component {
     // See  https://github.com/inveniosoftware/invenio-app-rdm/issues/2849
     const recordOwner = result?.parent?.access?.owned_by;
 
+    let selfLink = result.links.self_html;
+    if (result?.deletion_status?.is_deleted) {
+      selfLink += "?include_deleted=1";
```
slint: nit: could we "formally" build the URL, by parsing and setting the param? This is more of a future-proofing thing, in case at some point we already bake into the URL a different querystring param

slint (approved):
LGTM, just a tiny nit, not critical for merging.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
ptamarit: Good point, done. Now using `new URL` and `searchParams.set("include_deleted", "1")`.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/records/search/SearchResultItemLayout.js
```diff
@@ -44,6 +44,13 @@ class SearchResultItemComponent extends Component {
     // See  https://github.com/inveniosoftware/invenio-app-rdm/issues/2849
     const recordOwner = result?.parent?.access?.owned_by;
 
+    let selfLink = result.links.self_html;
+    if (result?.deletion_status?.is_deleted) {
```
utnapischtim, line 48: question: why do we do that in the component and not in the backend?

ptamarit: I don't think it makes sense to expose the "include_deleted" links in the API response, since only admins have access to it, and it's only needed for the admin panel.

utnapischtim: my concerns are that we mix responsibilities and that could introduce problems in the future. as of now i thought the only place where links are created is in the links section in the service layer.

slint: We would need a conditional `self_html_deleted` link that also takes into account the user's identity... Which I think we have access to from `ctx["identity"]`. It could work tbh, and it would be more backend-driven indeed.

Though another PR in `invenio-rdm-records` 😬

PR #3375, opened 2026-03-20
refactor(views): Move get_record_requests to RecordCommunitiesService

### LINKED ISSUE #568: "View comments" button under the community is not visible on other record versions (landing page)
Currently, if a request is accepted for a record version, the "View Comments" button under the accepted community only shows up for that specific version. It is stored in the "topic" sub-field of the request's payload.

There should be a way to view the request comments from the landing page of all versions.

This will require to have the parent_id on the requests index, to be able to show the button on all other versions.

### Solution proposal (to be discussed and finalized)
The link can redirect to the dashboard with the results containing all the related requests on the parent id.
This way if there is scenario where a community inclusion request of version was not accepted but the version 2 was accepted, the link will redirect to the requests dashboard containing all requests for those records in that community to be able to refer to the older request for reference.

#### Steps
1. Add parent id to the request index and update index (all new requests will contain parent_id)
2. Re-index the requests (where parent_id doesn't exist, all requests before we updated the index)
3. Update the link in the UI at the end (this way we avoid any searches on the parent till we have it indexed on all)

## Additional context
Another thing to consider is that on the request dashboard currently there is no way to view all open/closed/owned/shared request in a single list view. We can have a way to be able to view.

sakshamarora1:
fixes: https://github.com/inveniosoftware/invenio-requests/issues/568
requires: https://github.com/inveniosoftware/invenio-communities/pull/1361 and https://github.com/inveniosoftware/invenio-rdm-records/pull/2294

### DIFF
--- a/invenio_app_rdm/records_ui/views/records.py
+++ b/invenio_app_rdm/records_ui/views/records.py
@@ -107,43 +107,6 @@ def get_record_community(record):
         return None, None
 
 
-def get_record_requests(record, identity):
-    """Return all requests that concern this record.
-
-    Output: {<Community-UUID>: <Request-UUID>}
-    """
-    can_review = current_rdm_records.records_service.check_permission(
-        identity, "review", record=record._record
-    )
-    if not can_review:
-        return {}
-
-    if type(identity) is AnonymousIdentity:
-        return {}  # secret link users do not have permissions to search requests
-
-    record_requests = current_requests_service.search(
-        identity,
-        extra_filter=dsl.Q(
-            "bool",
-            must=[
-                dsl.Q("term", **{"topic.record": record["id"]}),
-                dsl.Q(
-                    "terms",
-                    **{
-                        "type": [
-                            CommunityInclusion.type_id,
-                            CommunitySubmission.type_id,
-                        ]
-                    },
-                ),
-            ],
-        ),
-        params={"sort": "oldest"},
-    )
-
-    return {r["receiver"]["community"]: r["id"] for r in record_requests}
-
-
 class PreviewFile:
     """Preview file implementation for InvenioRDM.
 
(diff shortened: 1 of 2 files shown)


carlinmack (approved):

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -119,6 +120,17 @@ def get_record_requests(record, identity):
     if type(identity) is AnonymousIdentity:
         return {}  # secret link users do not have permissions to search requests
 
+    # Get all accepted requests that led to the record being added to the community
+    parent = record._record.parent
```
kpsherva: this bypasses the service layer. Is there any way to get the parent of the record without accessing `_record`?

sakshamarora1: Yeah, I thought before about moving this whole function to the service class [here in rdm-records](https://github.com/inveniosoftware/invenio-rdm-records/blob/997d748788d21979f32d8c83b772b52d74c22736/invenio_rdm_records/services/communities/service.py#L48) as a service method.

Because now, we are accessing the DB here as well.
Updated the PR and moved it to the service class: https://github.com/inveniosoftware/invenio-rdm-records/pull/2294

OliverGeneser (approved):

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -107,43 +107,6 @@ def get_record_community(record):
...
-    """Return all requests that concern this record.
-
-    Output: {<Community-UUID>: <Request-UUID>}
-    """
-    can_review = current_rdm_records.records_service.check_permission(
-        identity, "review", record=record._record
-    )
-    if not can_review:
-        return {}
-
```
fenekku, line 120: It is because this logic was not preserved in the move to invenio-rdm-records https://github.com/inveniosoftware/invenio-rdm-records/pull/2294/changes#diff-77774ee36bf43d671108f634d89205e2f4bfe634703deaaf95d6a28f9b8ab726R507 that the (signposting) tests for this PR are failing. New `get_record_requests` requires a "review" permission and is used in the view for a record's detail page. But a record's detail page can be seen by anonymous users which don't have this permission. The signposting test flags this because an anonymous HEAD/GET calls are made to the record landing page.

Reverting to using this conditional seems the most straightforward and inline with other similar cases in the code base.

fenekku (commented):
I think a fix in invenio-rdm-records is needed for `get_record_requests`. see below.

PR #3106, opened 2025-07-11
semantic-ui: componenets: add url fetching in copy button

sakshamarora1:
fixes: https://github.com/CERNDocumentServer/cds-rdm/issues/291

Screenshot 2025-07-11 at 17 15 11

Screenshot 2025-07-14 at 17 26 58

PR #3379, opened 2026-03-21
fix(components): use ClipboardItem API for Safari-compatible clipboard writes

### LINKED ISSUE #1266: Export - Copy to clipboard doesn't work in Safari
Using the `Copy to clipboard` button in the Export section of an item doesn't work in Safari. The buttons for the DOI and Citation sections do work, just not the Export section.

Clicking the button in the Export section gives the following error in the Javascript console with every format I try.

```
Unhandled Promise Rejection: NotAllowedError: The request is not allowed by the user agent 
or the platform in the current context, possibly because the user denied permission.
```

It sounds like it might be trying to write to the clipboard during an async call. Safari doesn't permit clipboard writes without user interaction, so the async calls have to be handled correctly (see for example, https://wolfgangrittner.dev/how-to-use-clipboard-api-in-safari/).

I'm not sure, but I think this PR is related: https://github.com/inveniosoftware/invenio-app-rdm/pull/3106

djchrisssssss:
## Summary

- Fix `CopyButton` export copy failing in Safari with `NotAllowedError`
- Replace `async/await` clipboard pattern with synchronous `ClipboardItem` + Promise API
- Add fallback to `writeText()` for browsers without `ClipboardItem` support

## Problem

PR #3106 replaced `react-copy-to-clipboard` with direct `navigator.clipboard.writeText()` calls. When the `url` prop is provided (Export section), `await fetch(url)` suspends execution and breaks the user-gesture call stack. Safari's Clipboard API requires `clipboard.write()` to be called **synchronously** within the user gesture context, so it rejects the write with:

```
NotAllowedError: The request is not allowed by the user agent or the platform
in the current context, possibly because the user denied permission.
```

DOI/Citation copy buttons are unaffected because they pass `text` directly (no fetch needed).

## Solution

Use the `ClipboardItem` API which accepts a **Promise** as its value. This keeps `navigator.clipboard.write()` synchronous within the user gesture, while the Clipboard API internally resolves the fetch Promise:

```javascript
const item = new ClipboardItem({
  "text/plain": fetch(url)
    .then((r) => r.text())
    .then((t) => new Blob([t], { type: "text/plain" })),
});
navigator.clipboard.write([item]);
```

A fallback using `writeText()` is included for browsers where `ClipboardItem` is not available (e.g., older Firefox).

Reference: [How to use Clipboard API in Safari](https://wolfgangrittner.dev/how-to-use-clipboard-api-in-safari/)

## Browser Compatibility

| Browser | `ClipboardItem` with Promise | Previous `async/await` pattern |
|---------|:---:|:---:|
| Safari 13.1+ | ✅ | ❌ |
| Chrome 76+ | ✅ | ✅ |
| Firefox 127+ | ✅ (fallback provided) | ✅ |

## Test Plan

- [ ] Verify Export "Copy to clipboard" works in Safari
- [ ] Verify Export "Copy to clipboard" still works in Chrome
- [ ] Verify Export "Copy to clipboard" still works in Firefox
- [ ] Verify DOI copy button still works in all browsers
- [ ] Verify Citation copy button still works in all browsers

Closes #3378
Refs zenodo/zenodo-rdm#1266

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/CopyButton.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/CopyButton.js
@@ -10,19 +10,33 @@ import { Button, Popup } from "semantic-ui-react";
 import { i18next } from "@translations/invenio_app_rdm/i18next";
 
 class SimpleCopyButton extends React.Component {
-  fetchUrl = async (url) => {
-    return await (await fetch(url)).text();
-  };
-
-  handleClick = async () => {
+  handleClick = () => {
     const { url, text, onCopy } = this.props;
-    let textToCopy = text;
-    if (url) {
-      textToCopy = await this.fetchUrl(url);
-    }
 
-    await navigator.clipboard.writeText(textToCopy);
-    onCopy(text);
+    if (url) {
+      // When a URL is provided, the content must be fetched before copying.
+      // Safari requires clipboard writes to occur synchronously within the
+      // user-gesture call stack. Using async/await with fetch() breaks this
+      // chain. The ClipboardItem API accepts a Promise, keeping the
+      // clipboard.write() call synchronous while resolving data internally.
+      const textPromise = fetch(url).then((response) => response.text());
+
+      if (typeof ClipboardItem !== "undefined") {
+        const item = new ClipboardItem({
+          "text/plain": textPromise.then(
+            (fetchedText) => new Blob([fetchedText], { type: "text/plain" })
+          ),
+        });
+        navigator.clipboard.write([item]).then(() => onCopy(text));
+      } else {
+        // Fallback for browsers where ClipboardItem is not available
+        textPromise.then((fetchedText) =>
+          navigator.clipboard.writeText(fetchedText).then(() => onCopy(text))
+        );
+      }
+    } else {
+      navigator.clipboard.writeText(text).then(() => onCopy(text));
+    }
   };
 
   render() {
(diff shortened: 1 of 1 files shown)


Samk13 (commented):
Thanks for this, the Safari issue and the use of ClipboardItem look correct to me.

One suggestion: you can simplify the control flow slightly by resolving the source upfront, which removes the duplication between the two clipboard paths:

```js
handleClick = () => {
  const { url, text, onCopy } = this.props;

  // Resolve what to copy (URL fetch or direct text)
  const dataPromise = url
    ? fetch(url).then((r) => r.text())
    : Promise.resolve(text);

  if (typeof ClipboardItem !== "undefined") {
    // Safari-safe: keep clipboard.write() synchronous
    const item = new ClipboardItem({
      "text/plain": dataPromise.then(
        (t) => new Blob([t], { type: "text/plain" })
      ),
    });

    navigator.clipboard.write([item]).then(() => onCopy(text));
  } else {
    // Fallback for browsers without ClipboardItem
    dataPromise.then((t) =>
      navigator.clipboard.writeText(t).then(() => onCopy(text))
    );
  }
};
```
WDYT?

djchrisssssss (commented):
Thanks @Samk13! Great suggestion — adopted the unified `dataPromise` approach in the follow-up commit. Changes:

1. **Unified `dataPromise`**: Merged the `url` fetch and direct `text` paths into a single Promise, removing the outer `if/else` duplication — exactly as you suggested.
2. **Removed dead `fetchUrl()` method**: It was no longer referenced after the initial commit.
3. **Prettier fix**: Collapsed the `ClipboardItem` callback to satisfy the project's prettier config.
4. **Linter pass**: Ran `run-js-linter.sh` — CopyButton.js has zero errors/warnings.

Note: I used a follow-up commit (rather than amend) so the incremental diff is easier to review. Happy to squash before merge if preferred.

PR #3421, opened 2026-04-13
feat: use new RDMSubCommunity request class

OliverGeneser:


slint (approved):
~~See https://github.com/inveniosoftware/invenio-rdm-records/pull/2314#issuecomment-4235634747~~ all good 👌

PR #3423, opened 2026-04-13
Integrate Flask-MultiProfiler

max-moser:
Integrate the new `flask-multiprofiler` as an optional dependency via the new `profiler` extra (`pip install invenio-app-rdm[profiler]`).

Video:

https://github.com/user-attachments/assets/b2ee88ea-a23a-4d00-a4fd-15fd82fcb755

### DIFF
--- a/invenio_app_rdm/ext.py
+++ b/invenio_app_rdm/ext.py
@@ -173,3 +173,17 @@ def init_menu(app):
         },
         **dict(icon="upload", permissions="can_read"),
     )
+
+    # register a menu entry for the profiler, if it is installed
+    if "multiprofiler" in app.extensions:
+        permission_check_func = app.config["MULTIPROFILER_PERMISSION"]
+        current_menu.submenu("profile-admin.profiler").register(
+            "profiler.index",
+            text=_(
+                "%(icon)s Profiler",
+                icon='<i class="bar chart icon"></i>',
+            ),
+            order=10,
+            visible_when=permission_check_func,
+            permissions="can_read",
+        )
--- a/invenio_app_rdm/ext.py
+++ b/invenio_app_rdm/ext.py
@@ -16,6 +16,24 @@ from invenio_i18n import lazy_gettext as _
 
 from .communities_ui.views.ui import _show_browse_page
 
+try:
+    from flask_multiprofiler import MultiProfiler
+
+except (ImportError, ModuleNotFoundError):
+
+    class MultiProfiler:
+        """Dummy profiler for entrypoints if Flask-MultiProfiler isn't installed.
+
+        Note that this dummy extension does not register itself to ``app.extensions``.
+        """
+
+        def __init__(self, app=None):
+            """Constructor."""
+
+        def init_app(self, app):
+            """Initialize application."""
+            self.app = app
+
 
 def _is_branded_community():
     """Function used to check if community is branded."""
--- a/invenio_app_rdm/ext.py
+++ b/invenio_app_rdm/ext.py
@@ -12,6 +12,7 @@ from datetime import timedelta
 
 from flask import request
 from flask_menu import current_menu
+from invenio_app import talisman
 from invenio_i18n import lazy_gettext as _
 
 from .communities_ui.views.ui import _show_browse_page
@@ -47,6 +48,15 @@ def finalize_app(app):
     """Finalize app."""
     init_menu(app)
     init_config(app)
+
+    # make CSP headers for profiler reports very lax
+    if "multiprofiler" in app.extensions:
+        orig_profiler_report_view = app.view_functions["profiler.report_view"]
+
+        @app.endpoint("profiler.report_view")
+        @talisman(content_security_policy={})
+        def wrapped_profiler_report_view(*args, **kwargs):
+            return orig_profiler_report_view(*args, **kwargs)
 
 
 def init_config(app):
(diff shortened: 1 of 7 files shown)


invenio_app_rdm/ext.py
```diff
@@ -30,6 +49,14 @@ def finalize_app(app):
     init_menu(app)
     init_config(app)
 
+    # make CSP headers for profiler reports very lax
```
max-moser, line 52: I've noticed that the default CSP settings prevent the result pages from being rendered properly, so I've loosened the CSP headers here.
Because there the output doesn't depend on arbitrary user input, I'm not terribly worried about that.

invenio_app_rdm/theme/templates/semantic-ui/flask_multiprofiler/index.html
```diff
@@ -0,0 +1,24 @@
...
+{%- extends config.BASE_TEMPLATE %}
+
+{%- block head_title -%}
+  <title>{{ _("Profiler") }}</title>
+{%- endblock -%}
+
+{%- block javascript -%}
+  {{ super() }}
+  {{ webpack["flask-multiprofiler-code.js"] }}
+{%- endblock %}
```
max-moser, line 19: Having the JS in its separately served file is necessary due to the default CSP settings.
It should be noted that since this template here is configured as the _parent_ template for the `profiler/index.html` in `flask-multiprofiler`, we cannot prevent the latter from still rendering its inline code snippet that will be blocked by the CSP settings.
That doesn't really hurt, but prints yet another warning in the console...

Perhaps we could go the other way and adjust the template mechanism in `flask-multiprofiler` to have the actual template being rendered configurable, rather than the parent template (which would then come for free anyway)?

invenio_app_rdm/ext.py
```diff
@@ -30,6 +49,14 @@ def finalize_app(app):
...
+    if "multiprofiler" in app.extensions:
+        talisman = app.extensions["invenio-app"].talisman
```
slint: minor: nice, I forgot we had `finalize_app` now so we're able to access Talisman.

My original issue on some better ergonomics on this will turn 8 years old in 2 weeks 👶📈: https://github.com/inveniosoftware/invenio-app/issues/22

invenio_app_rdm/ext.py
```diff
@@ -12,10 +12,29 @@
 
 from flask import request
 from flask_menu import current_menu
+from invenio_administration.permissions import administration_permission
```
utnapischtim: nit: did you forget to remove this import?

max-moser: Good catch; I've moved the config from profiler extension initialization to `config.py`; that was still a leftover from then.

invenio_app_rdm/ext.py
max-moser: I've heard that after 8 years, open GitHub issues gain consciousness!
Since in this day and age, there's no such thing as being too cautious, I've created the following PR to address that (and hopefully close the issue in time): https://github.com/inveniosoftware/invenio-app/pull/109

invenio_app_rdm/ext.py
```diff
@@ -30,6 +49,15 @@ def finalize_app(app):
     init_menu(app)
     init_config(app)
 
+    # make CSP headers for profiler reports very lax
+    if "multiprofiler" in app.extensions:
+        orig_profiler_report_view = app.view_functions["profiler.report_view"]
+
+        @app.endpoint("profiler.report_view")
+        @talisman(content_security_policy={})
```
max-moser, line 57: This now requires https://github.com/inveniosoftware/invenio-app/pull/109 
Making it work without that PR only requires removing the import, and setting `talisman = app.extensions["invenio-app"].talisman`

invenio_app_rdm/ext.py
```diff
@@ -155,3 +183,17 @@ def init_menu(app):
...
+        permission_check_func = app.config["MULTIPROFILER_PERMISSION"]
+        current_menu.submenu("profile-admin.profiler").register(
+            "profiler.index",
+            text=_(
+                "%(icon)s Profiler",
+                icon='<i class="bar chart icon"></i>',
+            ),
+            order=10,
+            visible_when=permission_check_func,
+            permissions="can_read",
```
slint, line 198:
```suggestion
```

I guess we don't need this one, since we're not making any service-based call to calculate permissions in the frontend (like e.g. for communities menus).

invenio_app_rdm/theme/templates/semantic-ui/flask_multiprofiler/index.html
slint: Agree, I think initially I just wanted to quickly add the instance "chrome", but we can make it so that we just build the final template here and just have "partials" in Flask-MultiProfiler.

As you said, not critical for merging this.

invenio_app_rdm/ext.py
```diff
@@ -12,10 +12,29 @@
...
 from invenio_i18n import lazy_gettext as _
 
 from .communities_ui.views.ui import _show_browse_page
 
+try:
+    from flask_multiprofiler import MultiProfiler
+
+except (ImportError, ModuleNotFoundError):
+
+    class MultiProfiler:
```
slint, line 25: Woah, this big-brain time! I was about to write a comment that we should probably add a config flag to allow disabling the profiler, but just saw it's an extra.

slint (approved):
LGTM! Python 3.9 test failed because it didn't pick up the new invenio-app v3.1.0 release with the top-level `talisman` export.

Otherwise good to merge, and I guess we can also close https://github.com/inveniosoftware/invenio-app-rdm/pull/3199? (ping @sakshamarora1)

PR #3426, opened 2026-04-16
feat(administration): allow blocking with removal reason

slint:
> AI use disclaimer: All of the UI parts were done with Claude Code, mostly focusing on UX while testing.

* Depends on:
  * https://github.com/inveniosoftware/invenio-users-resources/pull/199
  * https://github.com/inveniosoftware/invenio-communities/pull/1374
  * https://github.com/inveniosoftware/invenio-rdm-records/pull/2326
* Introduces a modal when blocking a user (via the Users administration
  or manage dropdown on the record landing page), which previews the
  user to be deleted and allows for choosing a removal reason and
  public note.

### Block modal via record manage dropdown (to showcase "view in admin" button)

https://github.com/user-attachments/assets/e14f1860-23e3-4ea4-b246-bf52957566bc

### DIFF
--- a/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
+++ b/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
@@ -1,0 +1,250 @@
+/*
+ * This file is part of Invenio.
+ * Copyright (C) 2026 CERN.
+ *
+ * Invenio is free software; you can redistribute it and/or modify it
+ * under the terms of the MIT License; see LICENSE file for more details.
+ */
+
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { Formik } from "formik";
+import { ErrorMessage, Image, TextAreaField, withCancel } from "react-invenio-forms";
+import { Button, Form, Header, Icon, Message, Modal, Segment } from "semantic-ui-react";
+import { NotificationContext } from "@js/invenio_administration";
+import { i18next } from "@translations/invenio_app_rdm/i18next";
+import * as Yup from "yup";
+import RemovalReasonsSelect from "../records/RemovalReasonsSelect";
+import { UserModerationApi } from "./api";
+
+export default class UserBlockForm extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { loading: false, error: undefined, blocked: false };
+    this.schema = Yup.object({
+      removal_reason: Yup.string().required(i18next.t("A removal reason is required.")),
+      note: Yup.string(),
+    });
+  }
+
+  static contextType = NotificationContext;
+
+  componentWillUnmount() {
+    this.cancellableAction && this.cancellableAction.cancel();
+  }
+
+  handleSubmit = async (values) => {
+    const { user } = this.props;
+    const { addNotification } = this.context;
+    const name = user.profile?.full_name || user.email || user.username || user.id;
+
+    this.setState({ loading: true });
+
+    const payload = {
+      removal_reason_id: values.removal_reason,
+      ...(values.note ? { note: values.note } : {}),
+    };
+
+    this.cancellableAction = withCancel(UserModerationApi.blockUser(user, payload));
+
+    try {
+      await this.cancellableAction.promise;
+      this.cancellableAction = null;
+      this.setState({ loading: false, error: undefined, blocked: true });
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
```diff
@@ -0,0 +1,251 @@
...
+import RemovalReasonsSelect from "../records/RemovalReasonsSelect";
+import { UserModerationApi } from "./api";
+
+export default class UserBlockForm extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { loading: false, error: undefined, blocked: false };
+    this.schema = Yup.object({
+      removal_reason: Yup.string().required(i18next.t("A removal reason is required.")),
+      note: Yup.string(),
```
OliverGeneser, line 26: Should we added a max char size?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
```diff
@@ -0,0 +1,251 @@
...
+      });
+    } catch (error) {
+      if (error === "UNMOUNTED") return;
+      this.setState({
+        loading: false,
+        error: error?.response?.data?.message || error?.message,
+      });
+      console.error(error);
+    }
+  };
```
OliverGeneser, line 70: nit; to keep the lifecycle clean, I think we should clear cancellableAction after it succeeded

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
```diff
@@ -0,0 +1,251 @@
...
+                    fluid
+                  />
+                </Form.Field>
+              </Form>
+            </Modal.Content>
+            <Modal.Actions>
+              <Button onClick={actionCancelCallback} floated="left">
+                {i18next.t("Cancel")}
+              </Button>
+              <Button
```
OliverGeneser, line 229: Why not type="submit"?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
```diff
@@ -0,0 +1,251 @@
...
+                icon="ban"
+                labelPosition="left"
+                color="red"
+                content={i18next.t("Block user")}
+                onClick={(event) => handleSubmit(event)}
+                loading={loading}
+                disabled={loading}
```
OliverGeneser: Shouldn't it also be disabled if no removal_reason has been selected?

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
slint: I see that as usual [we don't have max length validation in the backend](https://github.com/inveniosoftware/invenio-rdm-records/blob/e17ad9826330234ceface80a3f7a7b3bfbc5813e/invenio_rdm_records/services/schemas/tombstone.py#L41)... Given that this is a backoffice/admin form and gets indexed as text anyways in search, I don't want to introduce the length constraint right now (but instead do a proper pass at some point and release a major since it's a breaking change).

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
slint: Why indeed! Fixed, I had to wrap the form in `<Form>` though for it to work.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/users/UserBlockForm.js
slint: Done by checking for `disabled={loading || !isValid}` from the Formik props. In practice the `RemovalReasonSelect` though automatically picks a value, so it's not possible to pass none (unless I think we fail to remotely load all the reasons...)

PR #3428, opened 2026-04-17
fix: display user dashboard header for community topic requests as before

fenekku:
Problem:

image

Fix:

image


slint (approved):

slint:
Ping @OliverGeneser, I remember it was weird that we're not passing this to all rendered user dashboard templates.

In general, requests UI rendering logic needs a bit of a refactor...

PR #3429, opened 2026-04-20
users: fix hardcoded deposit URL in uploads dashboard

### LINKED ISSUE #3340: User Dashboard: Uploads RDMEmptyResults uses hardcoded URL
On the user's dashboard uploads search app, we have a hardcoded URL on the "New Upload" button:

https://github.com/inveniosoftware/invenio-app-rdm/blob/e5afec45c5160d44eb761c695d4b4f939d2d9d4f/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads.js#L149

This breaks the "New Upload" button when `APP_RDM_ROUTES` is configured to use a different `deposit_create` route.

Needs to be fixed to use the same deposit_create link as in the dashboard header template:

https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html#L26

sushmithainjeti:
* FIX Replaces hardcoded '/uploads/new' URL with dynamic URL from APP_RDM_ROUTES config.

## What does this PR fix?

Fixes #3340

## Changes

The "New Upload" button in the user dashboard had a hardcoded URL '/uploads/new' in two places in `uploads.js`. This breaks the button when `APP_RDM_ROUTES` is configured to use a different 'deposit_create' route.

## How was it fixed?

- Added 'data-deposit-url' attribute to the template 'uploads.html' using 'url_for('invenio_app_rdm_records.deposit_create')' to dynamically resolve the URL
- Updated 'uploads.js' to read the URL from the DOM instead of using the hardcoded value.

### DIFF
--- a/invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/uploads.html
+++ b/invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/uploads.html
@@ -19,7 +19,7 @@
   <div
     id="invenio-search-config"
     data-invenio-search-config='{{ search_app_rdm_user_uploads_config(app_id="InvenioAppRdm.DashboardUploads") | tojson }}'
-    data-deposit-url="{{ url_for('invenio_app_rdm_records.deposit_create') }}"
+    data-deposit-url="{{ invenio_url_for('invenio_app_rdm_records.deposit_create') }}"
   >
 
 </div>
--- a/invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/uploads.html
+++ b/invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/uploads.html
@@ -19,6 +19,7 @@
   <div
     id="invenio-search-config"
     data-invenio-search-config='{{ search_app_rdm_user_uploads_config(app_id="InvenioAppRdm.DashboardUploads") | tojson }}'
+    data-deposit-url="{{ url_for('invenio_app_rdm_records.deposit_create') }}"
   >
 
 </div>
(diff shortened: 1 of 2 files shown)


sushmithainjeti:
I have tested the fix locally. The 'New Upload' button now correctly navigates to the upload page using the dynamic URL from APP_RDM_ROUTES config.
test-screenshot

invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/uploads.html
```diff
@@ -19,6 +19,7 @@
   <div
     id="invenio-search-config"
     data-invenio-search-config='{{ search_app_rdm_user_uploads_config(app_id="InvenioAppRdm.DashboardUploads") | tojson }}'
+    data-deposit-url="{{ url_for('invenio_app_rdm_records.deposit_create') }}"
```
tmorrell: Could you use invenio_url_for? That does a better job of managing API/UI urls https://inveniordm.docs.cern.ch/releases/v13/version-v13.0.0/#miscellaneous-additions

tmorrell (changes requested):
Thanks for the fix. It needs at least one change to be merged.

invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/uploads.html
sushmithainjeti: Updated to  invenio_url_for as suggested.

tmorrell (approved):
Looks good to me. This will need one more approval to merge.

sushmithainjeti:
Thanks for the review! I’ll wait for the second approval. please let me know if anything else is needed from my side.

wgresshoff (approved):
This is exactly what our frontend dev suggested!

sushmithainjeti:
> This is exactly what our frontend dev suggested!

Great to hear. that’s reassuring. Thanks!

PR #3434, opened 2026-04-24
fix(tombstone): return not-found page + log exception for PIDDeletedError

palkerecsenyi:


invenio_app_rdm/records_ui/views/records.py
```diff
@@ -510,7 +509,13 @@ def record_tombstone_error(error):
         record_ui = UIJSONSerializer().dump_obj(result_item.to_dict())
 
     # render a 404 page if the tombstone isn't visible
-    if not record.tombstone.is_visible:
+    if record_ui is None or not record.tombstone.is_visible:
```
zzacharo, line 512: why did you select here the `record_ui` to check and not e.g the `record`?

palkerecsenyi: If the error is a `PIDDeletedError` then `error.result_item` is None. And therefore `record_ui` will never be set so it will also be None. And in that case we are trying to pass `record=None` to the tombstone Jinja template which of course fails since the template tries to access various properties of the record.

The `error.record` value in `PIDDeletedError` cannot be dumped via the `UIJSONSerializer` and we cannot look up the record because the PID is of course deleted, so we don't know its UUID. As such I couldn't find a logical way to populate `record_ui` so I decided to return a not-found error which is more graceful than the 500 error we had until now (caused by the Jinja template failing to access attributes of `record`).

kpsherva (approved):

PR #3447, opened 2026-05-07
files: hide archive button over the size cap

slint:
Depends on https://github.com/inveniosoftware/invenio-records-resources/pull/682

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
@@ -99,9 +99,13 @@ To access its content, please download it and open it locally.') }}
         <th class>
           {%- if config.RDM_ARCHIVE_DOWNLOAD_ENABLED %}
             {% set archive_download_url = record.links.archive_media if is_media else record.links.archive %}
-            <a role="button" class="ui compact mini button right floated archive-link" href="{{ archive_download_url }}">
-              <i class="file archive icon button" aria-hidden="true"></i> {{_("Download all")}}
-            </a>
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
```diff
@@ -99,9 +99,13 @@ <h3 class="active title panel-heading {{ record.ui.access_status.id }} m-0">
         <th class>
           {%- if config.RDM_ARCHIVE_DOWNLOAD_ENABLED %}
             {% set archive_download_url = record.links.archive_media if is_media else record.links.archive %}
-            <a role="button" class="ui compact mini button right floated archive-link" href="{{ archive_download_url }}">
-              <i class="file archive icon button" aria-hidden="true"></i> {{_("Download all")}}
-            </a>
+            {%- set archive_max_size = config.RECORDS_RESOURCES_ARCHIVE_DOWNLOAD_MAX_SIZE %}
+            {%- set archive_total_bytes = files|map(attribute='size', default=0)|sum() %}
```
OliverGeneser, line 103: Should we add record.files.total_bytes to the template context?

OliverGeneser (approved):

invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
slint: I tried to, but we're apparently passing a serialized record... I guess it's for permission purposes as well, plus it's a bit dangerous to have "live" API/model instances in templates.

PR #3449, opened 2026-05-12
Housekeeping: Various linter fixes

max-moser:
Fix some complaints made by `ruff`

### DIFF
--- a/invenio_app_rdm/redirector/resource.py
+++ b/invenio_app_rdm/redirector/resource.py
@@ -88,7 +88,7 @@ class RedirectorResource(Resource):
         # Retrieve url from target if it's a method
         if callable(target):
             target_output = redirect_url()
-            if type(target_output) == tuple:
+            if type(target_output) is tuple:
                 redirect_url, code = target_output
             else:
                 redirect_url = target_output
--- a/tests/ui/test_sitemaps.py
+++ b/tests/ui/test_sitemaps.py
@@ -20,15 +20,15 @@ def test_sitemap_section_of_communities(running_app, community_input, create_com
     # by default converted to UTC w/ time 00:00:00
     with time_machine.travel(datetime.date(2025, 3, 27)):
         community_input["slug"] = "my-test-community-1"
-        c1 = create_community(data=community_input)
+        create_community(data=community_input)
         community_input["slug"] = "my-test-community-2"
         community_input["access"]["visibility"] = "restricted"
-        c2 = create_community(data=community_input)
+        create_community(data=community_input)
     with time_machine.travel(datetime.date(2025, 3, 28)):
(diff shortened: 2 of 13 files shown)


Samk13 (approved):

tests/ui/test_sitemaps.py
```diff
@@ -20,15 +20,15 @@ def test_sitemap_section_of_communities(running_app, community_input, create_com
     # by default converted to UTC w/ time 00:00:00
     with time_machine.travel(datetime.date(2025, 3, 27)):
         community_input["slug"] = "my-test-community-1"
-        c1 = create_community(data=community_input)
+        _c1 = create_community(data=community_input)
         community_input["slug"] = "my-test-community-2"
         community_input["access"]["visibility"] = "restricted"
-        c2 = create_community(data=community_input)
+        _c2 = create_community(data=community_input)
```
utnapischtim: question: why not remove the assignment entirely?

max-moser: because that left a "hole" in the naming (c1, c3) that i think would likely create more confusion than the rename
i also toyed with the thought of removing c2 and renaming c3 to c2, but that didn't feel great either

utnapischtim: now i am confused, i thought all of them are not used? so c1, c2, c3 are useless variables?

invenio_app_rdm/redirector/resource.py
```diff
@@ -88,7 +88,7 @@ def redirect_view_factory(self):
         # Retrieve url from target if it's a method
         if callable(target):
             target_output = redirect_url()
-            if type(target_output) == tuple:
+            if isinstance(target_output, tuple):
```
egabancho: Although correct, there is a little nuance with this change: `isinstance` is aware of inheritance, `type() ==` is not.
I don't expect this to cause any issues here, I'm probably just pointing it out because it has caused "unexpected behaviours" for me in the past 😅

egabancho (approved):

tests/ui/test_sitemaps.py
max-moser: right, it seems like i mixed up the `cX` variables here with the other named & unused variables below (`rX`, `draft_only`, `record_with_draft`, ...): it's actually the removal of `r2` that would cause the "hole" between `r1` and `r3`

looking at it again, we can remove all the `cX` variables as they're not used; i just took the same approach for all the variables.
for the others, the names do seem to convey some sort of semantics, so i'd keep them.

invenio_app_rdm/redirector/resource.py
max-moser: fair point!
i've replaced the check now with `type(...) is tuple`, which also made the linter happy and preserves the semantics better 😄

PR #3457, opened 2026-05-22
Update license headers to use SPDX

max-moser:
This PR transforms all license headers from the old format to SPDX, which is machine-readable, shorter, etc.

I've automated most of the transformation via a small Python script (no AI used), though there was also quite some manual labor involved (e.g. adding previously missing copyright statements for several files, based on the git history).

The noteworthy exception is `*.{overrides,variables}` which didn't have any copyright/license headers before, and which I'd postpone to a follow-up task for future me.


invenio_app_rdm/administration/records/__init__.py
```diff
@@ -1,9 +1,5 @@
-# -*- coding: utf-8 -*-
```
fenekku, line 1: I was going to mention the loss of the encoding line, but it seems like Pep 8 for Python itself doesn't recommend it anymore because utf-8 is the norm: https://peps.python.org/pep-0008/#source-file-encoding . So we should be fine too. Maybe for shell script we should be careful, but we never use non-ascii range characters there.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/auditLogs/index.js
```diff
@@ -1,8 +1,5 @@
-// This file is part of InvenioCommunities
-// Copyright (C) 2025 CERN.
-//
-// Invenio RDM is free software; you can redistribute it and/or modify it
-// under the terms of the MIT License; see LICENSE file for more details.
+// SPDX-FileCopyrightText: 2025 CERN.
+// SPDX-License-Identifier: MIT
```
fenekku: We should probably align on using the `/* *** */` pattern for these js comments. Not a blocker (but since we are here...)

fenekku (approved):
Overall that's fine by me!

Samk13:
Is it a good idea to add this commit to [.git-blame-ignore-revs](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/.git-blame-ignore-revs)?

Samk13 (approved):
LGTM 👍 If possible, let's exclude the `.po` and `.pot` files and also add this change to `.git-blame-ignore-revs` across all other packages as well.

invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/administration/auditLogs/index.js
max-moser: # KNEEL BEFORE THE MIGHTY GNU `sed` MAGIC

`rg '// SPDX' | cut -d: -f 1 | sort | uniq | xargs -l sed -i -e '1s_^_/*\n_;1{:a;s_// SPDX_ * SPDX_;Tb;n;ba;:b;/^$/s__ */\n_}'`

image

Here, I search for all files that have `// SPDX` in them; if piped into a command or file, `rg` reports matches in the shape of `filename: match`.
From these matches, I extract the filename with `cut`, sort the output and deduplicate with `uniq`.
This gives me a list of all files that have `// SPDX` in them; and for each filename I run a little GNU `sed` script.

# Explainer for the `sed` script

It's a two-step `sed` script; the semicolon (`;`) delimits commands (similar to other languages), leaving us with:

`1s_^_/*\n_`, which addresses the first line and `s`ubstitutes the beginning (`^`) with `/*\n` (i.e. `/*` and a newline).
The three underscores (`_`) are delimiters for the `s` command.
Typically you'd use `/` for that, but another character can be chosen as long as all three characters are replaced.
This is handy e.g. when forward slashes appear in the regexes.

`1{:a;s_// SPDX_ * SPDX_;Tb;n;ba;:b;/^$/s__ */\n_}'`, which also matches the first line, and calls a few subcommands in a block (in squiggly brackets):

---

`:a` defines the label `a`, which can be jumped to.
`s_//SPDX_ * SPDX_` substitutes `//SPDX` with ` * SPDX`.
`Tb` checks if there was a successful `s`ubstitution recently (since the last line was read) and *if not* jumps to the label `b` [1]. Basically, if we encounter a "different" line, we jump.
`n` reads the next line.
`ba` unconditionally jumps to the label `a`.
`:b` defines the label `b`.
`/^$/s__ */\n_` checks if the line is empty (`/^$/`) and if so, adds ` */\n` (i.e. ` */` + newline) [2].

[1] `T` is a GNU extension. Like I've said before, `sed` without the GNU extensions is just `sed`.
[2] I guess the `/^$/` check is almost detrimental because it requires an empty line to be present between the license header and anything else (e.g. code) and could/should be removed... But this is the version I ran.

## TL;DR
Basically, we insert `/*` on the first line and then run a little loop that eats up and transforms lines looking like `^// SPDX` into ` * SPDX` until we encounter something else.
If that other thing is an empty line, we add ` */` and call it a day.

fenekku: I kneel to your sed-fu.

max-moser: Turns out the `/^$` check was actually detrimental; removing it helped.
I also have a feeling that some things could be made a bit less awkward if one had a better understanding of the difference between pattern/hold space for `sed`, but it is what it is.

max-moser:
For the record: Some of the PRs are marked as closed but have actually been merged (via the CLI, to preserve the commit hashes better, and avoid some extra work with `.git-blame-ignore-revs`).

On some PRs, I didn't force-push the PR base branch before merging, so GitHub got confused and shows the PR as closed.

Samk13:
Thanks @max-moser for pushing this through, you've literally made it impossible to blame you for this 😅

max-moser:
There are some packages that didn't have a `.git-blame-ignore-revs` where I didn't introduce it, so it should still be possible to `blame` me in some cases :smile: 
Also, there were a few packages seemingly fully licensed under GPL that I didn't touch because I was unsure if that was just a leftover or not.

But overall, looks like I might've now deleted more lines than added in InvenioRDM overall!

PR #3459, opened 2026-05-27
feat(upgrade): patch role IDs for bleeding-edge instances (Not for v13 migration)

Samk13:

### Description

* Add one-time migration for restoring role ids from role names and updating related grants.
* Reindex groups after migration so search data matches the updated role records.
* Depends on https://github.com/inveniosoftware/invenio-users-resources/pull/201

The data migration script is only needed for instances that were running bleeding-edge InvenioRDM after the role name/id split and that used the roles feature before the beta release. Instances upgraded before the beta release do not need to run this migration.

#### Database migration

Execute the data migration:

```bash
invenio shell $(find $(pipenv --venv)/lib/*/site-packages/invenio_app_rdm -name migrate_role_ids_to_names.py)
```

#### Reindex

After the data migration, rebuild the search indices:

```bash
invenio index destroy --yes-i-know
invenio index init
invenio rdm rebuild-all-indices
```
If you prefer not to rebuild all indices, you can instead run the groups/users reindex custom script:

invenio shell [here](https://github.com/Samk13/invenio-dev-latest/blob/master/scripts/indices-migration-v13-14/user-groups-indices-migration.py)

> Note: This PR includes AI-assisted contributions and has been reviewed and refined manually.

### Checklist

**Frontend**

**Reminder**

### DIFF
--- a/invenio_app_rdm/upgrade_scripts/migrate_role_ids_to_names.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_role_ids_to_names.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2026 CERN.
+#
+# Invenio-App-RDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Role migration script for InvenioRDM 14.0.
+
+Disclaimer: This script is intended to be executed *only once*, for v14
+instances where roles may have generated ids different from their names.
+It restores the invariant that ``accounts_role.id == accounts_role.name``.
+"""
+
+import sys
+
+from click import secho
+from invenio_access.models import ActionRoles
+from invenio_accounts.models import Role
+from invenio_db import db
+
+
+def execute_upgrade():
+    """Execute the role id/name migration."""
+    secho("Starting role id/name migration...", fg="green")
+
+    roles = Role.query.filter(Role.id != Role.name).all()
+
+    errors = [
+        f"Role <{role.id}> cannot be migrated to <{role.name}>: "
+        "target id already exists."
+        for role in roles
+        if db.session.get(Role, role.name)
+    ]
+
+    if errors:
+        db.session.rollback()
+        secho("Upgrade aborted due to the following errors:", fg="red", err=True)
+        for error in errors:
+            secho(error, fg="red", err=True)
+        sys.exit(1)
+
+    for role in roles:
+        old_id = role.id
+        new_id = role.name
+        users = list(role.users)
+        secho(f"Updating role <{old_id}> to <{new_id}>.", fg="yellow")
+
+        # Role.id is referenced by foreign keys without ON UPDATE CASCADE, so
+        # replace the row instead of mutating the primary key in place.
+        role.name = None
+        db.session.flush()
+
+        replacement_role = Role(
+            id=new_id,
+            name=new_id,
+            description=role.description,
+            is_managed=role.is_managed,
+        )
+        replacement_role.users = users
+        db.session.add(replacement_role)
+
+        ActionRoles.query.filter_by(role_id=old_id).update(
+            {ActionRoles.role_id: new_id}, synchronize_session=False
+        )
+        db.session.delete(role)
+
+    secho("Committing to DB", fg="green")
+    db.session.commit()
+    secho(
+        f"Data migration completed. Updated {len(roles)} role(s). "
+        "Please rebuild the search indices now.",
+        fg="green",
+    )
+
+
+if __name__ == "__main__":
+    execute_upgrade()
--- a/invenio_app_rdm/upgrade_scripts/migrate_role_ids_to_names.py
+++ b/invenio_app_rdm/upgrade_scripts/migrate_role_ids_to_names.py
@@ -1,9 +1,6 @@
-# -*- coding: utf-8 -*-
-#
-# Copyright (C) 2026 CERN.
-#
-# Invenio-App-RDM is free software; you can redistribute it and/or modify
-# it under the terms of the MIT License; see LICENSE file for more details.
+# SPDX-FileCopyrightText: 2026 KTH Royal Institute of Technology.
+# SPDX-FileCopyrightText: 2026 Northwestern University.
+# SPDX-License-Identifier: MIT
 
 """Role migration script for InvenioRDM 14.0.
 
@@ -16,8 +13,10 @@ import sys
 
 from click import secho
 from invenio_access.models import ActionRoles
-from invenio_accounts.models import Role
+from invenio_accounts.models import Role, userrole
+from invenio_communities.members import MemberModel
 from invenio_db import db
+from sqlalchemy import update
 
 
 def execute_upgrade():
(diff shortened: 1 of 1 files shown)


invenio_app_rdm/upgrade_scripts/migrate_role_ids_to_names.py
```diff
@@ -0,0 +1,78 @@
...
+
+import sys
+
+from click import secho
+from invenio_access.models import ActionRoles
+from invenio_accounts.models import Role
+from invenio_db import db
+
+
+def execute_upgrade():
```
egabancho, line 22: Do you think we should treat this one as an Alembic recipe inside Invenio-Accounts?

Samk13: I'm not very familiar with Alembic recipes myself. If you'd like to take care of it, feel free to pick this one up. I can revert the migration script on my side.

utnapischtim: the problem is, with an alembic script in invenio-accounts, i would have to start again a major bump chain, and with invenio-accounts, it is a longer one. the second argument against an alembic script is, that normally we do alembic scripts to change tables not values. since this migration script is to change values, i would go with an migration script and keep it here

ntarocco: When choosing between:
1. a script or alembic recipe in `invenio-accounts`
2. a script here, when upgrading InvenioRDM
should take into account if the data migration is required also by who is using `invenio-accounts` directly, without InvenioRDM (other Invenio installations, or custom RDMs). Even if tedious, a release chain might be required depending on what's best.

In this case, is it only an InvenioRDM feature? Or is it related to invenio-accounts only?

invenio_app_rdm/upgrade_scripts/migrate_role_ids_to_names.py
```diff
@@ -0,0 +1,78 @@
...
+    """Execute the role id/name migration."""
+    secho("Starting role id/name migration...", fg="green")
+
+    roles = Role.query.filter(Role.id != Role.name).all()
```
ntarocco, line 26: Should we exclude `is_managed == False` when running this? Unmanaged groups are coming from external identity systems.

Samk13: I don’t think we should exclude `is_managed == False`.

 After discussing the scenarios with @utnapischtim @egabancho @zzacharo : 
IdP groups usually come in a response as string names, without a separate ID. So the group name is the identifier we should persist.
 This also matches the current service behavior: new groups are created with `id == name`, duplicate names are rejected, and names are not editable.

 If we leave old unmanaged groups as e.g.:`id=<uuid>, name="admin"`, a later sync could treat "admin" inconsistently. Keeping `id == name` for unmanaged
 Groups, too, will keep "admin" as a single DB role, and any IdP-specific sync logic stays in the IdP service layer.

invenio_app_rdm/upgrade_scripts/migrate_role_ids_to_names.py
Samk13: The tables/models are from invenio-accounts true. Still, the behavior that created the problematic data was introduced via `invenio-users-resources` role/group handling, not by `invenio-accounts` itself, so that would belong to app-rdm instead of accounts IMO.

egabancho: Thank y'all for the discussion. I looked into the Alembic route, and it did open my eyes ☺️ 

@ntarocco, I think you raised a valid concern, this isn't RDM-specific, it's an Invenio-Accounts model constraint. Ideally, the migration should live where all consumers can access it (i.e., app-ils). The natural candidate would be an Alembic recipe in Invenio-Access, since it owns ActionRoles (the main FK to accounts_role.id), where permissions and roles are checked.

However, looking at the actual dependency tree, a major bump (due to the Alembic recipe) in Invenio-Access would cascade through Invenio-Records-Permissions → Invenio-Records-Resources → and, from there, into basically every higher-level package (~15+ repos).

I agree with @utnapischtim, the Alembic route isn't worth the cascade here. I think we should keep the script here. If other flavours like App-ILS need it down the line, the script is straightforward enough to copy or extract without tying it to a major release chain.

invenio_app_rdm/upgrade_scripts/migrate_role_ids_to_names.py
```diff
@@ -0,0 +1,78 @@
...
+
+    for role in roles:
+        old_id = role.id
+        new_id = role.name
+        users = list(role.users)
+        secho(f"Updating role <{old_id}> to <{new_id}>.", fg="yellow")
+
+        # Role.id is referenced by foreign keys without ON UPDATE CASCADE, so
+        # replace the row instead of mutating the primary key in place.
+        role.name = None
```
fenekku: Hmm this doesn't work because of https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/models.py#L128 . Will have to think of another way.

Samk13: This was working before last month's change prevented renaming persisted roles.

We can adjust the migration to work around the setter without changing the Role model, probably by doing the update directly at the DB level WDYT?

Samk13: Maybe we can just change:
```py
role.name = None
db.session.flush()
```
to:
```py
db.session.execute(
    Role.__table__.update()
    .where(Role.__table__.c.id == old_id)
    .values(name=None)
)
```
This should bypass the setter while keeping the rest of the migration unchanged ?

fenekku: Ended up having to do what is in the new commit.

- Lean on SQLAlchmey 2.0 modern syntax
- Circumvents contraints of name not being modifiable AND name being autoset to id at creation
- Deals with all foreign keys to Role

fenekku:
I can look at black on Monday (would put a `# fmt: no` directive block  personally to make it more readable).

invenio_app_rdm/upgrade_scripts/migrate_role_ids_to_names.py
Samk13: With a quick AI review, it seems this could be simplified by avoiding the temporary role name and `is_managed=False`, and freeing the existing name directly:
```py
db.session.execute(
    update(Role.__table__)
    .where(Role.__table__.c.id == old_id)
    .values(name=None)
)
```
I haven't executed it though, so if your version works for you, I wouldn’t push on this further.

fenekku: Tried it and it works, thanks. Can you put a validation mark and we will merge this.

fenekku (approved):
This script is actually necessary for roles that have been created long enough ago.

invenio_app_rdm/upgrade_scripts/migrate_role_ids_to_names.py
```diff
@@ -0,0 +1,107 @@
...
+        # and delete old Role afterwards. And do it all using "low-level"
+        # SQL primitives that avoid the application-level constraints.
+
+        db.session.execute(update(Role).where(Role.id == old_id).values(name=None))
+
+        # Create new Role with id == name
+        replacement_role = Role(
+            id=new_id,
+            name=new_id,
+            description=role.description,
```
Samk13, line 60: Should we also preserve the `is_managed` state here?
```suggestion
            description=role.description,
            is_managed=role.is_managed,
```

fenekku: Oups, absolutely! Updated.

fenekku:
Let's also get this in 14.0.1 . So flask-caching PR: https://github.com/inveniosoftware/invenio-app-rdm/pull/3558 should be get in first.

PR #3486, opened 2026-06-19
feat(records-ui): eagerly provide the default record citation text

max-moser:
This saves an XHR from the landing page, and doesn't really cost anything because we already have the record in question resolved.

### DIFF
--- a/invenio_app_rdm/records_ui/views/records.py
+++ b/invenio_app_rdm/records_ui/views/records.py
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: 2019-2025 CERN.
 # SPDX-FileCopyrightText: 2019-2021 Northwestern University.
-# SPDX-FileCopyrightText: 2021-2023 TU Wien.
+# SPDX-FileCopyrightText: 2021-2026 TU Wien.
 # SPDX-FileCopyrightText: 2025 CESNET i.a.l.e.
 # SPDX-License-Identifier: MIT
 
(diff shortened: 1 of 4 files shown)


invenio_app_rdm/records_ui/views/records.py
```diff
@@ -311,6 +317,18 @@ def record_detail(
         )
     )
 
+    # since we've already resolved the record, we might as well generate
+    # the citation text in the default style, to save ourselves an XHR
+    default_citation_style = current_app.config.get("RDM_CITATION_STYLES_DEFAULT", None)
+    csl_serializer = CSLJSONSerializer()
+    style = get_style_location(default_citation_style)
+    record_citation = get_citation_string(
```
Samk13: What happen if it fails for any reason? should we try/catch or log something here?

max-moser: ACK
I've added a `try/except` so that if *anything* goes wrong around here, the behavior falls back to what it was before (basically embeds an empty citation into the HTML, which causes the React component to fetch the citation from the backend).

invenio_app_rdm/records_ui/views/records.py
```diff
@@ -311,6 +317,23 @@ def record_detail(
...
+    # since we've already resolved the record, we might as well generate
+    # the citation text in the default style, to save ourselves an XHR
+    try:
+        default_citation_style = current_app.config.get(
+            "RDM_CITATION_STYLES_DEFAULT", None
+        )
+        csl_serializer = CSLJSONSerializer()
+        style = get_style_location(default_citation_style)
+        record_citation = get_citation_string(
+            csl_serializer.dump_obj(record.data),
```
Samk13, line 329: Looking [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/201bdaeb61ead81d6ccf1f8e5d5f191d7a61eaf1/invenio_rdm_records/resources/serializers/csl/__init__.py#L151-L155) should we also inject the `_extras["links"]`?

PR #3495, opened 2026-06-29
chore(setup): migrate from setuptools to hatchling

max-moser:
This PR moves from `setuptools` as build tool to `hatchling`.

The largest chunk of the translation was done automatically via `uvx hatch new --init`, which translated most of the content from `setup.cfg` to `pyproject.toml`.
This translation is not 100% complete however, and needs some manual tweaks:
* the dependencies weren't ordered as nicely as before, I fixed that manually
* moved the `[project.urls]` section to the top
* manually translated the configuration for some tools:
  * `[tool.isort]` for `isort`
  * `[tool.pytest.ini_options]` for `pytest`
  * `[tool.pydocstyle]` for `pydocstyle`

Two tool configs are currently still missing, namely `sphinx` and the translation config (`babel`?).
That means, roughly the following chunk from `setup.cfg` still needs to be translated:

```ini
[build_sphinx]
source-dir = docs/
build-dir = docs/_build
all_files = 1

[compile_catalog]
directory = invenio_app_rdm/translations/
use_fuzzy = True

[extract_messages]
copyright_holder = CERN
msgid_bugs_address = info@inveniosoftware.org
mapping_file = babel.ini
output_file = invenio_app_rdm/translations/messages.pot
add_comments = NOTE

[init_catalog]
input_file = invenio_app_rdm/translations/messages.pot
output_dir = invenio_app_rdm/translations/

[update_catalog]
input_file = invenio_app_rdm/translations/messages.pot
output_dir = invenio_app_rdm/translations/
```

## Translation

At TU Wien we have a shell script [`babel.sh`](https://gitlab.tuwien.ac.at/crdm/invenio/invenio-config-tuw/-/blob/master/babel.sh?ref_type=heads) (and `babel.ini`) in our `Invenio-*-TUW` packages that (hopefully) take care of this configuration, but we're not really using translations (yet).
Ping @Samk13 because you said you have a possible solution here at hand?

## Sphinx

I cannot actually remember ever having used Sphinx for docs in InvenioRDM packages (and https://invenio-app-rdm.readthedocs.io/ gives a 404).
Should we remove the `docs/` and Sphinx?
Ping @slint @ntarocco 

---

**Edit**: The [Sphinx docs](https://sphinx-themed.readthedocs.io/en/stable/usage/advanced/setuptools.html) make it sound like we've been using Sphinx in a way that the Setuptools integration (which is deprecated and will be removed as well) was never used anyway?
> Once configured, call this by calling the relevant command on setup.py:
> `$ python setup.py build_sphinx`

In `run-tests.sh` we set the source and output directories along with some values (`quiet`, `nitpicky`, `no-color`) explicitly: `python -m sphinx.cmd.build -qnN docs docs/_build/html`
If we're serious about the `write-all` flag, we can simply add that to the calls?

**Edit 2**: Quickly checked for any noticeable negative effects due to the removed Sphinx config in https://github.com/inveniosoftware/invenio-files-rest/pull/358 since `Invenio-Files-REST` actually uses the docs, couldn't find any ill effects on my locally built docs on a glance.


Samk13:
> ## Translation

Yes, @max-moser I think we should do something similar to [invenio-collections pyproject.toml](https://github.com/inveniosoftware/invenio-collections/pull/123/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) and include this line in [push messages CI](https://github.com/Samk13/invenio-collections/blob/c83d413560a48ff1e6ab06eb2ed56a6f71cadf22/.github/workflows/i18n-push.yml#L16), as push strings to transifex is now supported [in i18n CI](https://github.com/inveniosoftware/invenio-i18n/pull/171)

Also, the publish workflow is here for reference: [publish CI](https://github.com/inveniosoftware/workflows/pull/30)

max-moser:
Sweet, thanks!
I've adapted the config from `Invenio-Collections` and added it to a new commit here, could you please give it a quick pass if that makes sense?

Samk13 (approved):
LGTM!

Not a blocker: 
If we could also update the instructions in the [.tx/config](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/.tx/config) file to adress the new commands, like in [here](https://github.com/Samk13/invenio-collections/blob/c83d413560a48ff1e6ab06eb2ed56a6f71cadf22/.tx/config#L28-L31), that would be great.

You can also safely remove the `babel.ini` file as well, since its configuration has now been migrated to `pyproject.toml`.

max-moser:
Thanks @Samk13 for the pointers and checks!
I've added the suggestions as well.

@ntarocco @slint I haven't observed any negative effects from omitting the Sphinx configuration in `pyproject.toml`, so I assume this is fine.
Perhaps we could remove the `docs/` stuff altogether in packages that haven't been using it so far (i.e. InvenioRDM packages), as I don't believe that we ever will use it there anymore.
This is a discussion to be had on another day though, perhaps the next maintainers meeting?

---

Mostly for my own future reference, here's the rundown of what I did in this PR:

* Run `hatch new --init` for the base variant of `pyproject.toml`
* Improve some content if desirable
* Add configuration for tools (`[tool.isort]`, `[tool.pytest.ini_options]`, `[tool.pydocstyle]`, `[tool.babel.*]` (project-specific)
* Remove `MANIFEST.in` because that's a `setuptools` thing that's not needed anymore
* Remove `python -m check_manifest` line from `run-tests.sh`
* Remove `setup.cfg` and `setup.py`
* Add `uv.lock` to `.gitignore` because that just adds noise
* Add `backend-package-path` in GH workflows (`i18n-push.yml`) (project-specific)
* Remove `babel.ini`
* Update commands in `.tx/config` if available (project-specific)
* Specify TOML indenting rule in `.editorconfig`
* Format and lint `pyproject.toml`
* Bump `requires-python` if needed (e.g. due to dependencies)

Last but not least, test everything with `uv sync` :slightly_smiling_face:

slint (approved):
Nice, 4 files less to pollute the root directory! 👌

pyproject.toml
```diff
@@ -1,3 +1,227 @@
...
+# SPDX-FileCopyrightText: 2024 Ubiquity Press.
+# SPDX-FileCopyrightText: 2025-2026 KTH Royal Institute of Technology.
+# SPDX-FileCopyrightText: 2026 TU Wien.
+# SPDX-License-Identifier: MIT
+
+[project]
+name = "invenio-app-rdm"
+description = "Invenio Research Data Management."
+readme = "README.rst"
+requires-python = ">=3.9"
```
max-moser, line 14: [`Invenio-Collections` requires at least 3.9.](https://github.com/inveniosoftware/invenio-collections/blob/master/pyproject.toml#L31)
Since that's a dependency for `Invenio-App-RDM`, requiring something lower here would be semantically invalid, and that actually gets caught by `uv`:

```sh
$ uv sync --extra opensearch2

  x No solution found when resolving dependencies for split (markers: python_full_version == '3.8.*'; included: invenio-app-rdm[opensearch2]; excluded: invenio-app-rdm[elasticsearch7], invenio-app-rdm[opensearch1]):
  `-> Because the requested Python version (>=3.7) does not satisfy Python>=3.9 and invenio-collections==10.0.0 depends on Python>=3.9, we can conclude that invenio-collections==10.0.0 cannot be used.
      And because only invenio-collections<=10.0.0 is available and your project depends on invenio-collections>=10.0.0, we can conclude that your project's requirements are unsatisfiable.

hint: While the active Python version is 3.14, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions using `requires-python`.
hint: The `requires-python` value (>=3.7) includes Python versions that are not supported by your dependencies (e.g., invenio-collections==10.0.0 only supports >=3.9). Consider using a more restrictive `requires-python` value (like >=3.9).%                                            
```

max-moser:
I've created a shell script that performs this migration mostly automatically (though some manual fixups may be necessary, e.g. `nvim -p pyproject.toml .gitignore`):

```sh
#!/bin/bash

trans_done=0
conflict_done=0

pkg_name="$(basename "$PWD")"
pkg_dir="${pkg_name//-/_}"

# populate the venv (necessary for hatch new)
uv venv --clear && uv pip install -e . && . .venv/bin/activate

# translate setup.cfg to pyproject.toml
uvx hatch new --init

# fix up the missing stuff
[[ -d "${pkg_dir}/translations/" ]] && trans_done=1 && cat >> pyproject.toml << EOF
# see https://babel.pocoo.org/en/latest/messages.html#toml-configuration-format
[tool.babel]
[tool.babel.compile_catalog]
directory = "${pkg_dir}/translations/"
use_fuzzy = true

[tool.babel.extract_messages]
copyright_holder = "CERN"
msgid_bugs_address = "info@inveniosoftware.org"
mapping_file = "pyproject.toml"
output_file = "${pkg_dir}/translations/messages.pot"
add_comments = "NOTE"

[tool.babel.init_catalog]
input_file = "${pkg_dir}/translations/messages.pot"
output_dir = "${pkg_dir}/translations/"

[[tool.babel.mappings]]
method = "javascript"
pattern = "**.js"
encoding = "utf-8"
extract_messages = "$._, jQuery._"

[[tool.babel.mappings]]
method = "python"
pattern = "**.py"
encoding = "utf-8"

[[tool.babel.mappings]]
method = "jinja2"
pattern = "**/templates/**.html"
encoding = "utf-8"

EOF

cat >> pyproject.toml << EOF
[tool.isort]
profile = "black"

[tool.pydocstyle]
add_ignore = "$(sed -nE 's/^add_ignore = (.*)/\1/p' setup.cfg)"

[tool.pytest.ini_options]
addopts = '--black --isort --pydocstyle --doctest-glob="*.rst" --doctest-modules --cov=${pkg_dir} --cov-report=term-missing'
testpaths = "tests ${pkg_dir}"
live_server_scope = "module"
EOF

if grep '^filterwarnings' setup.cfg; then
  sed -nE 's/^filterwarnings = (.*)/filterwarnings = ["\1"]/p' setup.cfg >> pyproject.toml
fi

grep -i 'invenio-search\[opensearch2]' pyproject.toml && conflict_done=1 && cat >> pyproject.toml << EOF

# UV needs to be told explicitly which extras are conflicting:
# https://docs.astral.sh/uv/concepts/projects/config/#conflicting-dependencies
[tool.uv]
conflicts = [
  [
    { extra = "opensearch1" },
    { extra = "opensearch2" },
    { extra = "elasticsearch7" },
  ]
]
EOF

# set license headers
{
  sed -n '/SPDX-FileCopyrightText/p' setup.cfg setup.py | sort | uniq
  echo '# SPDX-FileCopyrightText: 2026 TU Wien.'
  echo "# SPDX-License-Identifier: MIT"
  echo
  sed -E 's/description = "\\"(.*)\\""/description = "\1"/' pyproject.toml
} > pyproject.toml2
mv pyproject.toml2 pyproject.toml

# remove unnecessary files
rm -f MANIFEST.in setup.cfg setup.py babel.ini

# ignore uv.lock
if grep Pipfile.lock .gitignore; then
  sed -E 's/^Pipfile.lock/Pipfile.lock\nuv.lock/' -i .gitignore
else
  echo uv.lock >> .gitignore
fi

# remove "check_manifest" call from run-tests
[[ -f run-tests.sh ]] && sed -e '/check_manifest/d' -e "s|^.*setup extract_messages.*|pybabel extract -F pyproject.toml ${pkg_dir} --output-file /dev/null|" -i run-tests.sh

# add the backend-package-path to the i18n push workflow
[[ -f ".github/workflows/i18n-push.yml" ]] && sed -E "s/^(.*)(extract-backend:.*)/\1\2\n\1backend-package-path: ${pkg_dir}/" -i ".github/workflows/i18n-push.yml"

# set indent rules for TOML
cat >> .editorconfig << EOF

# Align TOML indenting with Tombi
[*.toml]
indent_size = 2
EOF

# reformat pyproject.toml
tombi format pyproject.toml

# create the commit
git add .
read -d '\n' commit_message << EOF
chore(setup): migrate from setuptools to hatchling

* remove MANIFEST.in because that's only relevant for setuptools
INSERT
* format pyproject.toml with Tombi

Co-authored-by: Sam Arbid <samaa2@kth.se>'
EOF

[[ "$trans_done" != "0" ]] && commit_message="${commit_message/INSERT/* translate the i18n/translation mechanisms as well\nINSERT}"
[[ "$conflict_done" != "0" ]] && commit_message="${commit_message/INSERT/* declare conflicting extras for uv\nINSERT}"
commit_message="$(echo "$commit_message" | sed -e '/^INSERT/d')"
git commit -m "$commit_message"
```

max-moser:
For the record: I've compared the built `sdist` and `wheel` distributions generated by the old `setuptools` and `hatchling` build backends, and identified no blockers.
I'll go ahead and start merging and releasing packages, starting with `Invenio-Assets` and other frontend-heavy packages, as I'd imagine the opportunities for any learnings are most pronounced there.

---

A quick comparison between the distributed files (for `sdist` and `wheel`) generated from the `master` branch (i.e. `setuptools`) and this branch (i.e. `hatchling`) showed that we're generally distributing fewer files:
```bash
delta <(tar tf dist/setuptools/*.tar.gz | sort) <(tar tf dist/hatchling/*.tar.gz | sort)
delta <(zipinfo -1 dist/setuptools/*.whl | sort) <(zipinfo -1 dist/hatchling/*.whl | sort)
```

For instance, `docs/` and `tests/` aren't included in the distributions anymore; same for "dev-only" stuff like `.github/` and `.tx/`.
I'd argue that development files aren't supposed to be part of source distributions anyway, so that's fine.
The exception is `.gitignore` which wasn't included before, but is included now.
Hatchling also seems to exclude (superfluous) directory entries from the `sdist` tar file (but of course adds the contained files), so the content listings are a few lines shorter because of that too.

In the source distribution, there also used to be some `*.egg-info/*` files.
The PyPA mentions "eggs" together with `setuptools` [1], so I believe that's just some leftover from that, which we now got rid of.

---

I found that with this migration, we're going to distribute a good amount of files fewer than before, which at times significantly reduces the distributions' sizes:
```bash
mmoser@mx ~/rdm/packages/invenio-rdm-records [mm/hatchling-setup] $ ll dist/*
dist/hatchling:
Permissions  Size User   Group  Date Modified Name
.rw-r--r--  1.6Mi mmoser mmoser  1 Jul 13:40  invenio_rdm_records-33.0.0-py3-none-any.whl
.rw-r--r--  1.1Mi mmoser mmoser  1 Jul 13:40  invenio_rdm_records-33.0.0.tar.gz

dist/setuptools:
Permissions  Size User   Group  Date Modified Name
.rw-r--r--  1.9Mi mmoser mmoser  1 Jul 13:41  invenio_rdm_records-33.0.0-py2.py3-none-any.whl
.rw-r--r--  1.4Mi mmoser mmoser  1 Jul 13:40  invenio_rdm_records-33.0.0.tar.gz
```

---

[1] https://packaging.python.org/en/latest/glossary/#term-Egg

max-moser:
Update 1: Released `Invenio-Assets` v4.2.3 as a kind of canary:
It got picked up by our v13 instance, everything is going fine.

- The expected files from the package are all present in the installation (in the virtual environment)
- The build runs through as before
- No changes in operation of the instance

@mesemus also got this package version in, and the release process went as expected.

---

Update 2: The testrig run for `Invenio-DB` reports zero errors: https://github.com/max-moser/invenio-testrig-client/actions/runs/28518270482

---

Update 3: Deployed `Invenio-Assets` v4.2.3 to production; see if something fails: https://researchdata.tuwien.ac.at/

---

Update 4: Successfully released `Invenio-Theme` v5.0.2 which will get picked up by `Invenio-App-RDM` v15 development releases soon (already does get picked up by the state on `master`).
Got it into a local v15 `my-site`; it installed and built successfully, and is running as expected.

The publish workflow for v5.0.1 was broken because of a step in the workflow still referencing `setup.py` (which got removed), the fix was to use the new workflow by setting `backend-package-path` in `pypi-publish.yml`: https://github.com/inveniosoftware/invenio-theme/pull/437

---

Update 5: Successfully released and installed `Invenio-RDM-Records` v33.1.0.
Installation, build, and operation work as expected.

I get the feeling we're in a good spot :slightly_smiling_face: 

---

Update 6: Released a few non `invenio-*` packages (like `dcxml`, `flask-kvsession`, etc.) built with `hatchling` that got picked up by our v13 setup.
The version is already live on our test system, no issues detected.
Actually to the contrary: It looks like before, some test dependencies (like `pytest-*`) were accidentally added by one of these packages; these test packages don't get pulled in to our setup anymore.

PR #3501, opened 2026-07-01
chore(migration): make theses migration optional

sakshamarora1:
The documentation and the changes here synced: https://github.com/inveniosoftware/docs-invenio-rdm/pull/920


fenekku (approved):
That works for me. Let's make sure to have the corresponding docs in the upgrade notes. There will be a discussion between maintainers before this approach is fully approved or something else is chosen, so I don't think we want to merge yet.

(the tests should be made to pass before too)

utnapischtim:
superseded by #3518

PR #3514, opened 2026-07-15
Backport to v14

tmorrell:

### Description

Backport the license headers as decided in call.

### DIFF
--- a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
+++ b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
@@ -1,12 +1,10 @@
 {#
-  Copyright (C) 2020-2026 CERN.
-  Copyright (C) 2023 Northwestern University.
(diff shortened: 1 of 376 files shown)


invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/files.html
```diff
@@ -144,13 +144,15 @@ <h3 class="active title panel-heading {{ record.ui.access_status.id }} m-0">
             </small>
             {%- endif %}
           </td>
-          <td>{%- if is_remote_file %}{{_("N/A (external)")}}{%- else -%}{{ file.size|filesizeformat(binary=binary_sizes) }}{%- endif %}</td>
+	  <td>{{ file.size|filesizeformat(binary=binary_sizes) or _("N/A")}}</td>
```
utnapischtim: this is a blocker for me. the problem i see is that the new msgid is not translated, and the old is useless. The same applies for "(external)"

tmorrell:
Now just the license headers @utnapischtim

PR #3518, opened 2026-07-21
chore(v14/migration): restructure script

utnapischtim:
* to enable the request comment feature it is enough to run following
  steps which are included into the normal upgrade guide:
  - invenio index destroy --yes-i-know
  - invenio index init
  - invenio rdm rebuild-all-indices

* the dissertation vocabulary change is optional and the script is moved
  to docs-invenio-rdm


fenekku:
@utnapischtim Looks like fastjsonschema is using type hint syntax that Python 3.9 doesn't support. So either we curtail the versions of that dependency to still support 3.9 or this marks the official end of supporting 3.9 and likely the *requirement* to use 3.14.

PR #3565, opened 2026-08-31
config: remove unused APP_RDM_RECORDS_EXPORT_URL config

ptamarit:
* The config used for this route is `record_export` in `APP_RDM_ROUTES`.
* I did not find any other traces of this config in the code or in the docs.


fenekku (approved):
Thanks! This is has indeed been replaced (was used for `record_export` prior to `records_ui/` refactoring and `APP_RDM_ROUTES` consolidation).

PR #3566, opened 2026-08-31
fix: optionally chain user in RecordManagement.recordOwnerID to account for system records

fenekku:
Without this fix, records with `record.parent.access.owned_by == null` (typical of System created records) will raise an error and break all JS on the record's landing page.

We need this fix as a large portion of our collection is System records.

I will backport this to v14.


tmorrell (approved):
This returns the behavior to back when this was run through python with a get. I don't think the intention was to require owned_by...and breaking the landing page would be a weird way to enforce that requirement. Good to go!

PR #3581, opened 2026-09-16
feat(versions): add resource_type to deserializer

liptakpanna:
Related to https://github.com/CERNDocumentServer/cds-rdm/issues/944


invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordVersionsList.js
```diff
@@ -22,6 +22,7 @@ const deserializeRecord = (record) => ({
   links: record.links,
   pids: record.pids,
   new_draft_parent_doi: record.ui.new_draft_parent_doi,
+  resource_type: record.ui.resource_type,
```
utnapischtim, line 25: nit: this seems like a small change, but it would be nice to understand why it is necessary. can you please add some reason to the commit message and use https://www.conventionalcommits.org/en/v1.0.0/ to write your commit message. so is this a fix or a feature?

liptakpanna: thanks, I added the explanation to the commit message, it is a feature. The reason for this change is to display the resource type in the versions list. For CDS, `RecordVersionsList.Item.container` is an overridden element, but currently the resource type isn’t available in the `item` object.

utnapischtim: thanks for updating it with the reason

still some small nit which i forgot before: please have a look to https://inveniordm.docs.cern.ch/community/code/best-practices/commits/ about line length and how the commits should be structured. thanks

fenekku (approved):

